PR #1, opened 2026-04-10
feat: attend — active awareness module with peer signaling

aaronsb:
## Summary

- Implements ADR-113: `attend` binary with sensor loop, peer awareness, inter-session signaling, and focus groups
- 4 sensors: context (interoceptive), git, peers, processes — all polling on adaptive schedules with disclosure governor
- Full CLI: `run`, `send`, `inbox`, `peers`, `focus`, `config`, `status`, `--version`
- Declarative config with project-scope overlay (ADR-115): `~/.config/attend/config.yaml` + `{project}/.claude/attend.yaml` with `+/-` sensor syntax
- State persistence: checkpoint/restore sensor baselines across restarts
- Self-reload: attend watches its own binary mtime and exec's itself on change
- `/attend` skill, attend way with dynamic macro, embedding vocabulary tuned
- ADR-113 revised from Draft to Accepted, ADR-115 (config pattern) drafted

### Inter-session collaboration

Multiple Claude Code sessions + human terminals communicate through filesystem-based signals. No external transport, no protocol — signal files in `~/.cache/attend/signals/` scoped by project. Focus groups let sessions form collaboration clusters. Sender identity automatically detects Claude sessions vs human terminals.

Validated in live session: 3 Claude instances (agent-ways, temp, game-ai-pro) + 1 human terminal, all communicating through attend.

## Test plan

- [x] `make setup` builds from clean (ways + attend + embedding engine)
- [x] All 15 embedding engine tests pass
- [x] `attend run` starts sensor loop, establishes baselines
- [x] `attend send` / `attend inbox` deliver messages between sessions
- [x] `attend focus add/list/clear` manages focus groups
- [x] `attend config init/show/path` manages configuration
- [x] Sidetone prevention: own messages filtered by from-field session ID
- [x] Forward-only mode: restart doesn't replay old signals
- [x] State checkpoint persists seen_signals and disclosed_thresholds
- [x] Context sensor coordinates with ways thresholds (no overlap)
- [x] Way vocabulary tuned: attend wins on attend-specific queries, ea/comms wins on generic messaging

Tracking issue for remaining work: #2


aaronsb:
## Code Review: attend -- active awareness module

**PR Size**: 5433 lines across 32 files, 36 commits. Reviewed at the pattern/structure/risk level given the scope.

**What this changes**: Introduces `attend`, a new Rust binary providing a persistent sensor loop for Claude Code sessions. It implements adaptive polling, peer session discovery, inter-session signaling via filesystem, a disclosure governor rate limiter, state persistence, self-reload on binary change, and declarative config with project-scope overlay. Adds ADR-113 and ADR-114, a way, a skill, Makefile integration, and several helper scripts.

**Architecture assessment**: The overall design is solid. The Sensor trait, SensorSlot wrapper, AdaptiveInterval, DeltaAccumulator, and DisclosureGovernor compose cleanly. Zero external dependencies (no serde, no clap, no tokio) is a strong choice for a long-lived binary. The ADRs are thorough and well-reasoned, with honest treatment of prior failures (ADR-101/102) and clear invariant documentation.

---

### Findings

#### 1. Likely compile error: dead field in PeerSensor

**Location**: `tools/attend/src/sensors/peer.rs:80`

**Problem**: The `PeerSensor::new()` constructor assigns a `signals_dir` field (line 80), but the struct definition (lines 16-33) does not declare this field. This would cause a compile error. The struct uses the free function `signals_base()` for all signal directory access, never `self.signals_dir`.

**Suggestion**: Either the struct definition in the diff is stale (missing the field declaration) or the constructor has a vestigial field from a refactor. If the free function `signals_base()` is the intended pattern, remove `signals_dir` from the constructor. If the field should exist, add it to the struct.

---

#### 2. Duplicated session ID discovery logic

**Location**: `tools/attend/src/main.rs:767-808` and `tools/attend/src/sensors/peer.rs:632-662`

**Problem**: `own_session_id()` in main.rs and `find_own_session_id()` in peer.rs implement nearly identical logic: walk the process tree via repeated `ps` calls, find the claude parent PID, scan `~/.claude/sessions/*.json` for a matching PID, extract the sessionId. This is ~40 lines duplicated.

**Why it matters**: These functions shell out to `ps` up to 10 times each, and any bug fix needs to be applied in two places.

**Suggestion**: Extract to a shared utility module (e.g., `src/session.rs`) with a single `fn find_session_id(pid: u32) -> Option<String>`. Both `main.rs` and `peer.rs` can call the shared version.

---

#### 3. main.rs is a monolith at 967 lines

**Location**: `tools/attend/src/main.rs`

**Problem**: main.rs contains the DisclosureGovernor, ScheduledSensor, all subcommand handlers (`cmd_run_with_catchup`, `cmd_inbox`, `cmd_peers`, `cmd_send`, `cmd_status`, `cmd_focus`), the entry point, and multiple helper functions (`identify_sender`, `detect_terminal`, `signals_base`, `encode_project`, `own_session_id`, `read_focus_list`, `write_focus_list`, `count_signals`). At 967 lines with 7+ top-level functions and a struct, this exceeds the "suggest focused module breakdown" threshold.

**Suggestion**: Consider extracting:
- `DisclosureGovernor` and `ScheduledSensor` to `src/governor.rs` (they are already self-contained)
- Subcommands to `src/commands/` or `src/cli.rs`
- Shared helpers (`signals_base`, `encode_project`, `own_session_id`, `identify_sender`, `detect_terminal`, `read_focus_list`, `write_focus_list`) to `src/util.rs` or `src/signals.rs`

This is not blocking -- the current structure is readable -- but becomes important as the planned script sensor runner and insistence tracker are added.

---

#### 4. Sensor construction in cmd_run_with_catchup is repetitive

**Location**: `tools/attend/src/main.rs:131-175`

**Problem**: Each of the 4 built-in sensors follows the same pattern: check config for enabled, get config overrides, construct sensor, push SensorSlot with `new_with_config`, push name. This is ~10 lines per sensor, 40 total, with identical structure. When script sensors land, this will grow further.

**Suggestion**: A helper function or macro that takes `(name, sensor_constructor, config)` and returns an `Option<SensorSlot>` would reduce this to 4 one-liners and make adding sensors trivial.

---

#### 5. JSON parsing fragility in peer.rs transcript reader

**Location**: `tools/attend/src/sensors/peer.rs:320-358`

**Problem**: `read_transcript_summary` reads the entire transcript file into memory (`fs::read_to_string`) even though it only needs the last ~8KB. The comment says "Read last few KB" and `read_from` is computed, but the actual `fs::read_to_string` call reads the whole file. For long-running sessions, JSONL transcripts can grow to tens of MB. The `start_byte` logic then slices into the string, but only after the full allocation.

**Why it matters**: Performance -- reading multi-MB files every 30 seconds per peer session adds unnecessary I/O and memory pressure for a sensor that runs in the background.

**Suggestion**: Use `File::open` + `seek(SeekFrom::End(-8192))` + `BufReader` to read only the tail. This matches the comment's intent.

---

#### 6. Context window size heuristic may drift

**Location**: `tools/attend/src/sensors/peer.rs:362-378`

**Problem**: The context window size is inferred from model name strings (`"opus-4"`, `"sonnet-4"` -> 1M, else 200K). This will silently produce wrong percentages as new models ship, and model name strings in Claude Code transcripts can change format.

**Why it matters**: Wrong context percentages for peers are misleading but not dangerous -- peers' context pressure is informational. However, this same pattern could propagate to the context sensor if it were to compute its own percentages rather than relying on `ways context`.

**Suggestion**: Document this as a known limitation in a code comment. Consider adding the heuristic to config so users can override per-model. This is informational, not blocking.

---

#### 7. Security: signal file content is not validated beyond format

**Location**: `tools/attend/src/sensors/peer.rs:176-234` and `tools/attend/src/main.rs:596`

**Problem**: Signal files are written and read by any process with filesystem access. The format `from|project|cwd|message` is parsed but the `message` field is displayed directly to Claude. A malicious process (or a user on a shared system) could write crafted signal files to `~/.cache/attend/signals/` that inject misleading instructions into Claude's context.

**Why it matters**: On a single-user system this is moot -- an attacker with filesystem access can do worse things. On shared systems (unlikely for Claude Code's target use), this could be a prompt injection vector.

**Assessment**: The ADR correctly identifies "filesystem as transport" and the trust model mirrors ways' approach. The XDG cache location (`~/.cache/`) has standard user-only permissions. This is acceptable as-is but worth a brief code comment acknowledging the trust assumption.

---

#### 8. Self-reload exec() does not flush stdout

**Location**: `tools/attend/src/main.rs:257-269`

**Problem**: When the binary detects its own mtime has changed, it checkpoints state and calls `exec()` to replace itself. The state checkpoint is flushed (atomic write), but any pending stdout buffer content is not explicitly flushed before `exec()`. `exec()` replaces the process image without running destructors or flushing stdio buffers.

**Suggestion**: Add `io::stdout().flush().ok();` before the `exec()` call to ensure any buffered Monitor notifications are delivered.

---

#### 9. write_focus_list is not atomic

**Location**: `tools/attend/src/main.rs:881-887`

**Problem**: `write_focus_list` uses `std::fs::write` directly, which is not atomic. `cmd_send` reads the focus file while `cmd_focus` may be writing it. In contrast, signal files and state checkpoints correctly use tmp+rename for atomic writes.

**Suggestion**: Use the same tmp+rename pattern used elsewhere in this codebase.

---

#### 10. No tests

**Problem**: No unit tests or integration tests are included in this PR. The sensor trait, DeltaAccumulator, AdaptiveInterval, DisclosureGovernor, state serialization/deserialization, and encode_project are all pure or near-pure logic that would benefit from unit tests.

**Why it matters**: The ADR lists this as a prototype/first implementation, so the absence is understandable. But the governor rate limiting and adaptive interval math are subtle enough that tests would catch regressions as the codebase evolves.

**Suggestion**: At minimum, tests for `StateSnapshot::serialize/deserialize` round-trip, `DeltaAccumulator` accumulation + summary, `AdaptiveInterval` ramp/decay behavior, `DisclosureGovernor` window + cooldown logic, and `encode_project` / `encode_cwd` consistency.

---

### Positive observations

- **Zero-dependency Rust** is a strong architectural choice. No serde, no clap, no tokio. The binary stays small and fast.
- **Atomic writes** for signal files and state checkpoints (tmp+rename) show careful attention to concurrent access.
- **Adaptive interval design** (fast ramp, slow hysteresis decay) is well-calibrated for the use case. The tick/delta/emit module decomposition is clean.
- **Disclosure governor** rate limiting is empirically grounded (3 per 120s) and the square-root scaling of cooldown is a nice touch.
- **Sidetone prevention** by checking the `from` field rather than filename is correct -- it works across all scoped directories.
- **Shell glob expansion fence** in `cmd_send` is a thoughtful defensive measure for a CLI that will be called from both Claude and human terminals.
- **Forward-only mode** by default (marking existing signals as seen on startup) prevents startup notification storms.
- **ADR-113** is one of the strongest ADRs in this project -- thorough context, honest failure analysis of prior attempts, clear invariants with a documented revision note.

### Assessment

The architecture is sound and well-reasoned. The sensor trait design follows Open/Closed cleanly -- new sensors can be added without modifying the scheduler. The codebase is readable and the module boundaries (tick, delta, emit, sensors) have clear Single Responsibility.

**Blocking**: Item 1 (probable compile error from dead struct field). Everything else is improvement-track.

**Priority for follow-up**: Items 2 (DRY), 3 (main.rs decomposition), 5 (transcript reading performance), and 10 (tests) are the highest-value improvements for the next iteration.

Issue #2, opened 2026-04-10
attend: remaining work from ADR-113/114 implementation


aaronsb:
## Context

Branch `feat/adr-113-114-attend-active-awareness` shipped the core attend implementation across 28 commits. The sensor loop, peer messaging, focus groups, context tracking, and ways integration are all working. Two Claude instances + a human terminal successfully collaborated through attend in a live session.

This issue tracks the remaining work that emerged during implementation.

**Audit 2026-04-12**: re-checked every item against the codebase. The vast majority of what was "remaining" has since shipped across PRs #1, #3, #4, #7, #10, #11, #14, #18. Only a handful of genuinely unstarted items remain — see "Still open" below.

### Signal model / awareness

- [ ] **Build event batching** — process sensor reports cargo/rustc start+exit as individual events. A higher-level "build started → completed (pass/fail)" observation would be cleaner for peer observers. Candidate: aggregate in the process sensor or as a derived sensor.
- [ ] **Consequence model / projection** — context sensor does velocity + projection, but the general consequence model (ADR-113 `consequence/` module) isn't built. Needed for non-context consequences like "branch is diverging from main at N commits/day."
- [ ] **Insistence tracker** — unacted observations should re-surface with escalating urgency. Currently observations are fire-and-forget. ADR-113's insistence emitter + acknowledgment tracking is not started. (ADR-119 shipped the action-potential model that replaces part of this — revisit scope before starting.)
- [ ] **Emission modes** — ADR-114 defines silent/informational/affordance/insistent/critical. Currently attend has magnitude-based priority but no formal mode system. (ADR-119 may subsume this; check.)

### Operational

- [ ] **Signal TTL and cleanup subcommand** — signals older than 5 minutes are cleaned up by the peer sensor during polls, but there's no periodic cleanup for dirs nobody is polling. An `attend cleanup` subcommand or cron-based cleanup would help.
- [ ] **Way vocabulary tuning** — attend way scores 0.44 on "attend focus group" but only 0.31 on "send signal to another claude". More test prompts and threshold tuning via `/ways-tests` would improve matching.

### ADR-113
- [x] Self-reload on binary change — PR #1
- [x] State persistence / checkpoint — PR #1
- [x] Config file (ADR-115 pattern) — PR #1 / PR #8
- [x] Permissions audit implementation (`ways permissions audit`, ADR-116) — PR #10
- [x] Script sensor runner (`tools/attend/src/sensors/script.rs`) — shipped

### ADR-114
[...]

PR #3, opened 2026-04-10
Extract agent-fmt crate, lint infrastructure, peer CWD fix

aaronsb:
## Summary

- **agent-fmt crate** — extract shared terminal formatting (`Table`, `Align`, `terminal_width`, `Banner`, `print_commands`) from ways-cli into a workspace crate. Both ways and attend now use consistent table rendering, figlet banners, and help text formatting.
- **`make lint`** — add clippy with `-D warnings` across the workspace, wired into `make test`. Fix all 56 clippy findings (9 attend + 47 ways-cli).
- **Peer CWD display** — attend peer notifications and inbox now show full paths (`project_name [/full/path]`) so `--to` messages can be targeted accurately.
- **ADR-116 draft** — declarative `requires:` field for way frontmatter and attend sensor configs, replacing the obscure `trusted-project-macros` file with per-way/per-sensor permission declarations auditable against `settings.json`.

Tracking: #2

## Commits

1. `refactor:` extract agent-fmt crate from ways-cli table module
2. `feat:` use agent-fmt tables in attend inbox and status commands
3. `feat:` shared banner and help formatting in agent-fmt
4. `chore:` suppress dead_code warnings in attend
5. `chore:` add make lint target and fix clippy warnings in attend
6. `chore:` fix all clippy warnings in ways-cli (47 findings)
7. `fix:` show full CWDs in attend peer notifications and inbox
8. `docs:` ADR-116 draft — declarative permission requirements

## Test plan

- [ ] `make lint` passes clean (zero warnings)
- [ ] `make ways-rebuild && make attend-rebuild` both succeed
- [ ] `ways help` shows coral gradient banner via agent-fmt
- [ ] `attend help` shows teal gradient banner via agent-fmt
- [ ] `attend inbox` renders table with Scope/From/Message/Source columns
- [ ] `attend status` renders signals and focus tables
- [ ] Peer notifications include `[/full/cwd/path]` in messages
- [ ] Review ADR-116 design for `requires:` field schema


aaronsb:
## Code Review: PR #3 — agent-fmt extraction, lint, peer CWD, ADR-116

**PR size**: ~450 additions / ~183 deletions across 42 files. Well-scoped for what it covers.

---

### Findings

#### 1. [concern] `pub use agent_fmt as table` is a compatibility shim that should be temporary

**Location**: `tools/ways-cli/src/main.rs:10`

**Problem**: Re-exporting the entire `agent_fmt` crate under the old `table` name means every internal `crate::table::*` callsite compiles today but conceals the migration. Anyone reading `crate::table::Banner` will not realize they are using `agent_fmt`. This alias also re-exports `Banner`, `print_commands`, and the gradient constants through the `table` namespace, which is semantically wrong — banners are not tables.

**Suggestion**: The alias is fine as a transitional step in *this* PR to keep the diff focused. But it should be marked with a `// TODO: migrate callsites from crate::table to agent_fmt` comment, or better, a follow-up item. Without that marker, this will calcify.

**Severity**: concern

---

#### 2. [suggestion] `terminal_width()` doc comment claims `$COLUMNS` fallback; code does not implement it

**Location**: `tools/agent-fmt/src/table.rs:9-10` (module doc) and `table.rs:31-50` (function body)

**Problem**: The module doc says "Terminal width is auto-detected via ioctl (Unix) or $COLUMNS." But the function only does the ioctl loop and falls back to hardcoded 80. There is no `$COLUMNS` / `$TERM_COLUMNS` check. This is a doc/implementation mismatch.

**Suggestion**: Either add the env-var fallback (simple: `std::env::var("COLUMNS").ok().and_then(|v| v.parse().ok()).unwrap_or(80)`) between the ioctl loop and the final `80`, or fix the doc comment to match reality. Given this is a shared library crate, the env fallback is cheap insurance for piped/non-tty contexts.

**Severity**: suggestion

---

#### 3. [concern] `attend/src/main.rs` remains at 940 lines — flagged as priority

**Location**: `tools/attend/src/main.rs`

**Problem**: This file was already 940 lines before the PR and stays at 940. The PR adds `agent_fmt` usage into `cmd_inbox` and `cmd_status` (good) but doesn't extract any of the ~10 `cmd_*` functions that are all monolithed in this single file. `cmd_inbox` alone is ~60 lines, `cmd_status` ~50, `cmd_send` ~70, `cmd_focus` ~60. The main dispatch at the bottom is another ~80.

**Why it matters**: attend is growing — ADR-116 will add `cmd_permissions_audit`. Adding more commands to a 940-line file compounds the problem.

**Suggestion**: Not blocking for this PR, but this is the right moment to plan extraction. A `tools/attend/src/cmd/` module directory with `inbox.rs`, `status.rs`, `send.rs`, `focus.rs`, `peers.rs` would mirror the pattern ways-cli already uses. Could be a follow-up PR.

**Severity**: concern (not blocking, but flagged per quality gate)

---

#### 4. [nit] `#[allow(dead_code)]` on `PeerSummary` struct and trait methods

**Locations**:
- `tools/attend/src/sensors/peer.rs:35` — `#[allow(dead_code)]` on `PeerSummary`
- `tools/attend/src/sensors/mod.rs:57` — `#[allow(dead_code)]` on `decay_threshold()`
- `tools/attend/src/sensors/mod.rs:75` — `#[allow(dead_code)]` on `SensorSlot::new()`
- `tools/attend/src/state.rs:161` — `#[allow(dead_code)]` on `StateStore::clear()`

**Problem**: These suppress warnings rather than resolve them. `PeerSummary` having dead fields means either the struct is carrying weight that should be trimmed, or consumers that should use those fields haven't been wired up yet. `SensorSlot::new()` being dead code is more concerning — it's a constructor for a core type.

**Suggestion**: If these are planned-but-not-yet-wired (e.g., `SensorSlot::new` will be used once script sensors land per ADR-115), add a comment noting the intent: `#[allow(dead_code)] // used by ScriptSensor runner (ADR-115)`. Bare `allow(dead_code)` without context is a smell.

**Severity**: nit

---

#### 5. [suggestion] `Banner::new()` hardcodes default title `"A G E N T"`

**Location**: `tools/agent-fmt/src/banner.rs:52`

**Problem**: The default title is "A G E N T" which makes sense for *this* project but makes `agent-fmt` less reusable if it's meant to be a shared crate. There's a `.subtitle()` builder method but no `.title()` method to override the title line.

**Suggestion**: Add a `.title()` builder method. Both callers currently want "A G E N T" so no immediate behavioral change, but it rounds out the API. The `text` field controls the figlet word; `title` controls the underlined header. Being able to set both is the expected API surface.

**Severity**: suggestion

---

#### 6. [suggestion] `print_commands` and `Banner::print` write directly to stdout — no `Write` sink

**Location**: `tools/agent-fmt/src/banner.rs:68-98, 103-108` and `tools/agent-fmt/src/table.rs:122-197`

**Problem**: All output goes directly to `println!`. This makes the formatters untestable without capturing stdout. For a shared library crate, accepting a `&mut dyn Write` or returning a `String` is more composable. `Table::print()` in particular does a lot of layout math that would benefit from unit tests, but currently there's no way to assert on its output without redirect tricks.

**Suggestion**: Consider adding `Table::render(&self) -> String` alongside `print()`, and similarly `Banner::render(&self) -> String`. The `print()` methods can just call `render()` and write. This unblocks unit testing for the formatting logic, which is the most complex part of agent-fmt.

**Severity**: suggestion

---

#### 7. [nit] `#[allow(clippy::too_many_arguments)]` on `lint_file` and render functions

**Locations**:
- `tools/ways-cli/src/cmd/lint.rs:222`
- `tools/ways-cli/src/cmd/render.rs:148, 155`

**Problem**: Suppressing `too_many_arguments` is pragmatic for this PR (it's a clippy fix pass), but these functions genuinely have too many arguments. `lint_file` takes 8 parameters; `write_way_row_with` takes 8+. That's a signal that a config/context struct would help.

**Suggestion**: Not for this PR, but a `LintContext` struct or `RenderContext` struct would clean these up and make the functions easier to extend without growing the signature.

**Severity**: nit

---

#### 8. [blocker-ish] ADR-116 — permission format relies on string matching, but `settings.json` patterns use glob semantics

**Location**: `docs/architecture/system/ADR-116-declarative-permission-requirements.md:163-174`

**Problem**: The ADR says "the audit is a direct string comparison — no translation layer needed." But `settings.json` permissions use glob patterns like `Bash(git:*)` where `*` has wildcard semantics. A way declaring `requires: ["Bash(git:*)"]` and settings.json granting `Bash(git:*)` works by string equality, but what about:

- Way requires `Bash(git:status)` — does `Bash(git:*)` in settings.json satisfy it? String comparison says no.
- Way requires `Bash(gh:pr create)` — is that one permission or should it be `Bash(gh:*)`?
- Settings.json grants `Bash(*)` — does that satisfy `Bash(git:*)`? String comparison says no.

The ADR needs to specify the matching semantics: is `requires` comparison exact string equality, or does it respect the glob hierarchy? If exact, the requires values must mirror settings.json entries verbatim, which is brittle. If glob-aware, you need the "translation layer" the ADR says you don't need.

**Suggestion**: Define matching as: a `requires` entry is satisfied if any `permissions.allow` entry in settings.json *covers* it. `Bash(*)` covers `Bash(git:*)` covers `Bash(git:status)`. This is a containment check, not string equality. Document this explicitly — it's the key design decision.

**Severity**: concern (calling it blocker-ish because this ambiguity will cause confusion during implementation)

---

#### 9. [suggestion] ADR-116 — no `requires` validation at lint time

**Location**: `docs/architecture/system/ADR-116-declarative-permission-requirements.md` (Implementation Plan)

**Problem**: The implementation plan adds `requires:` to frontmatter-schema.yaml and builds audit commands, but doesn't mention validating `requires` values during `ways lint`. Invalid requires entries (typos like `Bash(gi:*)` or non-existent tool names) would silently pass through until someone runs the audit.

**Suggestion**: Add a lint rule: `requires` values must match the documented permission format (`Read`, `Write`, `Edit`, `Bash(...)`, `*`). This catches typos early and is trivial to implement — it's a regex on each entry.

**Severity**: suggestion

---

#### 10. [nit] Peer CWD display may leak long paths in tight terminals

**Location**: `tools/attend/src/sensors/peer.rs:437-438`

Format strings like `"peer session started: {} [{}] ({}, {}, ctx {:.0}%)"` with full CWD paths can produce very long notification lines. The attend monitor delivers these as single-line notifications. A path like `/home/user/projects/some-very-long-project-name` makes these hard to parse visually.

**Suggestion**: Consider truncating the CWD to basename + parent, or using the table formatter for peer notifications too. Minor since the inbox already uses the Table which handles truncation.

**Severity**: nit

---

### What looks solid

- **The extraction itself is clean.** `table.rs` moved with minimal diff (just adding `is_empty`). `banner.rs` consolidates duplicated figlet/gradient code from two places into one. The font assets moved with the code. Dependency graph is right — `agent-fmt` is leaf, both binaries depend on it.
- **Clippy fixes are all mechanical and correct.** `strip_prefix`/`strip_suffix` replacing manual slicing, `is_some_and`/`is_none_or` replacing `map_or`, `saturating_sub` replacing manual bounds checks, `keys()` replacing `for (k, _)`, `.clamp()` replacing `.max().min()`. Every one of these is a strict improvement.
- **Lint infrastructure is well-placed.** `make lint` before `make test` catches warnings before running tests. `-D warnings` means CI will catch regressions.
- **The `pub use agent_fmt as table` alias** is a pragmatic way to avoid a massive rename diff in this PR. It keeps the PR focused.

### Summary

The agent-fmt extraction and clippy cleanup are solid work. The main technical debt items are the attend monolith (940 lines, pre-existing) and the stdout-coupled API surface (testability). ADR-116 needs the matching semantics clarified before implementation — that's the one item I'd want resolved before this moves forward.

---
*Review by Claude (code-reviewer agent)*

PR #4, opened 2026-04-11
ADR-114 attend-to-ways integration, ScriptSensor runner

aaronsb:
## Summary

- **ADR-114 implementation** — `ways show attend <signal>` command that looks up ways declaring `trigger.type: attend` with matching signal types, runs through the ADR-104 disclosure gate. First signal handler: `meta/attend/context-pressure` for reflection guidance as context approaches compaction.
- **Affordance strings** — high-salience context notifications (85%+) now include `ways show attend context-pressure` hint so Claude can request deeper guidance.
- **Schema** — `trigger.type: attend` + `signals` field added to `frontmatter-schema.yaml`.
- **Linter** — understands attend signal handlers: shows INFO instead of description/vocabulary warning, validates `signals` field is present.
- **ScriptSensor runner** — shell scripts can now be registered as attend sensors via config (`+sensor-name: { script: path, interval: N }`). Script contract: stdout lines as `magnitude|description`, exit 0. This is the extensibility path for attend.
- **Makefile fix** — `ways-rebuild` now uses workspace manifest so the symlink points to the correct binary.
- **Housekeeping** — removed completed todo files from ADR-110/111.

Tracking: #2

## Commits

1. `feat:` implement ADR-114 attend-to-ways integration
2. `docs:` promote ADR-114 from Draft to Accepted
3. `feat:` linter recognizes attend signal handlers, fix ways build path
4. `chore:` remove completed todo files from ADR-110 and ADR-111
5. `feat:` implement ScriptSensor runner for attend

## Test plan

- [ ] `make lint` passes clean
- [ ] `ways show attend context-pressure --session test` returns reflection guidance
- [ ] `ways show attend nonexistent --session test` returns "no way handles" message
- [ ] `ways lint --check --global` shows INFO for attend way, 0 errors, 0 warnings
- [ ] ScriptSensor: config with `+test: { script: /tmp/test.sh, interval: 60 }` registers and polls
- [ ] `make ways-rebuild` symlinks to `tools/target/release/ways` (not `tools/ways-cli/target/`)
- [ ] Review ADR-114 Accepted status


aaronsb:
## Code Review: ADR-114 attend-to-ways integration + ScriptSensor

**5 commits, ~330 lines added across Rust (attend + ways-cli), YAML schema, one new way file, Makefile fix, todo cleanup.**

---

### 1. ScriptSensor (`tools/attend/src/sensors/script.rs`) -- Missing timeout guard

**Severity: concern**

`Command::new("bash").arg(&script_path).output()` is a **blocking call with no timeout**. A script that hangs (waiting on network, a stuck pipe, an infinite loop) blocks the entire attend tick loop indefinitely -- no other sensor can poll until it returns.

Every other sensor in attend does CPU-local work (read files, parse JSON, check processes). ScriptSensor breaks that assumption by shelling out to arbitrary user scripts.

**Suggestion**: Add a timeout. The simplest approach without pulling in async:

```rust
use std::process::Stdio;

let mut child = Command::new("bash")
    .arg(&script_path)
    .current_dir(&self.working_dir)
    .stdout(Stdio::piped())
    .stderr(Stdio::null())
    .spawn()?;

// Use the sensor's base_interval as a generous upper bound
let timeout = self.base_interval;
match child.wait_timeout(timeout) {  // needs wait-timeout crate, or manual poll
    Ok(Some(status)) => { /* read stdout */ }
    Ok(None) => {
        child.kill().ok();
        eprintln!("[attend] script sensor '{}': timed out after {:?}", self.name, timeout);
        return Vec::new();
    }
    Err(e) => { /* handle */ }
}
```

Alternatively, `std::process::Child::try_wait()` in a loop with a bounded iteration count would work without a new dependency.

---

### 2. ScriptSensor -- stderr is silently swallowed

**Severity: nit**

`.stderr(Stdio::null())` means if a script fails with a useful error message, the operator gets nothing. Since attend already logs to stderr with `eprintln!`, consider capturing stderr on non-zero exit and logging it (truncated) for debuggability:

```rust
if !output.status.success() {
    let stderr = String::from_utf8_lossy(&output.stderr);
    if !stderr.is_empty() {
        eprintln!("[attend] script sensor '{}': exit {}, stderr: {}",
            self.name, output.status, &stderr[..stderr.len().min(200)]);
    }
    return Vec::new();
}
```

---

### 3. ScriptSensor -- no magnitude validation

**Severity: suggestion**

Parsed magnitudes from script output are not validated. A script outputting `999.0|something happened` or `-5.0|negative event` passes through. Other built-in sensors carefully bound their magnitudes (e.g., context sensor maps to 0.3--0.9 range). An unbounded magnitude from a script sensor could dominate the delta accumulator and starve other sensors.

**Suggestion**: Clamp to a reasonable range, or at minimum document the expected range in the contract comment at the top of the file.

---

### 4. `extract_attend_signals` frontmatter parsing (`tools/ways-cli/src/cmd/show/helpers.rs:113-152`)

**Severity: concern**

The parser looks for `type: attend` anywhere in frontmatter, but this is a nested field (`trigger.type: attend`). If a future way ever has a top-level `type: attend` for a different purpose, or if the `type` field appears under a different parent, this would false-match.

Currently the context-pressure way has it nested:
```yaml
trigger:
  type: attend
  signals:
    - context-pressure
```

The parser works today because `line.trim() == "type: attend"` matches the indented line. But it would also match a hypothetical `type: attend` at the top level. The same concern applies to the linter's `is_attend` check at `lint.rs:287`.

This is low-risk given the controlled schema, but worth noting. If you ever want to harden it, tracking YAML indent depth (2-space for nested under `trigger:`) would disambiguate.

---

### 5. `attend()` in `show/mod.rs:301-362` -- all-files scan on every invocation

**Severity: suggestion**

Every call to `ways show attend <signal>` does a full `walkdir` + `read_to_string` + frontmatter parse of every `.md` file in the ways tree. With ~85+ ways this is fast enough, but the pattern of "scan everything, match by content" could be avoided since attend ways live in a known subtree (`meta/attend/`).

A targeted scan of just the attend directory would be more efficient and self-documenting. Not blocking, but worth considering if more signal types are added.

---

### 6. Context sensor affordance string (`tools/attend/src/sensors/context.rs:167-170`)

**Severity: suggestion**

The hardcoded `ways show attend context-pressure --session $CLAUDE_SESSION_ID` string embeds a specific signal name in the sensor. If the way is renamed or additional context-related signals are added, the sensor needs a code change.

This is pragmatic for a first implementation. If you want to decouple, the sensor could emit a generic `attend:context-pressure` signal type and let the ways layer resolve it. But that's a bigger refactor -- fine to defer.

---

### 7. No tests for new code

**Severity: concern**

Neither `ScriptSensor` nor `extract_attend_signals` nor `normalize_way_id` have unit tests. All three have testable logic:

- `extract_attend_signals` -- edge cases: no frontmatter, signals field with inline values (`signals: [a, b]`), empty signals list, `type: attend` without signals
- `normalize_way_id` -- `"a/b/b"` collapses, `"a/b/c"` doesn't, single segment, empty string
- `ScriptSensor::resolve_script` -- absolute vs relative paths

`normalize_way_id` and `extract_attend_signals` are pure functions -- trivial to unit test. Worth adding before merge.

---

### 8. Makefile build path fix

**Severity: positive observation**

The `Makefile` change from `tools/ways-cli/Cargo.toml` to `tools/Cargo.toml -p ways` correctly reflects the workspace structure. Both `ways` and `ways-rebuild` targets are updated consistently. Clean.

---

### 9. Linter attend-awareness (`tools/ways-cli/src/cmd/lint.rs:286-310`)

**Severity: positive observation**

The linter correctly:
- Exempts attend ways from the vocabulary requirement (since they're matched by signal name, not semantic embedding)
- Validates that `signals:` is present when `type: attend`
- Emits an INFO line so the operator sees attend ways are recognized

The `is_attend` check at line 287 uses the raw frontmatter string, consistent with how other linter checks work. Good integration.

---

### 10. Security: script sensors run arbitrary shell scripts

**Severity: suggestion (documentation)**

The trust model for script sensors relies on the attend config being user-controlled (`~/.claude/attend.toml` or similar). This is reasonable -- the user opts in by declaring a script path. However, the trust boundary isn't documented anywhere in the codebase.

Consider adding a brief comment in `script.rs` or the attend config docs noting: "Script sensors execute user-configured shell scripts. The trust boundary is the attend configuration file itself -- only scripts explicitly declared by the user are executed."

---

### 11. `main.rs` at 955 lines (pre-existing)

**Severity: nit (pre-existing)**

`tools/attend/src/main.rs` is at 955 lines, above the 800-line priority threshold. This PR adds ~15 lines to it (ScriptSensor registration). Not caused by this PR, but the file continues to grow. The `cmd_run_with_catchup` function in particular is doing a lot: config loading, sensor construction, state restoration, tick loop, signal handling. Worth splitting in a future PR.

---

### Summary

The core design is solid -- attend signals route through the existing way disclosure pipeline (session markers, redisclosure gates, scope checks), which means all the ADR-104 machinery applies automatically. The ScriptSensor is clean and follows the Sensor trait contract.

**Before merge, I'd want to see:**
1. A timeout mechanism on `ScriptSensor::poll()` (the blocking call is the main risk)
2. Unit tests for `extract_attend_signals` and `normalize_way_id`

**Nice to have but not blocking:**
3. Magnitude clamping or documented range for script output
4. Trust model comment in `script.rs`

PR #6, opened 2026-04-11
Fix: validate attend send --to paths and show peer table

### LINKED ISSUE #5: attend send --to should validate peer paths and suggest corrections
## Problem

`attend send --to /path/to/project` requires knowing the exact filesystem path of the target peer. This is fragile — during testing, a session sent to `/home/aaron/Projects/game-ai-pro` when the actual path was `/home/aaron/Projects/data/game-ai-pro`. The signal was silently written to a nonexistent peer directory with no error.

Additionally, `attend peers` only shows a count (e.g., "8 sessions (0 in this project)") without listing the actual project paths, so there's no way to discover the correct path.

## Proposed behavior

1. **`attend send --to` should validate** the target path against known active peers
2. **On validation failure**, return an error with the list of valid peer paths
3. **Fuzzy suggestion** when a close match exists (e.g., "did you mean: /home/aaron/Projects/data/game-ai-pro")
4. **`attend peers` should list project paths**, not just counts

Example expected UX:
```
$ attend send --to /home/aaron/Projects/game-ai-pro "hey"
error: no peer found at /home/aaron/Projects/game-ai-pro

active peers:
  /home/aaron/Projects/data/game-ai-pro
  /home/aaron/temp
  /home/aaron/.claude

did you mean: /home/aaron/Projects/data/game-ai-pro?
```

## Context

Discovered during a multi-session attend test between three Claude Code sessions (game-ai-pro, temp, .claude). The temp session sent a directed message to the wrong path and it silently failed — no delivery, no error.

aaronsb:
## Summary

- **`attend send --to` validation** — target path is validated against active peers before sending. On mismatch: error message, full list of active peer paths, and fuzzy suggestion of closest match.
- **`attend peers` table** — now shows Project, Path, Status, Context columns using agent-fmt Table instead of just a count.

Fixes #5

## Example

```
$ attend send --to /home/aaron/Projects/game-ai-pro "hey"
error: no active peer at /home/aaron/Projects/game-ai-pro

active peers:
  /home/aaron/Projects/data/game-ai-pro (game-ai-pro)
  /home/aaron/temp (temp)

did you mean: /home/aaron/Projects/data/game-ai-pro?
```

## Test plan

- [ ] `make lint` passes
- [ ] `attend peers` shows table with paths
- [ ] `attend send --to /bad/path "msg"` errors with peer list and suggestion
- [ ] `attend send --to /valid/peer/path "msg"` succeeds normally
- [ ] `attend send "msg"` (no --to) still works unchanged
- [ ] `attend send --broadcast "msg"` still works unchanged

### DIFF
--- a/tools/attend/src/main.rs
+++ b/tools/attend/src/main.rs
@@ -608,6 +608,19 @@ fn cmd_send(args: &[String]) {
                     eprintln!("\ndid you mean: {}?", suggestion);
                 }
             }
+            // Show focus group if configured
+            let focus_file = base.join("focus");
+            if let Ok(content) = std::fs::read_to_string(&focus_file) {
+                let groups: Vec<&str> = content.lines()
+                    .filter(|l| !l.trim().is_empty())
+                    .collect();
+                if !groups.is_empty() {
+                    eprintln!("\nfocus group (use `attend send` without --to to reach all):");
+                    for g in &groups {
+                        eprintln!("  {}", g);
+                    }
+                }
+            }
             std::process::exit(1);
         }
     }
--- a/tools/attend/src/main.rs
+++ b/tools/attend/src/main.rs
@@ -5,7 +5,7 @@ mod delta;
 mod emit;
 mod sensors;
 
-use sensors::{ContextSensor, Focus, GitSensor, PeerSensor, ProcessSensor, ScriptSensor, Sensor, SensorSlot};
+use sensors::{ContextSensor, Focus, GitSensor, PeerSensor, ProcessSensor, ScriptSensor, SensorSlot};
 use std::collections::BinaryHeap;
 use std::time::{Duration, Instant};
 
@@ -520,17 +520,21 @@ fn cmd_inbox() {
 }
 
 fn cmd_peers() {
-    let focus = Focus::default_focus();
-    let mut sensor = PeerSensor::new();
-    let observations = sensor.poll(&focus);
-    // First poll returns empty (baseline). The baseline info goes to stderr.
-    // Poll again to get the actual state — but since nothing changed, just
-    // show the baseline that was printed to stderr.
-    if observations.is_empty() {
-        // Baseline was printed to stderr by the sensor. For the CLI, also
-        // list what we found by doing a second poll (which will show no deltas).
-        let _ = sensor.poll(&focus);
+    let sensor = PeerSensor::new();
+    let peers = sensor.list_peers();
+
+    if peers.is_empty() {
+        println!("no active peer sessions");
+        return;
     }
+
+    let mut t = agent_fmt::Table::new(&["Project", "Path", "Status", "Context"]);
+    t.max_width(0, 20);
+    for (cwd, project, status, ctx) in &peers {
+        t.add(vec![project, cwd, status, &format!("{ctx:.0}%")]);
+    }
+    t.print();
+    println!("  {} peer(s)", peers.len());
 }
 
 fn cmd_send(args: &[String]) {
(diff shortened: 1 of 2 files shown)


aaronsb:
## Code Review — PR #6: attend send --to path validation

Small, well-scoped PR. The core idea is sound. A few issues worth addressing before merge.

---

## Issue: Correctness — `list_peers` takes `&self` but `discover_peers` may implicitly rely on mutable state

**Location**: `tools/attend/src/sensors/peer.rs:90`

**Problem**: `list_peers` takes `&self`, which is fine since `discover_peers` is also `&self`. However `discover_peers` reads from the filesystem directly — it does not use or update `self.prior`. This means `list_peers` returns a snapshot that is **independent of any baseline**. That's actually the right behavior here, but it diverges from how `poll()` works (which compares against `self.prior` and surfaces only deltas). The asymmetry is not documented. If a caller expected `list_peers` to return only \"active\" peers as the polling loop defines them, they'd get different results.

**Severity**: Low — currently correct, but worth a doc comment noting the distinction.

**Suggestion**: Add a doc comment to `list_peers` clarifying it returns the raw filesystem snapshot regardless of poll baseline state.

---

## Issue: Edge case — `canonicalize` silently falls back to the raw input on non-existent paths

**Location**: `tools/attend/src/main.rs:587-590`

```rust
let resolved = std::fs::canonicalize(path)
    .map(|p| p.to_string_lossy().to_string())
    .unwrap_or_else(|_| path.clone());
```

**Problem**: If the user passes a path that does not exist (typo, unmounted volume, etc.), `canonicalize` returns an error and the code falls back to the raw input string. The raw input is then compared against canonicalized peer paths from `discover_peers`, which will **never match** because peer paths go through `sf.cwd` from the session JSON (already absolute) but are not re-canonicalized. So a path like `/home/aaron/../aaron/Projects/foo` would fail to validate even if valid. Conversely, a non-existent path gets a clean error path for the right reason, but the error message says \"no active peer at /bad/path\" rather than \"path does not exist\" — which is slightly misleading.

**Severity**: Low — functional, but the error message conflates two different failure modes.

**Suggestion**: Distinguish the two cases. If `canonicalize` fails with `NotFound`, emit \"path does not exist\" before the peer list. If it fails for other reasons (permissions), note that too. This avoids user confusion.

---

## Issue: Performance — peer discovery runs twice for `--to` sends

**Location**: `tools/attend/src/main.rs:593` and then implicitly again later in the send path

**Problem**: As noted in the review request, validation calls `sensor.list_peers()` which calls `discover_peers()` (filesystem scan + process liveness checks via `pid_is_claude`). The subsequent send path likely does not re-use this result. For a tool that scans `~/.claude/sessions/*.json` and checks PIDs, this is two filesystem traversals and multiple `kill -0` or `/proc` checks. For a CLI invoked interactively, it is not a real problem. If `attend` ever becomes hot-path or is called from a script in a loop, it would be worth threading the peer list through.

**Severity**: Negligible for current use case. Worth a `// TODO:` comment if the send path ever grows.

---

## Issue: Fuzzy matching — suffix comparison is reversed

**Location**: `tools/attend/src/main.rs:808-828`

```rust
let target_parts: Vec<&str> = target.rsplit('/').collect();
// rsplit('/') on "/home/aaron/Projects/foo" → ["foo", "Projects", "aaron", "home", ""]
```

**Problem**: `rsplit('/')` iterates from the end of the string, so `target_parts[0]` is `"foo"` (the last component) and `target_parts[1]` is `"Projects"`, etc. The `zip` then compares `target_parts[0]` against `peer_parts[0]`, which is also the last component of the peer path. This correctly matches from the right-most path segment inward — so the logic works. However, the variable is named `target_parts` without any comment, and the reversed iteration is non-obvious. A future reader unfamiliar with `rsplit` might assume it compares from the left.

**Severity**: Low — correct behavior, readability concern only.

**Suggestion**: Add a comment: `// compare from rightmost segment inward so "foo" matches "…/Projects/foo"`.

Also: the function returns `Option<&'a str>` with a lifetime tied to the `peers` slice. The call site passes `&peer_paths` (a `Vec<&str>` of borrows into `peers`). This is fine, but if the function signature is ever refactored the lifetime coupling could be surprising. Not a current bug.

---

## Issue: UX — column order in `attend peers` table

**Location**: `tools/attend/src/main.rs:539`

```rust
let mut t = agent_fmt::Table::new(&["Project", "Path", "Status", "Context"]);
t.add(vec![project, cwd, status, &format!("{ctx:.0}%")]);
```

**Problem**: The headers say `["Project", "Path", ...]` but `list_peers` returns `(cwd, project_name, ...)`. The `add` call uses `project` and `cwd` in the right order — so the output is correct. But the tuple destructuring `(cwd, project, status, ctx)` on the line just above maps positionally and relies on the reader keeping track of which is which. This is fine at 4 fields; it becomes fragile if the tuple grows.

**Severity**: Negligible — no bug, minor fragility.

---

## What looks good

- The validation is placed before the actual send, so no partial sends to invalid targets. Correct ordering.
- Exiting with code 1 on validation failure is correct for CLI tooling.
- `list_peers` being a pure read on `&self` (no mutation) is the right API shape for validation callers that do not want to disturb the poll baseline.
- The `cmd_peers` rewrite is a clear improvement — the old double-poll workaround to get output is gone, replaced with a direct call. Simpler and more predictable.
- Sorting by `cwd` in `list_peers` gives deterministic output.

---

## Summary

No blocking issues. The validation logic is correct for the happy path. The two items worth a follow-up before merge (or filed as issues) are the canonicalize failure mode conflation (misleading error message) and the fuzzy match comment for readability. Everything else is low/negligible.

PR #7, opened 2026-04-11
Attend CI workflow, signal handler ways, build affordances

aaronsb:
## Summary

- **CI workflow** — `.github/workflows/build-attend.yml` with 4-platform builds, clippy lint, artifact collection, and tag-based GitHub Releases (`attend-vX.Y.Z`). Modeled on the ways CI workflow.
- **Signal handler ways** — `build-complete` and `reflection-overdue` join `context-pressure` as attend signal handlers (ADR-114). Total: 3 attend ways, all recognized by the linter.
- **Build affordances** — process sensor now emits `ways show attend build-complete` hint when build tools (cargo, make, gcc, npm, etc.) exit.
- **Settings.json permissions** — expanded allow list for autonomous session operation (mkdir, cargo, git, make, gh, attend, ways, ps, python3, etc.).

Tracking: #2

## Test plan

- [ ] `make lint` passes
- [ ] `ways lint --check --global` shows 3 attend INFO lines, 0 errors, 0 warnings
- [ ] `ways show attend build-complete --session test` returns guidance
- [ ] `ways show attend reflection-overdue --session test` returns guidance
- [ ] CI workflow triggers on push (verify after merge)
- [ ] Build tool exit notifications include affordance string


aaronsb:
## Code Review — PR #7

**What this changes**: CI pipeline for attend binary, two new attend signal handler ways, build tool detection in the process sensor, and expanded settings.json permissions.

---

### CI Workflow (build-attend.yml)

**[INFO] No `cargo test` job.** `build-ways.yml` has a `test` job (unit tests + session_sim integration tests) and `collect` gates on `[build, test]`. `build-attend.yml` drops this entirely — `collect` gates on `[build, lint]` only. If attend has any tests, they're not running in CI. If it genuinely has no tests, that's worth a comment in the workflow or a follow-up issue.

**[INFO] Smoke test comment omitted.** `build-ways.yml` has a clarifying comment above the smoke test step ("Native builds can run directly; cross-compiled binaries can only be verified with `file`"). The attend workflow omits it. Minor, but it was clearly intentional documentation in the original.

**[OK]** Structure, matrix, cross-compilation strategy, release job, and tag naming convention (`attend-vX.Y.Z`) are all consistent with `build-ways.yml`. The `workflow_dispatch` trigger is present. Release notes include a build-from-source hint. Nothing structurally broken.

---

### Signal Handler Ways

**build-complete.md**: Guidance is appropriate and well-scoped. The "Only engage if the build is relevant to current work" note is exactly the right guardrail — prevents noisy interruption on background activity. Point 3 (note warnings) is actionable. **[OK]**

**reflection-overdue.md**: Content is strong. The framing — "window closing," explicit preference for three honest sentences over a page of status — is calibrated well for the scenario. The four capture prompts cover decisions, failures, state, and memory hand-off. **[OK]**

Both ways correctly use `type: attend` trigger and `scope: agent`. No frontmatter issues visible.

---

### Process Sensor Build Detection

**[WARN] `app.contains(t)` matching is loose.** The check does substring matching on the process name. This means a process named `typescript-server` would match `tsc`, and `nvm` or `yarn-berry` could behave unexpectedly. The build_tools list also lacks `cmake`, `ninja`, `meson`, `mvn`, `gradle`, `pip`, `bundle`, `bazel` — common enough that omitting them is a real gap for non-Rust/JS stacks. Not a blocker, but the matching strategy should probably be exact (`app == t` or `app.starts_with(t)`) to avoid false positives, with the list expanded incrementally.

**[INFO] Hardcoded `$CLAUDE_SESSION_ID` in the affordance string.** The format string emits `$CLAUDE_SESSION_ID` as a literal — it won't be shell-expanded at emit time (this is Rust, not a shell). Whoever invokes the suggestion needs to expand it themselves. That may be intentional (the consumer runs it in a shell), but if attend emits this to stdout and Claude Code reads it literally, the session ID won't resolve. Worth verifying the intended consumption path.

**[OK]** The priority bump (2.5 vs 2.0) for build exits is reasonable. The observation format matches the existing style.

---

### Settings.json Permissions

**[WARN] `Bash(rm:*)` and `Bash(chmod:*)` are broad.** `rm:*` allows deletion of any path without qualification. `chmod:*` allows permission changes on any file. These are the two entries I'd flag for scrutiny — both have destructive potential if Claude acts under incorrect assumptions. The rest of the list (mkdir, cp, mv, ln, touch, make, cargo, git, gh, attend, ways, ps, sha256sum, python3, env, id, date, sleep, command) are reasonable for autonomous session operation and align with the stated purpose.

**[INFO] `Bash(python3:*)` is open-ended.** Allows arbitrary Python execution. Acceptable for a trusted local config but worth noting if this file is shared across projects.

**[INFO] `Bash(git:*)` and `Bash(gh:*)` without path restriction allow destructive git operations (force push, branch delete) and GitHub mutations (PR merge, repo settings). Expected for this workflow, but document the trust model if this settings.json is used as a template.

---

### Missing / Should Be Split Out

**[INFO]** The `$CLAUDE_SESSION_ID` affordance string in process.rs is functionally coupled to how attend output is consumed by Claude Code. No documentation or ADR note explains this contract. If the consumption model changes (e.g., attend emits structured signals rather than text), this string becomes dead. A brief note in the sensor or a comment in the way linking them would help.

**[INFO]** No test for the build tool detection path in process.rs. The substring matching bug noted above would be caught by a unit test with a few process name fixtures.

---

### Summary

| Area | Status |
|------|--------|
| CI workflow structure | OK — minor test job gap |
| build-complete way | OK |
| reflection-overdue way | OK |
| Process sensor detection | Warn — loose matching, literal env var |
| Settings permissions | Warn — rm/chmod breadth |

No blocking issues. The test job gap and `app.contains(t)` matching are the two things worth addressing before or after merge.

---
*AI-assisted review — Claude Code*

PR #8, opened 2026-04-11
Declarative config module for ways-cli

aaronsb:
## Summary

- **Config module** — `config::Config` with layered resolution: built-in defaults → `~/.claude/ways.json` (legacy) → `$XDG_CONFIG_HOME/ways/config.yaml` → `$PROJECT/.claude/ways.yaml`. Mirrors attend's config pattern (ADR-115).
- **`config::global()`** — `LazyLock`-based accessor. Each call site is annotated with `// config::global() — future migration: ctx.config.field` as grep targets for the daemon/RequestContext refactor.
- **5 direct ways.json reads replaced**: bm25_threshold default, default_scope, parent_threshold_multiplier, domain_disabled, configured engine, resolve_language, disabled domains.
- **`ways config init/show/path`** — subcommands matching attend's pattern.
- **Net -7 lines** — centralized config is more concise than scattered reads.

Tracking: #2

## Design direction

`config::global()` is option C from a design discussion about config threading:
- Config is process-wide and immutable → global static is appropriate
- Session/project state remains in function signatures → explicit
- Migration to daemon: grep for `config::global()`, replace with `ctx.config`, bundle params into `RequestContext`

## Test plan

- [ ] `make lint` passes
- [ ] `cargo test -p ways` — all 10 simulation tests pass
- [ ] `ways config path` shows user/legacy/project paths
- [ ] `ways config show` shows resolved config with language from ways.json
- [ ] `ways config init` creates XDG config file
- [ ] `ways status` still works (engine, disabled domains from config)
- [ ] Way matching behavior unchanged (thresholds, scopes from config)

### DIFF
--- a/docs/architecture/system/ADR-117-sensor-crate-extraction-and-feature-flags.md
+++ b/docs/architecture/system/ADR-117-sensor-crate-extraction-and-feature-flags.md
@@ -1,0 +1,140 @@
+---
+status: Draft
+date: 2026-04-10
+deciders:
+  - aaronsb
+  - claude
+related:
+  - ADR-113
+  - ADR-115
+---
+
+# ADR-117: Sensor Crate Extraction and Feature Flags
+
+## Context
+
+attend (ADR-113) ships with four built-in sensors: context, git, peers, and processes. These are compiled directly into the attend binary as modules under `src/sensors/`. The ScriptSensor runner (shipped in PR #4) provides an extensibility path for shell-script sensors, but built-in sensors have no separation from the orchestrator.
+
+This creates several problems:
+
+- **No isolation** — sensor code shares the same module tree as the tick loop, state management, and CLI. Changes to one sensor can break another through shared internal interfaces.
+- **No selective compilation** — a user who only wants git and context awareness still compiles peer discovery and process scanning. On constrained systems or custom builds, this matters.
+- **No clear trait contract** — the `Sensor` trait lives in `sensors/mod.rs` alongside the `SensorSlot` runtime scaffolding. The boundary between "what a sensor must implement" and "how the orchestrator runs it" is blurred.
+- **Future daemon model** — if attend becomes long-running (paralleling the direction noted for ways-cli), hot-reloading sensors or dynamically enabling them requires a cleaner separation than in-binary modules.
+
+Meanwhile, the workspace already demonstrates the crate extraction pattern: `agent-fmt` was extracted from ways-cli (PR #3) and is consumed by both tools. The same pattern applies here.
+
+## Decision
+
+Extract each built-in sensor into its own workspace crate. Introduce a `sensor-trait` crate that defines the `Sensor` trait, `Focus` struct, and supporting types. Wire sensors into attend via Cargo feature flags.
+
+### Workspace Structure
+
+```
+tools/
+  agent-fmt/           # shared terminal formatting
+  sensor-trait/        # Sensor trait, Focus, SensorSlot
+  sensor-git/          # GitSensor
+  sensor-context/      # ContextSensor
+  sensor-peers/        # PeerSensor
+  sensor-processes/    # ProcessSensor
+  attend/              # orchestrator — depends on sensor crates via features
+  ways-cli/            # ways CLI
+```
+
+### Trait Crate
+
+`sensor-trait` defines the contract between attend and any sensor:
+
+```rust
+pub trait Sensor: Send {
+    fn name(&self) -> &str;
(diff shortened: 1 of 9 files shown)


aaronsb (commented):
## Code Review — PR #8: Declarative Config Module

### Summary

The design is sound. LazyLock is appropriate, layered resolution is clean, and the migration comments serve their stated purpose. Three issues need attention before merge, one of which is a silent behavior change.

---

### Issue 1: BREAKING — ways.json engine field name mismatch [BLOCK]

**Location**: `tools/ways-cli/src/config.rs:102`

**Problem**: The legacy `apply_ways_json` reads `v.get("engine")`, but the old `read_ways_json_engine` in `status.rs` (now deleted) read `v["semantic_engine"]`. The actual `~/.claude/ways.json` on disk has neither field — it only has `disabled` and `output_language` — so no user is broken today, but any user who had manually set `semantic_engine` in their ways.json will silently lose that setting. The unit test at line 211 also uses `"engine"` not `"semantic_engine"`, so the test doesn't catch this divergence.

**Why it matters**: Silent silent-corruption of user config is the worst failure mode. The field rename is a breaking change to the undocumented-but-real schema.

**Suggestion**: Either read both keys (preferring the new one), or document the rename in a migration note. At minimum, update the test to include a `semantic_engine` key and assert it is not silently dropped.

```rust
// apply_ways_json: handle both names
if let Some(engine) = v.get("semantic_engine").or_else(|| v.get("engine")).and_then(|v| v.as_str()) {
    self.matcher = engine.to_string();
}
```

---

### Issue 2: Dead code — `_ways_json_language` [WARN]

**Location**: `tools/ways-cli/src/agents/mod.rs:48`

**Problem**: `_ways_json_language` is prefixed with `_` to suppress the unused-function warning, but the function body still opens and parses ways.json. It is never called. The comment says "Kept for callers outside the config chain" but there are no such callers.

**Why it matters**: Dead code with live file I/O in its body creates confusion about what is actually reading configuration files. It will accumulate drift as the config schema evolves.

**Suggestion**: Delete it. The migration comment on `resolve_language` is sufficient documentation. If a future caller needs raw ways.json language, they can use `config::global().language` or reintroduce the function with a clear purpose.

---

### Issue 3: `corpus_stale_secs` is inert [WARN]

**Location**: `tools/ways-cli/src/config.rs:41, 53, 132`

**Problem**: `corpus_stale_secs` is defined, defaulted, YAML-parsed, and documented in the generated config template — but is not consumed anywhere in the codebase. A user who sets this field will get no effect.

**Why it matters**: Config fields that are parsed but not actioned erode trust in the config system. A user debugging corpus freshness issues who sets this value will be confused when it does nothing.

**Suggestion**: Either wire it to the corpus staleness check (the appropriate place in scan or rebuild logic), or remove the field from this PR and add it when the rebuild path is updated. Adding it to the template comment without hooking it up is the worst outcome.

---

### Issue 4: LazyLock pattern — sound for a CLI [NOTE]

`LazyLock<Config>` is appropriate here. The process is single-user, config is immutable after first access, and `std::sync::LazyLock` is `Sync + Send`. The `CLAUDE_PROJECT_DIR → PWD → "."` fallback chain is reasonable. No concerns.

One observation: `commands::Config { action: ConfigCommand::Show }` calls `Config::load()` directly rather than `config::global()`, which means it always reloads from disk. This is probably correct behavior for a diagnostic command, but worth a comment explaining the intentional bypass of the global.

---

### Issue 5: Test coverage gaps [WARN]

The 4 tests cover the happy path well. Missing cases:

- Malformed YAML input to `apply_yaml` — the eprintln path at line 115 is untested. Confirm it returns the unchanged config, not a partial one.
- `disabled_domains` YAML override: no test for the sequence parsing path (line 138).
- The layered resolution order itself (`load()`) is not tested — only the individual `apply_*` methods are. A test constructing temp files would catch the `semantic_engine` issue above.

The existing 4 tests are necessary but not sufficient for a config system that governs all matching behavior.

---

### Summary Table

| # | Severity | Issue |
|---|----------|-------|
| 1 | Block | `engine` vs `semantic_engine` field name break in legacy JSON layer |
| 2 | Warn | Dead `_ways_json_language` function should be removed |
| 3 | Warn | `corpus_stale_secs` parsed but never consumed |
| 4 | Note | `config show` bypasses global intentionally — add a comment |
| 5 | Warn | Test gap: malformed YAML, disabled_domains sequence, load() integration |

Good structural work overall. The grep-anchor migration comments are a practical pattern. Fix issue 1 before merge — the rest can follow.

---

PR #10, opened 2026-04-11
feat: ADR-116 declarative permission requirements

aaronsb:
## Summary

- Add `requires:` field to way frontmatter and attend sensor config declaring tool permissions needed
- Shared containment-aware permission matching engine in `agent-fmt::permissions` (11 unit tests)
- `ways permissions audit` and `attend permissions audit` commands diff declared requirements against `settings.json` grants
- `ways lint` warns on `macro:` without `requires:`; `--fix` auto-populates by scanning macro.sh for external commands
- All 15 macro-bearing ways populated with `requires:` fields via lint --fix
- Built-in attend sensor defaults: context→Read, git→Bash(git:\*), peers→Read, processes→Bash(ps:\*)
- Deprecation notice for `trusted-project-macros` in audit output
- ADR-116 promoted from Draft to Accepted

First real audit found `Bash(jq:*)` missing from settings.json (4 ways need it).

## Test plan

- [ ] `make lint` passes (clippy -D warnings, zero tolerance)
- [ ] `cargo test -p ways` passes (10 scenarios)
- [ ] `cargo test -p agent-fmt` passes (11 permission matching tests)
- [ ] `ways lint --global` shows 0 errors, 1 warning (migration.md — no adjacent macro.sh)
- [ ] `ways permissions audit --global` outputs table with granted/MISSING status
- [ ] `attend permissions audit` outputs table with 4 built-in sensor permissions
- [ ] Revert a way's requires field, run `ways lint --fix --global`, verify it's re-populated correctly

### DIFF
--- a/docs/architecture/system/ADR-116-declarative-permission-requirements.md
+++ b/docs/architecture/system/ADR-116-declarative-permission-requirements.md
@@ -1,5 +1,5 @@
 ---
-status: Draft
+status: Accepted
 date: 2026-04-10
 deciders:
   - aaronsb
--- a/hooks/ways/core.md
+++ b/hooks/ways/core.md
@@ -1,5 +1,6 @@
 ---
 macro: prepend
+requires: ["Bash(awk:*)", "Bash(grep:*)", "Bash(sed:*)", "Bash(sort:*)", "Bash(tr:*)", "Bash(wc:*)"]
 ---
 # Core Ways of Working
 
--- a/hooks/ways/frontmatter-schema.yaml
+++ b/hooks/ways/frontmatter-schema.yaml
@@ -96,6 +96,12 @@ way:
       required: optional
       role: Run macro.sh alongside way content
 
+  permissions:
+    requires:
+      type: array
+      required: optional
+      role: Tool permissions this way needs (e.g. Bash(gh:*), Read). Audited against settings.json.
+
   extended:
     scan_exclude:
       type: regex
--- a/hooks/ways/meta/introspection/introspection.md
+++ b/hooks/ways/meta/introspection/introspection.md
@@ -6,6 +6,7 @@ pattern: pull.?request|create.*pr|pr.*create|write.*pr|open.*pr
 commands: gh\ pr\ create
 macro: prepend
 scope: agent
+requires: ["Bash(jq:*)", "Bash(ways:*)"]
 ---
 <!-- epistemic: heuristic -->
 # Introspection Way
--- a/hooks/ways/meta/knowledge/optimization/optimization.md
+++ b/hooks/ways/meta/knowledge/optimization/optimization.md
@@ -4,6 +4,7 @@ vocabulary: optimize vocabulary suggest gaps coverage unused threshold tune scor
 threshold: 2.0
 macro: prepend
 scope: agent
+requires: ["Read", "Bash(awk:*)", "Bash(find:*)", "Bash(grep:*)", "Bash(sed:*)", "Bash(sort:*)", "Bash(ways:*)"]
 ---
 <!-- epistemic: heuristic -->
 # Way Optimization
--- a/hooks/ways/meta/memory/memory.md
+++ b/hooks/ways/meta/memory/memory.md
@@ -6,6 +6,7 @@ threshold: 80
 pattern: remember|save.*(to|this|that).*memory|note.*(for|this).*(later|next)|don't forget|keep.*in.*mind
 macro: prepend
 scope: agent
+requires: ["Bash(jq:*)", "Bash(sed:*)", "Bash(ways:*)", "Bash(wc:*)"]
 ---
 <!-- epistemic: convention -->
 # Memory
--- a/hooks/ways/meta/todos/todos.md
+++ b/hooks/ways/meta/todos/todos.md
@@ -4,6 +4,7 @@ threshold: 75
 repeat: true
 macro: prepend
 scope: agent, subagent
+requires: ["Bash(jq:*)", "Bash(ways:*)"]
 ---
 <!-- epistemic: heuristic -->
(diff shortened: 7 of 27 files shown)


aaronsb:
## Code Review -- ADR-116 Declarative Permission Requirements

Reviewed all 4 commits across 26 changed files. The core matching engine is solid, tests are comprehensive, and the auto-population from macro.sh is a nice DX touch. Findings below are ordered by significance.

---

### Finding 1: Manual JSON parser is fragile (warn)

**Location**: `tools/agent-fmt/src/permissions.rs:142-182`

`load_settings_permissions` hand-rolls a JSON parser that tracks `in_allow` state line-by-line. This works for well-formatted settings.json but will silently misparse if:

- The `"allow"` key appears in a different context (e.g. a comment-like string, or a nested object that also has `"allow"`)
- The array spans a single line with nested objects
- There are escaped quotes inside permission strings

The comment says "Using serde_json would add a dependency to agent-fmt" -- but `serde_json` is already a transitive dependency via `serde_yaml` in the workspace. Even if you want to keep agent-fmt lean, consider `jq`-style extraction via a tiny helper, or at minimum add test cases for the parser covering: single-line arrays, permissions with special characters, and a settings.json that has `"allow"` in both `permissions.allow` and some other context.

**Risk**: A user with a slightly unusual settings.json format gets "no permissions found" and a confusing audit result with everything marked MISSING.

---

### Finding 2: Display logic duplication (suggestion, non-blocking)

**Location**: `tools/ways-cli/src/cmd/permissions.rs:157-212` vs `tools/attend/src/main.rs:978-1021`

These two functions (`display_audit` in ways and the inline display in attend) are nearly identical -- same table setup, same ANSI coloring, same missing-count accumulation, same summary format. The only differences are the column header ("Way" vs "Sensor") and attend lacking the trusted-project-macros deprecation notice.

Since both tools already depend on `agent_fmt`, this is a natural candidate for a shared `display_audit_table(results, source_label, extras)` function in the `permissions` module. Not blocking, but it would prevent the two from drifting.

---

### Finding 3: Shell tokenizer edge cases (info)

**Location**: `tools/ways-cli/src/cmd/lint.rs:740-807`

The `shell_tokens` function is intentionally simple, which is the right call for a best-effort scanner. Two edge cases worth noting:

1. **Here-docs**: A macro.sh with `cat <<EOF ... some_command ... EOF` will flag `some_command` as a required permission even though it's just data. Low practical risk since here-docs in macro.sh are uncommon, but worth a comment.

2. **Trap handling**: Lines 791-797 have a comment about catching trap commands but the implementation doesn't actually extract the command inside the trap's quoted argument. The `expect_cmd` flag is set to false after consuming "trap", so the quoted command string won't be parsed in command position. This is conservative (misses rather than false-positives), which is fine -- just noting the gap.

3. **`line_uses_command` substring risk**: The function iterates `EXTERNAL_COMMANDS` and calls `shell_tokens` per line. If a macro defines a function like `git_cleanup()`, the `{` would be stripped by `trim_end_matches` but `git_cleanup` wouldn't match `git` because it checks exact token equality. This is correct behavior -- good.

---

### Finding 4: `extract_requires_list` duplicated across crates (suggestion)

**Location**: `tools/ways-cli/src/cmd/lint.rs:845-881` and `tools/ways-cli/src/cmd/permissions.rs:118-154`

`extract_requires_list` (lint.rs) and `extract_requires` (permissions.rs) are functionally identical -- same inline-array and YAML-list parsing. Both live in the same crate (`ways-cli`). Consider extracting to a shared `util` or `frontmatter` module within ways-cli.

---

### Finding 5: `is_valid_permission` is stricter than `Permission::parse` (info)

**Location**: `tools/ways-cli/src/cmd/lint.rs:884-907` vs `tools/agent-fmt/src/permissions.rs:30-50`

The lint validator (`is_valid_permission`) restricts tool names to `Read | Write | Edit | Bash`, while `Permission::parse` accepts any uppercase-starting identifier. This means a permission like `WebSearch(query:*)` would pass `parse` but fail lint validation. This is probably intentional (only Claude Code built-in tools are valid), but the asymmetry could surprise someone extending the system later. A comment in `is_valid_permission` noting this deliberate restriction would help.

---

### Finding 6: Auto-populated requires look reasonable (positive)

Spot-checked several:

- `github.md` requires `Read` + 11 Bash scopes -- matches the macro.sh which calls `gh`, `git`, `jq`, `grep`, `sed`, etc. Correct.
- `adr.md` requires `Bash(chmod:*)`, `Bash(cp:*)`, `Bash(mkdir:*)`, `Bash(touch:*)` -- the macro checks for adr-tool existence and suggests scaffolding. Correct.
- `core.md` requires `Bash(awk:*)` through `Bash(wc:*)` -- the core macro does context gathering. Correct.

The `Read` permission is correctly inferred when macros use `cat`, `head`, `tail`, etc.

---

### Finding 7: Attend sensor defaults are well-chosen (positive)

**Location**: `tools/attend/src/config.rs:58-98`

Built-in sensor defaults map cleanly to what each sensor actually does:
- `context` -> `Read` (reads context state)
- `git` -> `Bash(git:*)`
- `peers` -> `Read` (reads session files)
- `processes` -> `Bash(ps:*)`

The `+` prefix convention for script sensors in the audit display is a nice UX touch for distinguishing built-in from user-added sensors.

---

### Assessment

The implementation is well-structured. The shared matching engine in `agent-fmt::permissions` is the right architectural decision -- single source of truth for containment logic with good test coverage. The `--fix` auto-population is pragmatic and saves manual work.

The main area to watch is the manual JSON parser (Finding 1). It works today but is the most likely source of a confusing failure mode. The duplication (Findings 2, 4) is worth addressing in a follow-up but not blocking.

No security concerns. No blocking issues.

PR #11, opened 2026-04-11
feat: ADR-117 sensor crate extraction with feature flags

aaronsb:
## Summary

- Extract 4 built-in sensors into individual workspace crates: `sensor-git`, `sensor-context`, `sensor-peers`, `sensor-processes`
- New `sensor-trait` crate defines the `Sensor` trait (+Send bound), `Focus`, `SensorSlot`, `AdaptiveInterval`, `DeltaAccumulator`
- attend becomes a thin orchestrator with feature-flagged sensor dependencies
- `register_sensors()` pattern: add a new sensor in 3 steps (crate, dep+feature, one cfg block)
- Config controls runtime activation; feature flags control compilation
- Minimal build with `--no-default-features` compiles clean (orchestrator + script sensors only)
- Workspace grows from 3 to 8 members
- Dead `delta/` and `tick/` modules removed from attend (moved to sensor-trait)
- ADR-117 promoted from Draft to Accepted

## Test plan

- [ ] `make lint` passes (clippy -D warnings across all 8 workspace crates)
- [ ] `cargo test -p agent-fmt -p ways` passes (21 tests)
- [ ] `make attend-rebuild` produces working binary
- [ ] `attend peers` lists active sessions
- [ ] `attend permissions audit` shows sensor permissions
- [ ] `cargo build -p attend --no-default-features` compiles clean
- [ ] Sensor loop works: `/attend` starts via Monitor, notifications arrive


aaronsb:
## Code Review: ADR-117 Sensor Crate Extraction

### What this changes

Extracts 4 built-in sensors from attend into individual workspace crates behind feature flags, introduces a shared `sensor-trait` crate for the `Sensor` trait and runtime types (`SensorSlot`, `AdaptiveInterval`, `DeltaAccumulator`, `Focus`), and adds a `register_sensors()` function with a macro for 3-step sensor addition. Clean decomposition of a monolith into composable units.

### Findings

#### 1. Dead files: old sensor sources still in attend/src/sensors/ (should delete)

**Location**: `tools/attend/src/sensors/git.rs`, `context.rs`, `peer.rs`, `process.rs`

**Problem**: The PR removes `mod git; mod context; mod peer; mod process;` from `sensors/mod.rs` (correctly), but the 4 old `.rs` files (1,328 lines total) are still sitting in the directory. They are unreferenced dead code — Rust won't compile them since they have no `mod` declaration, but they create confusion about which is canonical.

**Why it matters**: Someone reading the tree sees both `tools/sensor-git/src/lib.rs` and `tools/attend/src/sensors/git.rs` with nearly identical content. That is a maintenance trap — an editor might modify the wrong one.

**Suggestion**: Delete the 4 files. They are fully superseded by the crate versions. The diff shows only two differences per file: (a) `use super::{Focus, Sensor}` changed to `use sensor_trait::{Focus, Sensor}` and (b) a `Default` impl added. Everything else is identical — this is a clean extraction.

#### 2. `Send` bound added to Sensor trait — correct and forward-looking

**Location**: `tools/sensor-trait/src/lib.rs:18`

The old `Sensor` trait in `attend/src/sensors/mod.rs` had no `Send` bound. The new trait adds `: Send`. Since the PR compiles clean (including clippy -D warnings), all 5 sensor implementations (`GitSensor`, `ContextSensor`, `PeerSensor`, `ProcessSensor`, `ScriptSensor`) satisfy `Send`. This is correct — none of them hold `Rc`, raw pointers, or other non-Send types. The comment explains the motivation (daemon model, separate threads) which is good forward documentation.

#### 3. Feature flag wiring is correct and complete

**Location**: `tools/attend/Cargo.toml:12-25`, `tools/attend/src/sensors/mod.rs:13-24`, `tools/attend/src/main.rs:448-454,523-529,796-801`

The `dep:sensor-*` syntax is used correctly for optional dependencies. Every `#[cfg(feature = "sensor-peers")]` has a matching `#[cfg(not(feature = "sensor-peers"))]` fallback in main.rs for `cmd_peers()`, `cmd_send()`, and `own_session_id()`. The `--no-default-features` path returns empty/None rather than failing.

One observation: when `sensor-peers` is disabled, `own_session_id()` returns `None`, which means `cmd_send()` will use the "external" sender identity and `cmd_inbox()` cannot filter out own messages (the `own_session_id` variable is empty). This is acceptable — the no-default-features build is a minimal compile check, not a production config — but worth a comment if anyone tries to use it for real.

#### 4. `register_sensors()` pattern — clean and extensible

**Location**: `tools/attend/src/sensors/mod.rs:44-125`

The 3-step documentation is accurate. The macro reduces boilerplate for the common case. PeerSensor gets a hand-written block because of `mark_existing_as_seen()` — that is the right call, a forced abstraction here would add complexity for no gain.

The `#[allow(unused_variables)]` on `cfg` and `#[allow(unused_macros)]` on `register_builtin!` are needed for the `--no-default-features` case. Correct.

#### 5. `DeltaAccumulator::Default` impl added — good

**Location**: `tools/sensor-trait/src/lib.rs:133-137`

The old `delta/mod.rs` lacked a `Default` impl (clippy would flag `new()` without `Default`). The new crate adds it. Good cleanup during extraction.

#### 6. `sensor-trait` is dependency-free — good design

**Location**: `tools/sensor-trait/Cargo.toml`

No external dependencies beyond `std`. This means any sensor crate (including out-of-tree ones) can depend on `sensor-trait` without pulling in a dependency tree. The trait contract uses only stdlib types (`Duration`, `Vec`, `String`, `f64`). This is the right Interface Segregation choice.

#### 7. Workspace manifest ordering

**Location**: `tools/Cargo.toml:1-11`

The ordering is `agent-fmt, sensor-trait, sensor-git, sensor-context, sensor-peers, sensor-processes, ways-cli, attend`. `attend` last is correct (it depends on the others). The sensor crates are grouped together. No issue.

### Assessment

This is a well-executed extraction. The logic is moved, not changed (confirmed by diffing old files against new — only import paths and Default impls differ). The feature flag wiring is complete with fallbacks. The `register_sensors()` pattern keeps extension simple. The `Send` bound is justified and all implementations satisfy it.

**One action item**: delete the 4 orphaned sensor files from `tools/attend/src/sensors/` (`git.rs`, `context.rs`, `peer.rs`, `process.rs`). They are dead code after this PR since `mod.rs` no longer declares them as modules.

Everything else is sound. The two-layer control model (feature flags for compilation, config for runtime) is cleanly separated. The sensor-trait crate has zero dependencies and a minimal surface area. Good Dependency Inversion — attend depends on the trait abstraction, not on concrete sensor implementations.

PR #14, opened 2026-04-11
feat: ADR-118 focus groups — dynamic agent grouping

aaronsb:
## Summary

Implements ADR-118: named focus groups replacing the old static focus file.

- `attend focus on/off/list/all/clear/pin/dissolve` — manage named groups
- `attend send --focus <name>` — send signals to a focus group
- `attend scene private/open` — named presets for focus configuration
- Peer sensor scans joined group directories for signals
- Unified `attend peers` view with focus group context
- Old focus file plumbing removed entirely
- Internal naming: groups.rs (not rooms — agents don't have rooms)

5 commits, 5 unit tests, clean lint + build.

## Test plan

- [ ] `attend focus on deploy && attend focus list` shows group
- [ ] `attend send --focus deploy "msg"` lands signal in @deploy/
- [ ] `attend scene private` clears all groups
- [ ] `attend scene open` joins "open" group  
- [ ] `attend peers` shows focus groups alongside agents
- [ ] `attend status` shows focus section
- [ ] `make lint` passes
- [ ] `cargo test -p attend` passes (5 tests)

### DIFF
--- a/hooks/ways/softwaredev/environment/attend/attend.md
+++ b/hooks/ways/softwaredev/environment/attend/attend.md
@@ -1,6 +1,6 @@
 ---
 description: attend binary — active awareness sensor loop, peer session discovery, inter-session signaling for Claude Code
-vocabulary: attend attend-run attend-send attend-focus attend-peers attend-status sensor-loop awareness-layer focus-group signal-file disclosure-governor peer-session peer-discovery session-awareness environmental-sensing inter-session claude-session another-claude
+vocabulary: attend attend-run attend-send attend-focus attend-peers attend-status attend-scene sensor-loop awareness-layer focus-group signal-file disclosure-governor peer-session peer-discovery session-awareness environmental-sensing inter-session claude-session another-claude scene-private scene-open
 embed_threshold: 0.28
 threshold: 2.0
 pattern: attend|awareness.?layer|peer.?session|peer.?discover|signal.?file|focus.?group|sensor.?loop
@@ -28,24 +28,33 @@ Use `/attend` or launch manually via Monitor with `attend run`.
 ## Peer Messaging
 
 ```bash
-attend send your message here              # send to focus scope
-attend send --broadcast important news     # send to all sessions
-attend send --to /path/to/project message  # directed to one project
+attend send "your message here"            # send to project + focus groups
+attend send --focus deploy "message"       # send to a focus group
+attend send --broadcast "important news"   # send to all sessions
 ```
-
-Send scope mirrors receive scope — if you have a focus group, messages go to all group members.
 
 ## Focus Groups
 
+Named groups for shared signal routing. Dynamic — join and leave as needed.
+
 ```bash
-attend focus add ~/Projects/foo            # join a focus group
-attend focus list                          # show current group
+attend focus on deploy                     # focus on a named group
+attend focus on infra --pin                # focus + persist when empty
+attend focus off deploy                    # release focus
+attend focus list                          # show your groups
 attend focus clear                         # project-only mode
+```
+
+## Scenes
+
+```bash
+attend scene private                       # leave all groups
+attend scene open                          # join shared "open" group
 ```
 
 ## Discovery
 
 ```bash
-attend peers                               # list active Claude sessions
+attend peers                               # sessions with focus groups
 attend status                              # instances, signals, focus state
 ```
--- a/skills/attend/SKILL.md
+++ b/skills/attend/SKILL.md
@@ -43,31 +43,44 @@ All commands below are one-shot — run with Bash, not Monitor.
 
 ### Peer messaging
 
-Send defaults to your current scope (own project + focus group). Always wrap the message in double quotes to prevent shell metacharacter expansion (`?`, `*`, `!`).
+Send defaults to your current scope (own project + focus groups). Always wrap the message in double quotes to prevent shell metacharacter expansion (`?`, `*`, `!`).
 
 ```bash
 attend send "your message here"
+attend send --focus deploy "message to a focus group"
 attend send --broadcast "important announcement for all sessions"
 attend send --to /home/user/other-project "directed message"
 ```
 
 ### Focus groups
 
-A focus group is a set of peer project directories you listen to and send to. Send scope mirrors receive scope.
+Named groups that agents focus on for shared signal routing. Groups are dynamic — join and leave as needed.
 
 ```bash
-attend focus list                    # show current focus group
-attend focus add ~/Projects/foo      # add a peer project
-attend focus add ~/temp ~/bar        # add multiple at once
-attend focus remove ~/temp           # remove a peer
-attend focus clear                   # back to project-only mode
+attend focus list                    # show groups you're focused on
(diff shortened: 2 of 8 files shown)


aaronsb:
## Code Review: PR #14 — ADR-118 Focus Groups

Reviewed: groups.rs, scenes.rs, main.rs, sensors/mod.rs, sensor-peers/src/lib.rs, SKILL.md, attend.md

---

### 1. Naming Consistency — "room" leakage throughout

The commit history shows a deliberate rename from rooms to groups (commit 5), with "focus" as the CLI verb. The CLI surface and docs are clean. However, the internal code still has extensive "room" terminology that didn't get swept in the rename:

**`groups.rs`** — 20+ instances:
- `ROOM_PREFIX` constant (line 29)
- Doc comments: "room's signal directory" (39), "Join a named room" (49), "Leave a named room" (70), "Pin a room" (81), "Dissolve a room" (100), "List rooms this session has joined" (117), "List all active rooms" (127), etc.
- Error messages: "room name cannot be empty" (242), "room name cannot start with" (245), "room name cannot contain" (248)
- Variable names: `rooms` in `all_groups()` (130), `rooms` in `parse_groups_yaml()` (280)
- Parser comments: "Top-level: room name" (293), "room properties" (311)

**`scenes.rs`**:
- `Scene::rooms` field (19) — this one propagates to config parsing and the YAML key `rooms:` in scenes.yaml, so renaming it affects the config file format. Intentional or oversight?
- `room_name` variable (55)
- Doc comments throughout (3, 7, 8, 41, 49, 54, 60)

**`main.rs`**:
- "Initialize rooms for signal routing" comment (114)
- "joined rooms" comment (607)

**`sensors/mod.rs`**:
- "room directories" comment (87)

**Assessment**: The user-facing surface is consistent — "focus" as verb, "groups" as noun. The internal naming is cosmetic noise, not a functional issue. But the error messages at lines 242-248 in groups.rs *are* user-facing — they leak "room" to the CLI. Those should say "group name" instead of "room name".

---

### 2. Concurrent Access — _groups.yaml TOCTOU

`_groups.yaml` is a shared mutable file with no locking. The pattern in every mutating method is: `load_state()` then `save_state()` — classic read-modify-write without atomicity.

Scenario: Two attend instances both call `join()` at roughly the same time. Both read the file, both see the same members list, both write back. One session's join is silently lost.

The signal file writing (cmd_send, line 636-641) already uses atomic write-then-rename. The state file does not — `save_state()` calls `fs::write()` directly (line 219).

**Severity**: Low-medium. In practice, joins/leaves happen infrequently and from interactive commands, so collisions are unlikely. But `cleanup_stale()` running inside `attend run` while another session does `focus on` is a realistic race.

**Suggestion**: At minimum, apply the same atomic write pattern already used for signals — write to `_groups.yaml.tmp`, then `rename()`. For full correctness, advisory file locking (`flock`) around the read-modify-write would be defensive. Not a blocker, but worth a follow-up.

---

### 3. Cleanup Logic — Double Save in `leave()` and `cleanup_if_empty()`

`leave()` (line 71-79) does:
1. Load state, remove member, save state
2. Call `cleanup_if_empty(name, &state)` with the just-saved state

`cleanup_if_empty()` (line 222-235) then:
1. Checks if empty and unpinned
2. **Clones the state**, removes the entry, saves *again*

This means every `leave()` that results in an empty unpinned group does two writes to `_groups.yaml` in rapid succession. The same pattern exists in `unpin()` (line 91-98).

Worse, `cleanup_if_empty()` clones the *passed-in* state to make a new HashMap, but between the first `save_state()` and the second, another process could have written to the file. The clone operates on the caller's in-memory copy, not the current file state. This interacts with the TOCTOU issue above.

**Suggestion**: Fold the cleanup logic into the mutation method itself. After removing the member, check if cleanup is needed *before* the first save, then save once. This eliminates the double write and the stale-state clone.

---

### 4. `session_alive()` — Stub That Always Returns True

`session_alive()` (line 257-263) always returns `true`, meaning `cleanup_stale()` will never actually remove any stale members. This makes `cleanup_stale()` (called from `focus all`) effectively a no-op for member pruning — it only cleans up groups that are already empty.

The TODO comment is honest about it. This means abandoned sessions permanently pollute group membership until someone manually calls `focus dissolve`. Acceptable for an initial implementation, but worth tracking.

---

### 5. Signal Routing Completeness

Checked all code paths:

| Path | Project dir | Broadcast | Focus groups |
|------|------------|-----------|-------------|
| `read_signals()` (sensor-peers) | yes | yes | yes (extra_scan_dirs) |
| `mark_existing_as_seen()` | yes | yes | yes (extra_scan_dirs) |
| `cmd_inbox()` | yes | yes | yes (joined_group_names) |
| `cmd_send()` default | yes | no | yes (joined_group_names) |
| `cmd_send --broadcast` | no | yes | no |
| `cmd_send --focus <name>` | no | no | yes (specific group) |

This looks correct. The `receive_dirs()` method on Groups (line 146-161) is marked `#[allow(dead_code)]` and is unused — the actual routing is done ad-hoc in each command. Consider removing it or wiring it in to avoid the implementations drifting apart.

**One concern**: `register_sensors()` captures the group dirs at startup (sensors/mod.rs line 88-93). If an agent joins a new group *after* `attend run` starts, the peer sensor won't scan the new group's signal directory until restart. The `extra_scan_dirs` are set once and never updated. This is a limitation worth documenting — the workaround is restart attend after changing focus groups.

---

### 6. Old Focus File Removal

Searched for `focus_file`, `focus_path`, `.focus`, `add_focus`, `remove_focus`, `focus_dirs` across both `attend/src` and `sensor-peers/src` — no matches. The old static focus file plumbing is cleanly removed.

**One remnant**: The startup banner in main.rs line 157 still says `attend focus add <path>` in the commands hint. That's the old syntax — should be something like `attend focus on <name>`.

---

### 7. Scene Activation

`activate()` in scenes.rs (line 42-63): Leaves all current groups, then joins the scene's groups. The leave/join order is correct — avoids the scenario where joining before leaving could hit name conflicts.

However, `groups.leave()` returns `Result` and the error is silently discarded with `.ok()` (line 51). If a leave fails (e.g., the group was already dissolved by another agent), the function continues and may join new groups while the old ones are partially released. Low risk in practice, but inconsistent with the error handling in `join()` which propagates errors.

---

### 8. Summary of Findings

**Should fix before merge:**
- Startup banner references old `focus add <path>` syntax (main.rs:157) — will confuse agents

**Should fix soon:**
- User-facing error messages say "room" not "group" (groups.rs:242-248)
- `cleanup_if_empty()` double-write pattern (groups.rs:222-235 called from leave/unpin)

**Track for follow-up:**
- `_groups.yaml` lacks atomic writes or locking (concurrent access risk)
- `extra_scan_dirs` set at startup only — new group joins not reflected until restart
- `session_alive()` stub means stale members persist indefinitely
- `receive_dirs()` dead code — either use it or remove it

**Design is sound**: The group/focus/scene layering is clean. CLI surface is well-organized. Signal routing covers all paths correctly. The scenes.yaml config format with inline arrays and list items is well-handled. Test coverage on YAML roundtrip and name validation is good.

Issue #27, opened 2026-04-13
sensor-processes: exit-code awareness


aaronsb:
`sensor-processes` currently fires "cargo exited" without knowing whether the build succeeded or failed. Both outcomes produce the same notification, which means the agent can't tell if it needs to react to an error or celebrate.

## What's missing

When a watched process exits, capture its exit code and surface it in the observation:

- `cargo exited (success)` — magnitude 2.0
- `cargo exited (failure, code 101)` — magnitude 3.5 (louder, breaks refractory)

## Implementation notes

Options:

1. **Parse `/proc/<pid>/status`** before the process disappears. Race condition: might miss fast exits.
2. **`wait4`/`waitid`** against the target PID if we started it ourselves. Doesn't apply here — we observe processes, we don't spawn them.
3. **ptrace or process accounting.** Heavy; adds kernel config dependencies.
4. **Parse shell history or build logs.** Unreliable.
5. **Marker file convention.** User's shell or build tool writes `$XDG_STATE_HOME/attend/last-build-status` with exit code; sensor reads it.

Option 5 is the pragmatic one: document a convention, let users opt in by adding a wrapper around their build commands. The sensor reads the marker on each poll. Not perfect, but it works without kernel gymnastics.

## Pairs with

#6 — build event batching. If we aggregate process events into "build happened," the exit code naturally attaches to the aggregate.

## Documentation

[`docs/attend-and-monitor/sensors.md`](../blob/main/docs/attend-and-monitor/sensors.md) already calls this out as a limitation:

> Doesn't distinguish success from failure — "cargo exited" fires whether the build succeeded or errored. For exit-code awareness, pair this with an external script sensor that watches $? or parses a build output file.

PR #40, opened 2026-04-13
feat(sensor-processes): configurable watch list (#28)

### LINKED ISSUE #28: sensor-processes: configurable watch list
The whitelist of watched processes in `sensor-processes` is hardcoded:

```
cargo, rustc, make, cmake, ninja,
gcc, g++, cc, c++, clang, clang++,
go, npm, yarn, pnpm, tsc,
mvn, gradle, pip, pip3
```

Users working in languages or toolchains outside this list don't get process awareness. A Rust + Python dev gets the Rust half; an Elixir dev gets nothing; a Zig dev gets nothing; a user running custom shell scripts gets nothing.

## What to add

A config-driven watch list:

```yaml
sensors:
  processes:
    watch:
      - cargo
      - rustc
      - mix        # elixir
      - zig
      - ./build.sh # custom
```

If `watch:` is absent, fall back to the current hardcoded defaults. If present, use it verbatim (no merging with defaults — explicit opt-in).

## Scope

- [ ] Parse `watch:` array from config (minor YAML parser extension)
- [ ] Pass the list into `ProcessSensor::new()` instead of the hardcoded constant
- [ ] Fall back to the current defaults when not set
- [ ] Document in `configuration.md` and `sensors.md`

## Non-goals

- Process-specific magnitude overrides (one magnitude per sensor is enough for now)
- Regex matching on process names (exact match is fine)

## Documentation

Called out in [`docs/attend-and-monitor/sensors.md`](../blob/main/docs/attend-and-monitor/sensors.md):

> A future extension could make this list configurable; today it's hardcoded.

aaronsb:
Closes #28.

## Summary

Replaces the hardcoded build-tool list in `sensor-processes` with a configurable `sensors.processes.watch` entry in attend config. An explicit list **replaces** the defaults verbatim — no merging — so the config is the single source of truth. An elixir dev who lists \`mix\` does not silently keep getting \`cargo\` enrichment.

### Plumbing

- **`sensor-processes::DEFAULT_WATCH`** now exposes the default list as a `const &[&str]`, used by `ProcessSensor::new()`. A new `ProcessSensor::with_watch(Vec<String>)` takes an explicit list (empty vec = disable enrichment entirely), and a private \`is_watched()\` helper replaces the inline \`contains\`-check in \`poll()\`. The hardcoded array literal is gone.
- **`attend::config::SensorConfig`** gains \`watch: Option<Vec<String>>\`. \`None\` means "use the sensor's built-in default"; \`Some(list)\` is the explicit override.
- **\`attend::sensors::mod::register_all\`** resolves the watch list once at startup — if config provides one it calls \`with_watch\`, otherwise \`new()\`.

### Parser extension (unlocks a real bug fix for \`requires:\` too)

The hand-rolled YAML parser now accepts **block-form list syntax** — the \`key:\` followed by indented \`- item\` lines style — for both \`watch:\` and \`requires:\`. Previously only inline arrays worked, so a perfectly reasonable YAML

\`\`\`yaml
requires:
  - Bash(git:*)
  - Bash(gh:*)
\`\`\`

was being silently dropped. The new code path tracks a \`current_list_key\` across lines, consumes \`- item\` continuations, and terminates on the next non-list line. Blank lines and comments inside a block list don't terminate it.

A shared \`parse_inline_list()\` helper now powers the inline form for both \`requires:\` and \`watch:\` so the quoting/whitespace rules stay in one place.

## Config shapes

Both YAML forms are accepted:

\`\`\`yaml
sensors:
  processes:
    watch:
      - cargo
      - rustc
      - mix          # elixir
      - zig
      - ./build.sh
\`\`\`

\`\`\`yaml
sensors:
  processes:
    watch: [cargo, rustc, mix, zig]
\`\`\`

Disable enrichment entirely:

\`\`\`yaml
sensors:
  processes:
    watch: []
\`\`\`

## Test plan

- [x] \`sensor-processes\`: 22 tests pass (18 pre-existing + 4 new — \`new_uses_default_watch_list\`, \`with_watch_replaces_defaults\`, \`with_watch_empty_disables_enrichment\`, \`default_watch_const_matches_expected_members\`)
- [x] \`attend\`: 12 tests pass (5 pre-existing + 7 new config tests). New coverage: inline watch, block-form watch, absent watch, watch terminating before a sibling property, **regression test for block-form requires** (which was silently broken before), inline requires (still works), and blank-line/comment tolerance inside a block list.
- [x] \`cargo clippy --workspace --all-targets -- -D warnings\` clean
- [ ] CI green
- [ ] Dogfood: drop \`watch: [cargo, rustc]\` into my own config, restart attend, verify no cargo exits are enriched by a \`mix\` marker (and vice versa). Not blocking — the unit tests cover the same paths.

## Docs

- **\`sensors.md\`** "Watched processes" section now explains the override, shows both YAML shapes, and calls out the explicit-replace contract (empty list = disable enrichment). Removed the "future extension" / "hardcoded" note.
- **\`configuration.md\`** adds \`processes.watch\` to the per-sensor keys table and updates the parser-notes section to list block-form lists as supported.

## Relationship to prior PRs

Composes cleanly with #27's marker correlation — a process on the user's custom watch list still gets the success/failure enrichment via the marker. The watch list only controls *which* processes participate in enrichment, not *how* enrichment works.

### DIFF
--- a/docs/attend-and-monitor/configuration.md
+++ b/docs/attend-and-monitor/configuration.md
@@ -182,6 +182,10 @@ All sensors (built-in or script) accept:
 - **`script`** (path, script sensors only): path to the executable
 - **`requires`** (list): permission strings audited against `settings.json` (ADR-116)
 
+Some sensors accept additional sensor-specific keys:
+
+- **`processes.watch`** (list): overrides the default build-tool watch list. Only processes on this list get exit-code enrichment (success/failure magnitudes, marker correlation); everything else produces a plain `X exited`. Explicit-replace — passing `watch:` drops the defaults. See [`sensors.md`](sensors.md#watched-processes) for the default list and format examples.
+
 ## Overlay semantics
 
 The overlay layers project-scope on top of user-scope. For each setting:
@@ -242,8 +246,9 @@ Attend's YAML parser is deliberately minimal — it's a hand-written subset that
 
 - Two-level section headers (`governor:`, `engagement:`, `sensors:`)
 - Four-space indent for sensor properties
-- Inline arrays for `requires:` (`requires: [Bash(gh:*), Read]`)
-- Comments (`#`) and blank lines
+- Inline arrays (`requires: [Bash(gh:*), Read]`, `watch: [cargo, mix]`)
+- Block-form lists on a new line (`requires:` followed by indented `- item` lines) for `requires:` and `watch:`
+- Comments (`#`) and blank lines (they don't terminate a block-form list)
 - `+name:` and `-name:` sensor prefixes
 
 It does **not** support:
--- a/docs/attend-and-monitor/sensors.md
+++ b/docs/attend-and-monitor/sensors.md
@@ -116,6 +116,8 @@ Tracks specific dev-tool processes running under the user's session. Not a gener
 
 ### Watched processes
 
+The sensor enriches a specific list of build/dev tools with "exited (success/failure)" events. The default list covers common toolchains:
+
 ```
 cargo, rustc, make, cmake, ninja,
 gcc, g++, cc, c++, clang, clang++,
@@ -123,7 +125,30 @@ go, npm, yarn, pnpm, tsc,
 mvn, gradle, pip, pip3
 ```
 
-A future extension could make this list configurable; today it's hardcoded.
+Any process *not* on this list still produces a plain `X exited` at magnitude 2.0 when it disappears from the snapshot — the watch list only controls who gets the marker-correlated enrichment (success/failure codes, louder magnitudes). Process **start** events are not gated by the watch list.
+
+**Overriding the list.** Set `sensors.processes.watch:` in your attend config. Both YAML shapes are accepted:
+
+```yaml
+sensors:
+  processes:
+    watch:
+      - cargo
+      - rustc
+      - mix          # elixir
+      - zig
+      - ./build.sh
+```
+
+Or inline:
+
+```yaml
+sensors:
+  processes:
+    watch: [cargo, rustc, mix, zig]
+```
+
+The explicit list **replaces** the defaults verbatim — no merging. If you want the defaults plus one extra entry, you list all of them. If you want to disable enrichment entirely, pass an empty list (`watch: []`). This contract makes the config the single source of truth: what you write is what the sensor runs.
 
 ### What it emits
 
--- a/tools/attend/src/config.rs
+++ b/tools/attend/src/config.rs
@@ -693,4 +693,37 @@ mod tests {
             vec!["cargo", "mix"]
         );
     }
+
+    #[test]
+    fn block_list_terminates_across_sensor_boundary() {
+        // Regression test flagged in code review: a block-form list in
+        // one sensor block must not bleed into the next sensor block
+        // when they're adjacent. The new sensor line clears
+        // `current_list_key` via the "any other line" path before
+        // `current_sensor` is reassigned.
+        let mut cfg = Config::default();
+        apply_config(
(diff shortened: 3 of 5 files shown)


aaronsb:
## Review Summary

Reviewed the parser state machine, wiring, and test coverage. The PR is in good shape — no blockers. A few notes below, then specifics.

**What this accomplishes:** Hoists `build_tools` to `pub const DEFAULT_WATCH`, adds `ProcessSensor::with_watch` / `is_watched`, and extends the hand-rolled YAML parser to handle block-form lists for both `watch:` and `requires:`. Retroactively fixes a silent bug where block-form `requires:` was being dropped.

**Assessment:** Clean. The explicit-replace contract is wired correctly (`sensors/mod.rs:85-88`), the state machine handles the realistic edge cases, and test coverage for the new surface is solid.

### Blocking
None.

### Should-fix
None. A couple of items I considered but don't think need to block merge:

- **Permissions audit behavior change** (`main.rs:1678`). The retroactive `requires:` fix means any existing user config with block-form `requires:` will now populate the audit where it previously showed empty. This is strictly correct (declarations were always in the file, just invisible), but it could surface audit findings that look new. Worth a line in the PR body or release notes so users aren't surprised. No code change needed.

### Nit
- **Unguarded list-item indent** (`config.rs:344`). `trimmed.starts_with(\"- \")` matches at any indent, so a stray top-level `- value` after a block list (without an intervening non-list line) would be consumed into the last `current_list_key`. This requires malformed YAML to trigger and the blast radius is \"one bad config value,\" so it's not worth fixing, but consider a defensive indent check if this parser grows more responsibilities.
- **`parse_inline_list` and quoted commas** (`config.rs:562`). Splits naively on `,`, so `watch: [\"a,b\", c]` would parse as three items. Real watch values are process names — commas are extremely unlikely — so this is fine. Document it in a comment if you want to lock the contract.
- **`DEFAULT_WATCH` length assertion** (`lib.rs:526`). `assert_eq!(DEFAULT_WATCH.len(), 20)` is intentionally brittle (the test comment says so). Good tradeoff; noting it so future readers know it's load-bearing.

### Traced edge cases (all pass)
- Block list followed by sibling property at indent 4 — terminates cleanly (covered by `watch_block_form_terminates_on_next_property`).
- Block list followed by a new sensor at indent 2 — `current_list_key.clear()` on the non-`- ` line, `current_sensor` reassigned.
- Block list followed by a new top-level section (`governor:`) — clears correctly.
- Block list with interleaved blank lines and comments — skipped before the `clear()`, list survives (covered by `block_list_items_survive_blank_lines_and_comments`).
- Empty `watch: []` → `Some(vec![])` → `with_watch(vec![])` → enrichment disabled, not defaults fallback. Correct and tested (`with_watch_empty_disables_enrichment`).

### Test coverage gaps (not blocking)
- No explicit test for the \"block list terminates when `current_sensor` changes to another sensor whose first key is also a block list\" case. The logic is sound by inspection but an explicit test would be cheap insurance if this parser grows further.
- No test for the sensor-boundary case specifically (block list in one sensor, another sensor block immediately after). Again, inspection-safe.

### What looks good
- `is_watched()` extraction cleanly hides the watch-list lookup and keeps `poll()` readable.
- `Option<Vec<String>>` on `SensorConfig.watch` is the right encoding for \"None = defaults, Some([]) = disabled, Some(list) = replace.\" The three arms are distinct and the wiring honors them.
- The `current_list_key.clear()` placement *after* the comment/blank-line skip is exactly right — it's the one thing most likely to be wrong in a hand-rolled parser and this PR gets it right.

PR #49, opened 2026-04-15
feat: ADR-123 firing dynamics — progression-axis unification

aaronsb:
Lands ADR-123: unify firing dynamics for attend and ways on a single progression-axis engine, plus the shared-lint hygiene that fell out along the way, plus Phase D reactive firing, Phase E empirical tuning, and the Phase F re-observation that validated the baseline.

**[ADR-123](docs/architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md)** is the canonical reference — now `Accepted` with a comprehensive Implementation Status section. **[Phase F re-observation](docs/hooks-and-ways/observed-behavior.md#re-observation--2026-04-14--post-adr-123)** is the empirical gate the `Accepted` flip cleared.

## Summary

Before this PR, attend and ways had two unrelated firing-dynamics implementations:
- attend ran ADR-119's action potential model with linear decay, time-windowed burst detection, and `Instant`-keyed state.
- ways ran a flat `REDISCLOSE_PCT: u64 = 25` step function with no refractory, no curves, no re-engagement reset.

The math was the same underneath; only the units differed (seconds vs token position). This PR extracts that shared math into a unit-agnostic `Curve` enum in `sensor-trait`, migrates both tools onto it, adds the missing pieces (reactive firing, empirical tuning, per-way visualization, cross-tool lint consistency) that the unification unlocked, and closes the Phase F validation loop against the 2026-03-17 baseline in observed-behavior.md.

### Engine
- **New `sensor-trait::curve::Curve`** enum with four variants: `Exponential`, `ActionPotential`, `ProgressiveStaircase`, `Flat`. Each has `salience_at(delta)` (outward gate) and `multiplier_at(delta, history, current)` (inward gate). Event-count burst detection replaces time-windowed burst detection to support chunky progression axes (ways' token position). 26 unit tests.
- **New `sensor-trait::engagement::EngagementState`** — unit-agnostic state owning a `Curve`, a tick-keyed fire history, and `should_fire` / `record_fire` / `current_salience` / `current_multiplier` / `in_absolute_refractory` / `effective_threshold` methods. Derives `Serialize/Deserialize` for per-session persistence.
- **New `Tick` / `TickDelta`** type aliases (`u64`) — the progression axis is caller-supplied and opaque to the engine.
- **`sensor_trait::epoch_secs()`** — canonical wall-clock-seconds tick source for attend.

### attend migration
- `SensorSlot::engagement` swapped to the new `EngagementState`. Construction reads attend's yaml engagement config, converts `decay_per_minute` to `multiplier_half_life` via `ln(0.5) / ln(1 - rate) × 60`, computes `peak_multiplier = 1 + step_multiplier`, and builds a `Curve::ActionPotential`.
- Yaml keys stay stable (`burst_threshold`, `step_multiplier`, `absolute_refractory`, `decay_per_minute`, `peer_activity_window`). `burst_window` is parsed for back-compat, has no runtime effect, and no longer shows up in `attend config show` when the file omits it (679b205).
- Old `EngagementState` (linear decay, `Instant`-keyed) deleted. `record_disclosure()` → `record_fire(tick, 1.0)`.

### ways integration
- **Every way declares an explicit `curve:` block.** The legacy `redisclose: N` field is fully removed from the Frontmatter schema; `detect_legacy_redisclose()` rejects leftover fields at parse time with a migration-pointing error.
- **99 way files migrated.** 8 had existing `redisclose` fields (mapped to `Curve::Exponential { half_life: N × 2000 }` on a 200k baseline); 91 others defaulted to `Curve::Exponential { half_life: 30000 }`. 100 ways total pass `ways lint --global`.
- **New `session::FireOutcome`** (`FirstFire | ReFire | Suppressed`), `way_fire_outcome()`, `record_way_fire()`. Per-way state persisted as JSON at `{session_dir}/way-engagement/{way_id}.json`.
- **`REFIRE_FLOOR = 0.5`** — `Curve::Exponential { half_life: H }` re-fires at delta H; `Curve::Flat { suppression: N }` re-fires at delta N.
- `REDISCLOSE_PCT`, `token_distance_exceeded()`, and `detect_context_window()` deleted from session.rs. Old marker paths (`way_is_shown`, `stamp_way_marker`, `stamp_way_epoch`) preserved for tree-metrics and scan-time "has this way been seen" queries.

### Reactive firing (Phase D)
- **`PostToolUse` and `PostToolUseFailure` hooks** wired in `settings.json`, matching `Edit|Write|Bash|Task`.
- **`hooks/ways/check-post.sh`** — shell dispatcher that walks each way dir for a `postcheck.sh`, pipes the tool_response through, and shells out to `ways show way` on exit 0. Reactive fire requests flow through the same engine gate as predictive fires.
- **`hooks/ways/softwaredev/code/quality/postcheck.sh`** — load-bearing demo. Exits 0 when an Edit/Write touches a file that has grown past 500 lines. Verified end-to-end, and then verified *again* in the Phase F observation session when it fired reactively on `tune_curves.rs` at epoch 83 as that file crossed the threshold mid-edit.

### Per-way visualization
- **`Curve::refire_delta(floor)`** — inverts salience_at to compute per-way re-fire distance.
- **`render::WayRow::refire_threshold_k()`** added to the trait. `compute_bar_positions`, `write_way_row`, `write_token_timeline`, `predict_next` now drive each row's bar and forecast off its own threshold. The "NK interval" footer becomes "min–max K intervals" for heterogeneous sessions.
- `session::way_refire_threshold_k(way_id, project_dir)` resolves a way id to its threshold by reading frontmatter curve. `list` and `rethink` both consume it; unknown ways fall back to 25% of context window.
- `cmd/context.rs` drops its 25% re-disclosure marker entirely. Per-way re-fire now belongs in `ways list`.
- `ways list` on a live session renders "20–30K intervals" (the two `half_life` values used across the 100 ways) instead of the pre-refactor hard-coded "250K interval".

### Empirical tuning (Phase E)
- **`ways tune-curves` subcommand.** Surveys `~/.claude/stats/events.jsonl`, pairs `way_fired` and `way_redisclosed` events by (way, session), computes token-position deltas between consecutive fires, and suggests a calibrated `Curve::Exponential` half-life per way. Dry-run by default, `--apply` rewrites the frontmatter `curve:` block via line-surgery. Floor of 3 delta samples, ±20% tolerance band, round to nearest 500 tokens. Filters: `--way PAT`, `--project PAT`, `--min-fires N`.
- **`way_fired` events enriched with `token_position`.** One-field addition to the existing `log_fields` vec in `show/mod.rs` — tune-curves needs per-fire tick values to compute deltas, and `token_pos` was already being computed for the marker-stamping path. Historical rows (pre-8b20782) lack the field and are correctly excluded from tune analysis.
- Verified against the live events log during the Phase F session: 2 ways with early same-session delta samples, both within ±20% of migration defaults — a positive signal that the initial `half_life: 30000` / 20000 / etc. guesses were well-calibrated for the ways that have accumulated data.

### Shared lint hygiene
- **`ways lint` split** from a single 938-line file into 7 focused modules (`schema`, `helpers`, `per_file`, `scanning`, `locale_stubs`, `requires`, `provenance`). Triggered by the Phase D `softwaredev/code/quality` postcheck firing on the file we were actively editing — the reactive firing path caught its own maintenance cost in real time.
- **`--fix` removes foreign top-level fields** plus any indented continuation, and foreign `when:` sub-fields. Uses the new `remove_top_level_field` / `remove_when_subfield` helpers.
- **`x-*` prefix escape hatch** (JSON Schema / OpenAPI / package.json convention) — fields starting with `x-` are silently skipped by UNKNOWN / DEPRECATED checks. Applies to way frontmatter, `when:` sub-fields, and locale stub keys uniformly.
- **Locale stub validation** — `.locales.jsonl` files are now validated per-line against a new `locale_stub:` schema section (`lang`, `description`, `vocabulary`, `embed_threshold`). 83 stub files newly covered. `--fix` uses `serde_json::Map::shift_remove` (not `remove` which is swap_remove-equivalent even with `preserve_order`) to avoid churning git diffs.
- **New `attend config lint`** subcommand — sibling to `ways lint` with matching semantics: UNKNOWN / DEPRECATED categories, `x-*` escape, `--fix` surgical line removal, `--check` for CI mode. Runtime loader in `config.rs` stays forgiving (unknowns silently ignored so typos never crash startup); lint is where typos surface. `engagement.burst_window` is the first DEPRECATED entry with a pointer to ADR-123's "burst window is implicit in multiplier_half_life" reframing.

### Secondary fixes
- **`attend config show` no longer shows synthesized burst_window default (679b205).** `EngagementConfig::burst_window` is now `Option<Duration>`. Absent-from-file is a first-class state, not silently filled with 900. Phase 2 full-removal tracked as issue #50.

### Docs
- **`docs/attend-and-monitor/engagement.md`** — full rewrite for ADR-123. New sections: "Progression axis: wall-clock seconds for attend", "Curves as first-class", "Event-count burst detection", "How attend calls the engine", "Shared engine, cross-tool". Mermaid state-diagram labels updated (`record_disclosure()` → `record_fire(tick, 1.0)`). Live tune output refreshed.
- **`docs/attend-and-monitor/salience.md`** — full rewrite (d991209). Old "designed, not implemented" framing replaced with "mechanism unified in ADR-123, ways is the first concrete implementation, attend sensor-peers application still deferred." Reframes inward/outward as one engine, two queries. Points at `session::way_fire_outcome` as the canonical code reference.
- **`docs/hooks-and-ways/context-decay.md`** — Steady-State mechanisms table row rewritten from "Once-per-session gating" to "Per-way curve" with a paragraph explaining how `REFIRE_FLOOR` replaces the old hard gate.
- **`docs/attend-and-monitor/loop.md`** — timers table: `burst_window` row dropped, `decay_per_minute` row rewritten as exponential half-life, mermaid note updated.
- **`docs/attend-and-monitor/configuration.md`** — engagement section rewritten with yaml-to-runtime mapping table. `burst_window` marked **DEPRECATED** with pointer to `attend config lint --fix`. New "Bootstrapping" entry for `attend config lint`.
- **`docs/hooks-and-ways/observed-behavior.md`** — new dated section (4672e44) documenting the Phase F re-observation against the new stack. Preserves the 2026-03-17 baseline entry above verbatim and structures the new entry for direct comparison.
- **ADR-104, ADR-119, ADR-121 rewritten in place** — Status → `Superseded`, `superseded_by: ADR-123` in frontmatter, original Context + Alternatives preserved as historical decision-trail, Decision sections rewritten to describe what runs, new "What ADR-123 changed" sections enumerate the deltas. Passes `adr lint`: 0 errors, 0 warnings across all 29 ADRs.
- **ADR-123 itself flipped Draft → Accepted (25d0d11)** with a comprehensive Implementation Status section listing shipped phases, deferred items, and resolved open questions.
- **INDEX.md regenerated** via `adr index`.

## Branch commits

| # | SHA | Title |
|---|---|---|
| 1 | 1f30b80 | feat(sensor-trait): progression-axis Curve enum and EngagementState (Phase A) |
| 2 | 9340185 | feat(attend): migrate to progression-axis EngagementState (Phase B) |
| 3 | b94cc25 | feat(ways): migrate frontmatter from redisclose to curve (Phase C1+C2) |
| 4 | ad20744 | feat(ways): wire per-way EngagementState into firing gate (Phase C3-C5) |
| 5 | 47bc9de | feat(ways): per-way re-fire visualization + mass curve migration |
| 6 | dddb4ca | feat(ways): reactive firing via PostToolUse + postcheck.sh (Phase D) |
| 7 | 057f3f3 | docs: rewrite attend engagement + loop + configuration for ADR-123 |
| 8 | 1397d68 | feat(ways): split lint.rs + x- escape + --fix removal + locale stub validation |
| 9 | f360ea6 | feat(attend): config lint subcommand with UNKNOWN/DEPRECATED + x- escape |
| 10 | c869841 | docs: configuration.md surfaces attend config lint |
| 11 | 7449a61 | docs(adr): rewrite ADR-104, 119, 121 to reflect what shipped via ADR-123 |
| 12 | 679b205 | fix(attend): hide synthesized burst_window default in config show |
| 13 | 8b20782 | feat(ways): log token_position with way_fired events for tune input |
| 14 | 28527e8 | feat(ways): tune-curves subcommand for empirical curve calibration (Phase E) |
| 15 | 4672e44 | docs: Phase F re-observation — ADR-123 stack preserves baseline |
| 16 | d991209 | docs: rewrite attend salience.md for ADR-123 unified engine |
| 17 | 25d0d11 | docs(adr): ADR-123 Draft → Accepted after Phase F validation |

### Automated (green on branch)
- [x] `cargo test --workspace` — all passing (sensor-trait 26, ways-cli 44 including 7 new tune_curves tests + 10 session-sim, attend 17 including 4 config_lint tests)
- [x] `ways lint --global` — 0 errors, 1 pre-existing warning (unrelated macro: without requires:)
- [x] `attend config lint` on the user's live config — clean after --fix removed `burst_window`
- [x] `ways corpus` — regenerates cleanly against the migrated frontmatter
- [x] `adr lint` — 0 errors 0 warnings across 29 ADRs

### Operator-level verification (done during Phase F observation session)
- [x] **B3 attend live parity check** — attend ran as a persistent Monitor during the Phase F observation session. Absolute refractory, relative suppression, exponential decay, governor cooldown, and graceful binary reload all verified in flight. No regression from pre-ADR-123 behavior.
- [x] **`ways list` / `ways rethink` over a real session** — `ways list` ground-truthed against the observation session, per-row re-fire positions rendered sensibly, footer correctly showed "20–30K intervals". `ways rethink` TUI walkthrough confirmed by operator: leading-edge compaction visible as an animated pattern — the cursor-following behavior ADR-123's decay model predicts.
- [x] **Reactive firing caught itself mid-edit** — the quality way fired via postcheck at epoch 83 on `tune_curves.rs` as that file crossed 500 lines during the Phase F session. End-to-end verification that no synthetic test can match.

### Phase F acceptance criterion (met)
- [x] New implementation is at least as good as the old one on the same task shape. See [Phase F re-observation in observed-behavior.md](docs/hooks-and-ways/observed-behavior.md#re-observation--2026-04-14--post-adr-123). 92 turns, 212K / 1000K context, zero operator redirections, several in-session detours handled without supertask drift.

## Follow-up issues (not blocking merge)

- **#50** — attend: remove deprecated `engagement.burst_window` from schema entirely (Phase 2 of the soft-removal started in 679b205). Wait at least two weeks after merge.
- **#51** — attend: split `tools/attend/src/main.rs` (2027 lines) into `cmd/*` modules. Proposed layout in the issue. Pure structural refactor, no ADR-123 dependency. `tools/ways-cli/src/main.rs` (now at 602 lines) mirror case tracked as an addendum comment on the same issue.
- **`attend config lint --check` in CI.** Wire alongside `ways lint --check` as a CI job.
- **Attend sensor-peers outward-gate application.** ADR-121's step 7, architecturally available under ADR-123. Pure new consumer of an existing facility. See [salience.md](docs/attend-and-monitor/salience.md#attends-deferred-application-still-to-ship) for the path forward.
- **`attend status` refractory state display.** ADR-119 step 7.
- **Motivation / reflection-overdue sensor.** ADR-119 steps 8–9.


aaronsb:
## Code Review — ADR-123 firing dynamics unification

Reviewed the cross-cutting concerns in the focus list. No blocking issues. Several follow-up observations worth tracking; one I'd ask you to look at before merge if you can spare 5 minutes.

### Correctness

**1. `Curve::Exponential::refire_delta` ceiling at the floor boundary** — `tools/sensor-trait/src/curve.rs:135`. `(half_life * ratio).ceil()` returns `H+1` (not `H`) for `floor=0.5` due to f64 rounding when `H` is large enough that `H * 1.0` lands a hair above `H` after the `log2` round-trip. The unit test at line 276 happens to pass because `half_life=100` round-trips exactly. With `H=30000` the result is plausibly `30001` rather than `30000`. The session.rs comment at line 13-15 ("re-fires at exactly H ticks post-fire") is therefore technically off-by-one in worst case. No user-visible impact (delta of 1 token is invisible), but worth either pinning the test at a non-power-of-two H or documenting the boundary.

**2. `way_fire_outcome` and progression-axis monotonicity** — `tools/ways-cli/src/session.rs:285`. `get_token_position` reads from the newest transcript; if the transcript is rotated or replaced (subagent path, session split), the tick can move backwards relative to a persisted `EngagementState.last_fire`. `current_salience` uses `saturating_sub`, so a backward jump returns delta=0 → salience=1.0 → Suppressed forever. Probably already accounted for by the per-session state path, but worth a one-line check or assertion in `record_fire` that ticks are monotonic per (way, session). Same concern applies to attend's `epoch_secs()` if NTP steps the clock backward, though that's a much rarer real-world case.

**3. Phase B conversion edge case** — `tools/sensor-trait/src/engagement.rs:30` `rate_per_min_to_half_life_secs`. `(1.0 - rate).ln()` is negative for `0 < rate < 1`, and `0.5_f64.ln()` is also negative, so the ratio is positive — good. But `rate = 0.999` gives a half_life of ~6 seconds, which the loader will then plug into a `multiplier_half_life` that makes the action-potential curve effectively a 6-second flat refractory. Not wrong, just operator-surprising; the docstring says "treat ≥1 as misconfig" but doesn't bound the upper-rate case where the user got 0.95+ working in a previous attend version. Consider clamping `rate >= 0.95` to `MAX` alongside the `>= 1.0` branch, or surfacing the resulting half_life in `attend config show` so the operator can spot a too-aggressive rate.

### Safety

**4. `check-post.sh:54` — `${way_dir#$WAYS_ROOT/}` is unquoted on the right** — bash's `${var#pattern}` treats the pattern as a glob; `$WAYS_ROOT` is unquoted so any glob metachars in the project path (`?`, `*`, `[`) would mis-strip the prefix. In practice `$WAYS_ROOT` is one of `${PROJECT_DIR}/.claude/ways` or `${HOME}/.claude/hooks/ways` — unlikely to contain glob chars, but `PROJECT_DIR` comes from `jq -r '.cwd // empty'` and is operator-controlled. Recommend either quoting (`way_id="${way_dir#"$WAYS_ROOT"/}"`) or sanity-checking the prefix matches before stripping. Same script: `find "$WAYS_ROOT" -type f -name postcheck.sh` is fine, no globs into find.

**5. `check-post.sh` postcheck stdin / shell-injection** — `INPUT` is captured via `INPUT=$(cat)` and passed verbatim to each postcheck via `printf '%s' "$INPUT" | "$postcheck"`. That's safe — no eval, no expansion into argv. The `tool_response` field never reaches a shell context. Good.

**6. `tune_curves::apply_one`** — `tools/ways-cli/src/cmd/tune_curves.rs:367`. Writes via `std::fs::write` which is non-atomic — a crash mid-write corrupts the way file. For a `--apply` path that the operator runs intentionally on dozens of files, consider write-to-temp-and-rename. Not a hot path, but the existing tests don't cover the partial-write case.

### Design observations

**7. `tune_curves::compute_cadences` deduplication may swallow legitimate same-tick re-fires** — `tune_curves.rs:223` calls `positions.dedup()` after sort. If two events log at identical token_position (same tool turn, batched fires), they collapse to one. Probably the intent (you can't re-fire on a zero delta), but worth a comment confirming that's deliberate and not just defensive.

**8. The new `way_redisclosed` filter in `load_fires`** — `tune_curves.rs:162`. The comment is correct: first-fire is `way_fired`, subsequent fires are `way_redisclosed`. I checked `cmd/show/mod.rs:52-57` and confirmed the FireOutcome → event-name path. But I didn't see where `way_redisclosed` is *emitted* — either I missed the call site or it's being emitted under a different label downstream. If you're shipping tune-curves with a filter for an event name that nothing currently writes, the cadence dataset will be permanently first-fire-only. Worth grepping `log_event` call sites once to confirm both names are emitted.

**9. `session.rs` is at 693 lines** — the way-quality flag fired on `tune_curves.rs` mid-session per your PR notes; `session.rs` is the next file in line for the same treatment. The natural seam is engagement (`way_fire_outcome`, `record_way_fire`, `load_engagement`, `save_engagement`, `engagement_path`, `way_refire_threshold_k`) into a `session/engagement.rs` submodule. Not blocking, but would close the SRP loop you opened with the lint.rs split.

**10. `EngagementState::current_salience` semantic split between callers is clean.** Confirmed: attend's `Curve::ActionPotential` has `salience_at = 1.0` always, and attend never reads `current_salience` — it uses `effective_threshold`/`should_fire` only. Ways' `Curve::Exponential` has `multiplier_at = 1.0` always, and ways uses only `current_salience`. The two callers compose against orthogonal facets of the same state. Good design — explicit comment in `curve.rs:74-79` confirms intent.

**11. `attend config lint` vs `ways lint` parity** — semantically aligned: same UNKNOWN/DEPRECATED/x- escape, same surgical line removal, same `--check` exit semantics. One divergence: ways lint's `remove_top_level_field` consumes indented continuation lines and rewinds trailing blanks (helpers.rs:182-197), while attend's `rewrite_without_lines` operates on a flat line-index set with no continuation logic (config_lint.rs:364). For attend's current schema (single-line scalars under known sections) this is fine, but if `engagement.x-experimental:` ever holds a block value, `attend config lint --fix` would leave the orphaned indented lines behind. Worth at least a TODO comment.

### Test coverage gaps

**12. `replace_curve_block` round-trip test is good but doesn't cover the rewind-trailing-blank edge** — `tune_curves.rs:482`. Your test exercises the happy path. Consider one additional case: a `curve:` block followed by a blank line and then another top-level key (`curve:\n  type: ...\n  half_life: 30000\n\nthreshold: 2.0`). The current logic at lines 426-428 rewinds blank lines, which is correct, but no test pins it.

**13. No test for `way_fire_outcome` first-fire vs re-fire vs suppressed.** I see the curve tests and the engagement state tests, but the integration at `session::way_fire_outcome` is exercised only by the in-flight Phase F observation. A unit test that constructs a fake session dir, runs FirstFire → record_fire → ReFire → record_fire → Suppressed at controlled tick offsets would be a 30-line addition and would lock the contract.

**14. `tune_curves::compute_cadences_groups_by_session` test is good but tests only the happy path.** No test for: empty fires, fires with identical token_position (dedup behavior — see #7), fires from a single session that doesn't meet `min_fires`. None are tautological — the existing tests do exercise real branches — just sparse.

### Nits

- `tools/sensor-trait/src/curve.rs:212` — `count_live_events` recomputes the `0.5_f64.powf` for each entry; if history grows, this is O(n) per `multiplier_at` call. `prune_decayed` keeps it bounded so this is fine in practice, just calling it out for the perf-curious.
- `session.rs:255-261` — `load_engagement` silently constructs a fresh state on JSON parse error. If a curve enum variant changes shape post-merge, every persisted file silently resets to the new default. Consider logging the parse failure once.
- `tune_curves.rs:391` — `unreachable!(\"tune_curves only suggests Curve::Exponential\")` is correct today; if the v2 ActionPotential path ever lands, the panic message makes the upgrade trivially greppable. Good.

---

**Overall**: solid PR. The two-caller engine split is the kind of architectural unification that usually leaves at least one caller bleeding semantic assumptions across the boundary; this one doesn't, and the comments at the curve seam (curve.rs:74-79, lib.rs:217-232) tell future-you exactly why. The Phase E tune-curves test coverage is thin but the math is sound and the line-surgery primitive is well-isolated.

The single thing I'd ask before merge: please confirm finding **#8** (that `way_redisclosed` events are actually being emitted somewhere, not just consumed). Everything else is follow-up material.

---

aaronsb:
Thanks for the thorough review! Working through the findings.

## Finding #8 — false positive (non-blocking, just setting the record straight)

`way_redisclosed` **is** emitted. The string literal doesn't appear directly in `log_event` because it's produced via a ternary:

\`\`\`rust
// tools/ways-cli/src/cmd/show/mod.rs:52-56
let outcome = session::way_fire_outcome(id, session_id, &curve);
if !outcome.is_allowed() {
    return Ok(String::new());
}
let is_redisclosure = outcome.is_redisclosure();
session::record_way_fire(id, session_id, &curve);

// ...

// line 133
("event", if is_redisclosure { "way_redisclosed" } else { "way_fired" }.to_string()),
\`\`\`

`outcome.is_redisclosure()` returns true exactly when the engine path produced `FireOutcome::ReFire`. So:
- First fire in a session → `outcome = FirstFire` → `is_redisclosure = false` → \`way_fired\` emitted
- Subsequent fire of the same way → `outcome = ReFire` → `is_redisclosure = true` → \`way_redisclosed\` emitted

Runtime evidence: during the Phase F observation session that I wrote up in \`observed-behavior.md\`, I watched \`way_redisclosed\` events with \`token_position\` land in \`~/.claude/stats/events.jsonl\` in real time. The events came from exactly this code path.

So `tune_curves::load_fires`'s filter accepting both event names is correct — it's what lets the subcommand pair consecutive same-session fires into meaningful deltas.

## Acting on the rest

No hard blockers, agreed. I'm landing the non-size findings as a follow-up commit or two before re-requesting review:

**Must / should / nice:**
- [x] check-post.sh:54 shell quoting
- [x] tune_curves::apply_one atomic writes (write-to-temp-and-rename)
- [x] way_fire_outcome backward-tick guard
- [x] Curve::Exponential::refire_delta ±1 fix
- [x] rate_per_min_to_half_life_secs clamp for rates >0.95

**Test gaps:**
- [x] way_fire_outcome FirstFire → ReFire → Suppressed contract test
- [x] replace_curve_block trailing-blank-rewind edge test

**Follow-up issues (size/structure, not blocking this PR):**
- session.rs at 693 lines — splitting along engagement-helpers seam, mirror of #51
- attend config_lint `rewrite_without_lines` indented-continuation parity with ways — latent, only triggers when attend grows a block-valued key

The checkboxes above are todo-shaped right now; will flip them as the follow-up commits land. A new session after merge will tackle the main.rs splits (#51 for attend, plus session.rs and the ways-cli main.rs mirror case) as a dedicated structural-refactor PR.

Issue #51, opened 2026-04-15
attend: split tools/attend/src/main.rs (2027 lines) into cmd/* modules


aaronsb:
## Context

`tools/attend/src/main.rs` has grown to **2027 lines** with **32 functions** — 2.5× the softwaredev/code/quality priority threshold (800 lines). The file accumulated through honest feature growth across ADRs 113, 114, 119, 121, 123, and each addition was locally small. The aggregate is now hard to navigate and makes PR diffs touch a single hot file.

The file was flagged by the quality way's postcheck during ADR-123 Phase 1 burst_window cleanup (commit 679b205). Split is out of scope for PR #49 — ADR-123 is a firing-dynamics change, not a layout refactor — but this issue captures the refactor while the seam lines are fresh.

## Why it's tractable

main.rs is already structured as a command dispatcher. `fn main()` (line 1910) is a flat match over subcommand strings that calls `cmd_*` functions. Each `cmd_*` function is almost fully self-contained with its own helpers co-located. The split does not need to discover boundaries — they're already visible in the function order.

Existing siblings (`config.rs`, `config_lint.rs`, `groups.rs`, `scenes.rs`, `state.rs`, `emit.rs`, `sensors/`) already cover the cross-cutting concerns. What's left in `main.rs` is per-command logic, which wants to live in a `cmd/` folder mirroring the ways-cli convention.

## Proposed module layout

Target: `main.rs` shrinks to ≤150 lines (just module decls + the argv dispatch match). Everything else moves into a new `cmd/` folder.

| New file | Moves from main.rs | Lines | Notes |
|---|---|---|---|
| `cmd/run.rs` | `cmd_run_with_catchup`, `collect_snapshot`, `DisclosureGovernor` struct + impl | ~425 | The hottest sensor-loop path. DisclosureGovernor is private to run so it lives here. |
| `cmd/inbox.rs` | `cmd_inbox`, `cmd_inbox_read`, `parse_signal`, `is_valid_signal_id`, `ParsedSignal` | ~230 | Inbox reader plus its signal-parsing helpers. |
| `cmd/peers.rs` | `cmd_peers` | ~55 | |
| `cmd/send.rs` | `cmd_send`, `identify_sender`, `detect_terminal`, `find_closest_peer` | ~200 | Sender identity helpers move with the command that uses them. |
[...]

PR #55, opened 2026-04-15
feat(attend): apply ADR-123 outward gate to sensor-peers

### LINKED ISSUE #22: attend sensor-peers: apply ADR-123 outward gate to signal presentation
Apply the ADR-123 outward gate (`Curve::salience_at`) to attend's signal presentation path. The engine landed across PRs #60–66 and the firing-dynamics unification (#49); ways is already a consumer. Attend's sensor-peers is the deferred second consumer called out explicitly in [ADR-123's Implementation Status](../blob/main/docs/architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md#implementation-status) and in [`salience.md`](../blob/main/docs/attend-and-monitor/salience.md#attends-deferred-application-still-to-ship).

ADR-121 is now `Superseded` (by ADR-123). The original issue was pre-unification and assumed a turn counter, per-signal metadata table, and curve math all had to be built. All of that is done. What's left is a **hook-up**, not an architecture change.

## What's missing in sensor-peers today

`tools/sensor-peers/src/lib.rs` gates presentation through `seen_signals: HashSet<String>` — once surfaced, a signal is suppressed forever (persisted across restarts via the `state`/`restore_state` checkpoint). That works for the bulk-blast case but cannot express:

1. **A new peer joining a focus group mid-session gets no backlog filter.** `mark_existing_signals_seen()` is all-or-nothing at startup. There's no "recent enough to surface, ancient ignored" decision.
2. **Signals that are replied to via the `re:` threading field (ADR-120) don't re-enter presentation.** `parse_signal` already extracts the re-id; there's nothing downstream that resets presentation state when a thread is still alive.
3. **Aging is not measured.** The seen/unseen bit has no notion of wall-clock delta since arrival; a 30-day-old signal and a 30-second-old signal look identical to the presentation path.

## What to build

Per-signal `EngagementState<Curve::Exponential>` in sensor-peers, keyed by signal id, consulted before emitting an observation, reset on `re:` reply.

- [ ] **Signal state module.** New small module (`tools/sensor-peers/src/salience.rs` or inline) owning a `HashMap<SignalId, EngagementState>`. Each entry is built with `Curve::Exponential { half_life: N }` from attend config. State persisted alongside `seen_signals` in the existing `state()`/`restore_state()` checkpoint — same file, new key prefix.
- [ ] **Tick source.** `sensor_trait::epoch_secs()` — already in production for the inward gate. No new clock, no new unit.
- [ ] **Presentation gate in `read_signals()`.** Before pushing a new observation to the return vec:
  ```rust
  let s = state.entry(signal_id).or_insert_with(|| new_exponential(half_life));
  if s.current_salience(epoch_secs()) < presentation_floor {
      continue;  // aged out — skip, don't emit
  }
  s.record_fire(epoch_secs(), 1.0);  // presenting it resets the salience clock
  ```
  The existing `seen_signals` set keeps its current role (first-time detection); salience layers on top for re-presentation decisions.
- [ ] **Re-engagement reset.** When `parse_signal` returns a `re:<id>` that matches a tracked signal, call `record_fire(epoch_secs(), 1.0)` on that signal's state so the thread-parent re-enters the presentation set. The re-id is already parsed — only the lookup-and-reset is new.
- [ ] **Config surface.** New `signals:` block in attend config:
  ```yaml
  signals:
    half_life_seconds: 1800   # 30 min default — subject to `attend tune`
    presentation_floor: 0.3
  ```
  Loader in `tools/attend/src/config.rs`; `attend config lint` gains the new keys (UNKNOWN/DEPRECATED semantics already in place from f360ea6).
- [ ] **Backlog filter on group join.** When the `extra_scan_dirs_fn` surfaces new directories mid-session (ADR-118), the salience gate naturally filters the backlog — signals older than the floor never emit. This replaces `mark_existing_signals_seen`'s all-or-nothing startup behavior.

## Out of scope

- Changes to the engine (`sensor-trait::Curve`, `EngagementState`) — done in Phases A–B of ADR-123.
- Changes to the disclosure governor (independent, unchanged).
- Per-signal-class half-lives — premature; one global value. `Curve::ActionPotential` is the escape hatch if per-class ever becomes needed.
- Changes to the 30-day disk-retention sweep in [`signals.md`](../blob/main/docs/attend-and-monitor/signals.md).
- Changes to the inward-gate engagement model (`peer_engagement_boost` stays as-is).

## Code references

- **Engine:** `tools/sensor-trait/src/curve.rs` — `Curve::Exponential`, `salience_at`, `EngagementState`, `record_fire`, `current_salience`.
- **Tick source:** `sensor_trait::epoch_secs()`.
- **Ways' mirror case:** `tools/ways-cli/src/session.rs` — `way_fire_outcome`, `REFIRE_FLOOR`, per-session state persistence at `{session_dir}/way-engagement/{way_id}.json`. The attend side will mirror this shape, keyed by signal id instead of way id, with the checkpoint file as the storage.
- **Surface to instrument:** `tools/sensor-peers/src/lib.rs::read_signals()` (~line 214) and the `parse_signal` consumer just downstream.
- **Config loader:** `tools/attend/src/config.rs` and `tools/attend/src/config_lint.rs`.

## Acceptance

- [ ] `attend run` on a session with old signals in the shared dir does not flood the presentation set on startup or on focus-group join.
- [ ] Re-engaging a faded signal via `re:<id>` from a peer brings it back into the presentation set (confirmed against a live two-session test).
- [ ] Unit tests for the salience gate covering: below floor suppresses, above floor passes through, `record_fire` resets, stale state rebuilt after checkpoint reload.
- [ ] `attend config lint` recognizes the `signals:` block; unknown sub-keys surface as warnings; `--fix` removes them.
- [ ] `salience.md` "deferred application" section is rewritten to describe the shipped behavior (the sketch in the "still to ship" subsection is the blueprint).
- [ ] No regression in existing sensor-peers behavior (focus filter, engagement boost, broadcast scan, checkpoint persistence).

## Follow-ups (out of scope here)

- `attend tune` surveying wall-clock turn cadence to auto-suggest `half_life_seconds` (mirror of `ways tune-curves` from Phase E).
- `attend status` visualizing current salience per tracked signal.

aaronsb:
Closes #22.

Bundle of three layers: (1) apply the ADR-123 outward gate to sensor-peers as the deferred second consumer (original scope of the PR), (2) code-review fix from the first review round, (3) the fresh-agent UX restoration work that live testing surfaced as necessary once the gate's backlog-admission side-effects revealed a regression. Telling the story as one PR because the UX work is not optional — without it, the correctness win of (1) comes at a usability cost that makes fresh agents ask "should I reply?" and fish for 50-char uuids in the filesystem. All three layers reinforce each other.

## Layer 1 — Engine hook-up (the original PR)

Applies `sensor_trait::{Curve::Exponential, EngagementState, salience_at, epoch_secs}` to `tools/sensor-peers/` as the second cross-tool consumer of the firing-dynamics engine. Ways was the first. The engine itself shipped across prior PRs and is not modified here.

- **New `tools/sensor-peers/src/salience.rs`** — `SignalSalience` struct holding one `EngagementState` per tracked signal, keyed by filename stem (the same shape `re:<id>` references in ADR-120 threaded replies, so reply resets are direct hash lookups). Methods: `gate(signal_id, arrival_tick, now)` seeds state with `record_fire(arrival_tick, 1.0)` on first call so freshly-scanned old backlog is gated as if it had been presented at its on-disk mtime; `reset(signal_id, now)` bumps parent state to 1.0 for threaded-reply resurfacing; `export_rows` / `import_row` for checkpoint persistence. Below-floor entries are pruned on the same call that suppressed them so the map stays bounded on long sessions. 8 unit tests.
- **Hook in `sensor-peers/src/lib.rs::read_signals`** — before each observation leaves the poll, an `extract_re_id` call resets any referenced parent, then the signal's own `gate()` check runs against its file mtime. Below-floor signals are suppressed *and* marked `seen_signals` so we don't re-evaluate them every poll.
- **Checkpoint persistence** — `export_state`/`import_state` extended with a `signal_salience` row key carrying `<id>\t<json>`. Old checkpoints without the key parse cleanly.
- **Config** — new `SignalsConfig { half_life_seconds, presentation_floor }` in `tools/attend/src/config.rs`, defaults `1800 / 0.3`. New `signals:` YAML block parser and default docs. `attend config lint` recognizes the block; unknown sub-keys surface as warnings; `x-*` escape and `--fix` removal work as in every other section.
- **Orchestrator wiring** — `peer_sensor.set_salience_params(...)` in `tools/attend/src/sensors/mod.rs`, mirroring `set_peer_activity_window`.
- **Docs** — `docs/attend-and-monitor/salience.md` and `configuration.md` rewritten to describe the shipped behavior.

## Layer 2 — Code review fix

From the first round of review on this PR:

- **`mark_existing_as_seen` removed entirely.** The legacy startup call pre-seeded `seen_signals` with every pre-existing file, short-circuiting the gate for the dominant `!catchup` flow. The review caught that the salience gate was dead code for the primary path. Removing the call makes the gate the sole arbiter in every mode. Recent pre-existing signals (within the salience window) now surface where they previously were silenced wholesale — correct, but a behavior change, documented in `salience.md` under a new "Replacing mark_existing_as_seen" subsection.
- **Self-pruning in `gate()`** — below-floor entries are dropped from the state map on the same call that suppressed them, bounding memory over long sessions.

## Layer 3 — Fresh-agent UX restoration

After live testing surfaced a real usability regression: the engine hook-up worked, but the resulting workflow pushed agents toward filesystem improvisation and permission-asking. Every fix in this layer is independently verifiable and mutually reinforcing.

### `attend reply` subcommand — remove the uuid from agent context

- **New `tools/sensor-peers/src/last_inbound.rs`** — persists the most-recent inbound signal id per own-session-id under `~/.cache/attend/state/<session>.last-inbound`. Written by `sensor-peers::read_signals` inside the observation-emit block, keyed on session id so concurrent attend processes don't collide. Filter: only signals originating from a cwd *different* from the observer's own working_dir are recorded — the own-session-id check only catches the *current* session uuid, so a previous incarnation of the same agent would otherwise pollute last_inbound and cause `attend reply` to auto-thread to the agent's own past self. Same-cwd matching is a reliable proxy for "self" given the one-agent-per-project norm.
- **`cmd_reply`** in `tools/attend/src/main.rs` — reads last_inbound for the current own-session-id, prepends `--re <id>`, and delegates to `cmd_send`. A caller never sees or touches the uuid; all routing flags (`--focus`, `--to`) flow through unchanged. Errors with a clear "use send instead" message if no prior inbound exists.

### `attend inbox` non-TTY output — make inbox actually useful for agents

`attend inbox` now detects `std::io::stdout().is_terminal()`. TTY path keeps the compact 6-column table for humans. Non-TTY path emits one block per message with labeled lines — full id, full body, no truncation — same idea as `ls` switching on pipe. Fixes the completely-unreadable table that agents were previously seeing through Bash tool capture.

### Self-reload banner visibility

Existing 10s mtime-based self-reload at `main.rs:249` previously emitted the misleading `[attend] restarted (unchanged)` after a binary hot-swap whenever sensor config was unchanged. Fix: pass `ATTEND_RELOADED_FROM` env var across the exec boundary so the new process prints an explicit `reloaded vX → vY — binary updated, state preserved across exec()` line instead. Bootstrap quirk: takes effect from the next rebuild onward because the binary that sets the env var has to exist first.

### Three-source skill rewrite

All three places that teach an agent about attend messaging were stale or invited filesystem improvisation. Rewritten in parallel so the guidance is consistent whenever an agent meets it:

1. **`skills/attend/SKILL.md`** — primer read at `/attend` invocation. Peer-messaging section collapsed to two commands (`send` / `reply`) with pick-by-intent rules. Deleted the `ls -t ~/.cache/attend/signals/` filesystem fallback entirely. Added explicit autonomy grant ("you don't need permission to reply"), "Silence is a valid reply" as a standalone callout, and "CLI is the whole interface" containment boundary.
2. **`tools/sensor-disclosure/src/disclosures/messaging.md`** — runtime reheat baked into sensor-disclosure. Rewritten to match the skill: new topic vs reply to a peer, CLI contract line, autonomy framing, silence still valid.
3. **`hooks/ways/softwaredev/environment/attend/attend.md`** — just-in-time way that fires via `commands: attend` frontmatter on every `attend` command. Rewritten to carry the same four points so the guidance is reinforced at tool-use time in addition to session start and runtime.

## Verification — clean-slate live test on the post-bundle binary

Live two-session acceptance test was run after cleanup (zero signals on disk, zero last_inbound state, both sides on fresh binary). Full transcript preserved in PR comments. Results:

- ✅ **Fresh agent steered to `attend reply` without coaching** — fresh temp session loaded SKILL.md, received one peer notification with zero backlog, and its first response was `attend reply "..."`. No `attend send` fallback, no id-fishing.
- ✅ **Autonomy — replies without asking permission** — temp responded directly with no "should I reply?" delay.
- ✅ **Silence-is-valid honored** — temp saw my confirmation message and explicitly said "no further action needed" instead of firing a ceremonial ack. When it *did* reply, it said "happy to chat or just swap status" — the silence norm landed without being brittle.
- ✅ **Id never enters agent context on either side** — both threaded replies verified in wire format (`re:<id>` field present) with neither agent having touched a uuid.
- ✅ **cwd-self filter works** — both sides threaded to the actual peer, not their own past session uuids.

Earlier less-clean rounds (documented in prior PR comments) also exercised: backlog-filter admission after `mark_existing_as_seen` removal, checkpoint restart roundtrip, the extract_re_id → reset path on both sides.

### Automated

- `cargo test --workspace` — all green. sensor-peers 19 → 30 (+11 tests across salience.rs and last_inbound.rs), attend 17 → 20 (+3 config + lint tests), sensor-trait unchanged.
- `make lint` (clippy with `-D warnings`) — clean.
- `bin/attend config lint` on live user config — clean.
- Synthetic "unknown key in signals: block" — correctly flagged.

## Remaining known gaps (not blocking merge)

- **Aged-backlog suppression (below floor) is unit-test-covered but not directly observed in the wild** — no signal old enough to fall below 0.3 at 1800s half-life during the live runs. The math is covered by `old_signal_below_floor_is_suppressed` in salience.rs.
- **`attend tune` doesn't derive the salience half-life yet** — tracked as a follow-up; the current 1800s default is a conservative first pick.
- **`attend status` doesn't visualize per-signal salience yet** — diagnostic, follow-up.

### DIFF
--- a/docs/attend-and-monitor/configuration.md
+++ b/docs/attend-and-monitor/configuration.md
@@ -57,6 +57,12 @@ engagement:
   decay_per_minute: 0.1      # rate at which elevated threshold returns to baseline
   peer_activity_window: 900  # sliding window for per-peer engagement boost
 
+# Presentation-layer aging for peer signals (ADR-121, unified in ADR-123)
+# Consumed by sensor-peers to suppress aged backlog and reset on `re:` replies
+signals:
+  half_life_seconds: 1800    # exponential decay half-life (30 min default)
+  presentation_floor: 0.3    # suppress below this salience
+
 # Background cleanup of the signals base
 cleanup:
   enabled: true              # master switch
@@ -129,6 +135,17 @@ The action potential model parameters. Governs per-sensor refractory behavior. S
 | `burst_window`        | *no runtime effect* (implicit via half-life)       |
 
 **Auto-tuning.** Run `attend tune` to derive these from real session history. See [`engagement.md`](engagement.md#tuning-with-attend-tune) for how tuning works and how the linear-derived `decay_per_minute` relates to the engine's exponential half-life.
+
+### `signals`
+
+Presentation-layer aging for peer signals. Consumed by `sensor-peers` to decide whether an about-to-be-emitted observation is loud enough to surface. See [`salience.md`](salience.md) for the full design and ADR-121 / ADR-123 for the decision trail.
+
+- **`half_life_seconds`** (u64, default 1800): exponential decay half-life. A signal's salience halves every `half_life_seconds` seconds of quiet since its arrival (or its last `record_fire` reset). 30 minutes is a conservative first value; subject to `attend tune` once survey coverage exists.
+- **`presentation_floor`** (f64, default 0.3): below-floor signals are suppressed from the peer observation stream, without being deleted from disk. The 30-day retention sweep under `cleanup` is independent and unaffected.
+
+The gate runs inside `sensor-peers::read_signals`, keyed by signal id (= filename stem — the same shape `re:<id>` references in ADR-120 threaded replies). A threaded reply automatically resets the parent signal's salience to 1.0 via `record_fire` at the reply's arrival tick, so thread-alive parents stay presentable for new observers joining the same shared signal dir.
+
+`attend config lint` validates both keys; unknown sub-keys surface as warnings and can be removed with `--fix`.
 
 ### `cleanup`
 
@@ -267,7 +284,7 @@ Use this to confirm your config will actually work before launching attend — a
 
 Attend's YAML parser is deliberately minimal — it's a hand-written subset that handles the specific shape described above, with no `serde` dependency. It supports:
 
-- Two-level section headers (`governor:`, `engagement:`, `sensors:`)
+- Two-level section headers (`governor:`, `engagement:`, `signals:`, `cleanup:`, `sensors:`)
 - Four-space indent for sensor properties
 - Inline arrays (`requires: [Bash(gh:*), Read]`, `watch: [cargo, mix]`)
 - Block-form lists on a new line (`requires:` followed by indented `- item` lines) for `requires:` and `watch:`
--- a/docs/attend-and-monitor/salience.md
+++ b/docs/attend-and-monitor/salience.md
@@ -2,7 +2,7 @@
 
 This page covers the **presentation-layer aging** mechanism: how a signal's visibility in the conversation fades as the progression axis advances, even while the signal file remains on disk. It's the cousin — not the opposite — of attend's inward-gate engagement model in [`engagement.md`](engagement.md). Both sides of the gate share a single engine; this page is the outward-side explainer.
 
-**Status.** The framing comes from [ADR-121](../architecture/system/ADR-121-salience-decay-for-signal-presentation-turn-based-exponential.md). The mechanism it describes was unified with attend's inward gate in [ADR-123](../architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md): both now consume the same `sensor_trait::Curve` type with the same `salience_at(delta)` query. **Ways is the first concrete implementation**, shipping cross-tool via ADR-123. **The attend sensor-peers application is still deferred** — sensor-peers emits signals as observations without consulting a salience gate today. This page describes the decision, points at the ways implementation as the canonical code reference, and sketches the attend-side path that's still pending.
+**Status.** The framing comes from [ADR-121](../architecture/system/ADR-121-salience-decay-for-signal-presentation-turn-based-exponential.md). The mechanism it describes was unified with attend's inward gate in [ADR-123](../architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md): both now consume the same `sensor_trait::Curve` type with the same `salience_at(delta)` query. **Ways was the first concrete implementation**, shipping cross-tool via ADR-123. **Attend's sensor-peers application now ships too** (issue #22) — each peer signal passes through a per-id `EngagementState<Curve::Exponential>` before emitting, so aged backlog fades without being deleted and threaded replies (`re:<id>`) reset the parent's salience. This page describes the decision and points at both production implementations as canonical references.
 
 ## The problem the outward gate solves
 
@@ -81,18 +81,51 @@ For ways with `Curve::Exponential { half_life: H }` and the 0.5 floor, re-fire h
 
 Per-way visualization in `ways list` and `ways rethink` uses `Curve::refire_delta(floor)` to render each row's bar and forecast position from its own threshold, not a shared global. See [ADR-123 §2](../architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md#2-pluggable-curve-as-first-class-parameter) for the curve types and [`engagement.md`](engagement.md) for the engine's other queries.
 
-## Attend's deferred application (still to ship)
+## Attend's concrete application (shipping now)
 
-Attend's sensor-peers path currently emits signals as observations without consulting an outward gate. Every poll that finds a new signal file surfaces it, regardless of whether the same signal has been surfaced before or how much wall-clock time has passed since its arrival. That's the case ADR-121 diagnosed, and ADR-123 made the fix path trivially available — the engine is already in place, the `Curve::Exponential` variant already exists in `sensor-trait`, the tick source (`epoch_secs()`) is already in production for the inward gate.
+sensor-peers consults a per-signal outward gate inside `read_signals` before each observation leaves the poll. The state lives in a small sibling module, `tools/sensor-peers/src/salience.rs`:
 
-What's missing is the *last mile*: a salience gate in `sensor-peers` that wraps each "about to present this signal" moment in a `state.current_salience(epoch_secs()) >= floor` check before emitting, and a per-signal `EngagementState` keyed by signal id (or by `(signal, agent)` for the multi-agent view). When that wiring lands:
+```rust
+// tools/sensor-peers/src/salience.rs
+pub struct SignalSalience {
+    states: HashMap<String, EngagementState>,  // keyed by signal id
+    half_life: TickDelta,                      // from attend's signals: config
+    floor: f64,
+}
 
-- Each signal carries an `arrival_tick` in wall-clock seconds (the `arrival_turn` of ADR-121, renamed for the progression-axis unification).
-- A `Curve::Exponential { half_life: N }` in attend's config controls the decay rate. Default proposed in ADR-121 was 20 turns; translated to seconds for attend it's roughly `half_life: 20 × median_turn_seconds`. Real value will come from `attend tune` survey, not the ADR-121 default.
-- A `presentation_floor: 0.3` (or whatever the operator tunes) controls when signals drop out of the presentation set.
-- Re-engagement (a reply, a reference via the `re:` threading field from ADR-120) calls `record_fire` on the signal's state, resetting salience to 1.0.
+impl SignalSalience {
+    pub fn gate(&mut self, signal_id: &str, arrival_tick: Tick, now: Tick) -> bool {
+        let state = self.states.entry(signal_id.to_string()).or_insert_with(|| {
+            let mut s = EngagementState::new(Curve::Exponential { half_life: self.half_life });
+            s.record_fire(arrival_tick, 1.0);
+            s
+        });
+        state.current_salience(now) >= self.floor
+    }
 
-The implementation is scoped to the peer sensor and a new small module for signal-state persistence. **No changes to the engine, the disclosure governor, the engagement model, or the signal file format** — this is purely a new consumer of an existing facility. That's the payoff from ADR-123's unification: the attend outward-gate application is now a hook-up, not an architecture change.
+    pub fn reset(&mut self, signal_id: &str, now: Tick) { ... }
+}
+```
+
+The key design points:
+
+- **Signal id** is the filename stem — the same shape `re:<id>` references in ADR-120 threaded replies, so reply resets are direct hash lookups. No rename table, no ownership bookkeeping.
+- **Arrival tick** comes from the file's on-disk mtime, not the first time this observer happens to scan the directory. A peer who joins a focus-group room mid-session sees old signals as already-decayed — the backlog-filter behavior ADR-121 designed, now working against real backlogs instead of hypotheticals.
+- **Re-engagement reset** runs unconditionally at scan time: if the content is a threaded 5-field signal (`re:<id>|`), the parent signal's `EngagementState` is bumped back to 1.0 at the current tick before the current signal's own gate check runs. The reset persists into the checkpoint so reconnection does not re-age the parent.
+- **Checkpoint persistence** uses the existing `sensor_trait::Sensor::export_state`/`import_state` wire format, adding a new `signal_salience` row key that carries `<signal_id>\t<json>`. Old checkpoints without the key parse cleanly — the sensor just starts fresh on those signals.
+- **No engine changes.** The full implementation is one new file in sensor-peers, one new config block, and ~30 lines of wiring inside `read_signals`. That's the payoff ADR-123 designed for.
+
+### Configuration
+
+Operators tune the gate through attend's `signals:` block:
+
+```yaml
+signals:
+  half_life_seconds: 1800   # exponential decay half-life (30 min default)
+  presentation_floor: 0.3   # suppress below this salience
+```
+
+Defaults are conservative first-value picks, subject to `attend tune` once survey coverage exists. `attend config lint` recognizes both keys; unknown sub-keys surface as warnings and can be removed with `--fix`.
 
 ### Why the session-length distribution still matters
 
@@ -102,9 +135,18 @@ ADR-121 grounded its half-life default in real session data:
 min=1  median=12  p75=45  p90=84  max=133  mean=32.6  turns
 ```
 
-This was turns, not seconds, under the ADR-121 framing. Under ADR-123 the shape of the argument is the same, the axis is different: attend should pick a wall-clock half-life that behaves reasonably against the same distribution when converted through the observed turn-cadence. `attend tune` already surveys turn cadence from real transcripts and derives engagement parameters; when the outward-gate path lands, tune can derive a second parameter — the salience half-life — from the same analysis.
+This was turns, not seconds, under the ADR-121 framing. Under ADR-123 the shape of the argument is the same, the axis is different: attend picks a wall-clock half-life that behaves reasonably against the same distribution when converted through the observed turn-cadence. `attend tune` already surveys turn cadence from real transcripts and derives engagement parameters; the outward-gate half-life is the natural next parameter for it to derive from the same analysis (tracked as a follow-up to issue #22).
 
-The intuition to preserve: **short sessions never see aging (nothing to prune), medium sessions see the oldest material drop out cleanly near the end, long sessions experience meaningful pruning.** Whatever wall-clock half-life attend eventually picks should produce this behavior against the live distribution, not against a one-off sweep.
+The intuition to preserve: **short sessions never see aging (nothing to prune), medium sessions see the oldest material drop out cleanly near the end, long sessions experience meaningful pruning.** Whatever wall-clock half-life attend eventually picks should produce this behavior against the live distribution, not against a one-off sweep. The 1800 s default is the placeholder — deliberately uncalibrated — until tune closes the loop.
+
+### Known v1 limitation: resurfacing already-presented parents
+
+The `seen_signals` invariant still prevents the same observer from surfacing the same signal twice, so a `re:` reply that resets the parent's salience does not cause the parent to re-appear in *this* observer's notification stream. The reset *does* affect:
+
+- Any other observer joining the same shared signal dir after the reply arrived.
+- Any future session of the same observer that re-reads the shared dir from checkpoint state.
+
+Widening the reset to also re-surface already-presented parents in the current observer is a targeted follow-up, not a v1 requirement — the primary ADR-121 win is backlog filtering on new-observer entry, which the above implementation fully delivers.
 
 ## Re-engagement resets
 
@@ -148,7 +190,8 @@ Preserved here for posterity. The arguments didn't change when the engine unifie
 - [`engagement.md`](engagement.md) — the inward-gate side of the same engine, attend-specific.
 - [`signals.md`](signals.md) — disk-side retention (the time-based bulk cousin).
 - [`loop.md`](loop.md) — where the presentation gate will sit in the attend tick loop when sensor-peers consumes it.
-- [`configuration.md`](configuration.md) — attend's config surface; salience-related keys will appear there when the sensor-peers gate lands.
-- **ADR-121** — the salience-decay decision record. Status under ADR-123: reframed in place; the ways implementation is the first concrete realization.
+- [`configuration.md`](configuration.md) — attend's config surface; the `signals:` block lives there alongside `engagement:`.
+- **ADR-121** — the salience-decay decision record. Status under ADR-123: reframed in place; ways was the first concrete realization, sensor-peers the second.
 - **ADR-123** — the progression-axis unification that made ADR-121's mechanism a cross-tool facility instead of an attend-local one.
-- **`tools/ways-cli/src/session.rs`** — the `way_fire_outcome` path is the first production consumer. Read it for a concrete implementation that the deferred attend-side path can mirror.
+- **`tools/ways-cli/src/session.rs`** — the `way_fire_outcome` path is the ways-side consumer, anchored to token position.
+- **`tools/sensor-peers/src/salience.rs`** — the attend-side consumer, anchored to wall-clock seconds. Mirror structure, same engine call.
--- a/docs/attend-and-monitor/salience.md
+++ b/docs/attend-and-monitor/salience.md
@@ -139,6 +139,10 @@ This was turns, not seconds, under the ADR-121 framing. Under ADR-123 the shape 
 
 The intuition to preserve: **short sessions never see aging (nothing to prune), medium sessions see the oldest material drop out cleanly near the end, long sessions experience meaningful pruning.** Whatever wall-clock half-life attend eventually picks should produce this behavior against the live distribution, not against a one-off sweep. The 1800 s default is the placeholder — deliberately uncalibrated — until tune closes the loop.
 
+### Replacing `mark_existing_as_seen`
+
+The legacy startup flow called `peer_sensor.mark_existing_as_seen(focus)` before the tick loop started, pre-populating `seen_signals` with every signal file already on disk. This prevented a startup blast but was all-or-nothing — old *and* recent pre-existing signals were silenced alike. The salience gate replaces that coarse filter: every pre-existing file now flows through `read_signals`, the gate evaluates each one against its on-disk mtime, and only below-floor signals are suppressed. Recent-but-pre-existing signals (e.g., a focus-group message from 5 minutes ago at startup) now surface where they previously would not. The `mark_existing_as_seen` method was removed entirely as part of issue #22.
+
 ### Known v1 limitation: resurfacing already-presented parents
 
 The `seen_signals` invariant still prevents the same observer from surfacing the same signal twice, so a `re:` reply that resets the parent's salience does not cause the parent to re-appear in *this* observer's notification stream. The reset *does* affect:
--- a/hooks/ways/softwaredev/environment/attend/attend.md
+++ b/hooks/ways/softwaredev/environment/attend/attend.md
@@ -13,6 +13,16 @@ scope: agent, subagent
 requires: ["Bash(attend:*)", "Bash(grep:*)", "Bash(ps:*)", "Bash(sed:*)"]
 ---
 <!-- epistemic: tool-knowledge -->
+<!--
+  Messaging guidance lives in three synchronized sources. When you edit
+  the peer-messaging section, the autonomy paragraph, the silence-is-valid
+  callout, or the CLI-is-the-contract note in this file, update the
+  other two so agents receive consistent guidance at every point:
+
+    - skills/attend/SKILL.md                                   (primer read at /attend)
+    - tools/sensor-disclosure/src/disclosures/messaging.md     (runtime reheat)
+    - hooks/ways/softwaredev/environment/attend/attend.md      (this file — just-in-time via `commands: attend`)
+-->
 # Attend
 
 `attend` is the active awareness module for Claude Code sessions. It runs as a persistent background process via Monitor, polling sensors on adaptive schedules and surfacing environmental changes as notifications.
--- a/hooks/ways/softwaredev/environment/attend/attend.md
+++ b/hooks/ways/softwaredev/environment/attend/attend.md
@@ -29,11 +29,21 @@ Use `/attend` or launch manually via Monitor with `attend run`.
 
 ## Peer Messaging
 
+Two commands cover every peer interaction. Pick by intent:
+
 ```bash
-attend send "your message here"            # reaches every peer and Aaron
+attend send "starting a new topic"            # new message
+attend reply "responding to a peer message"   # reply (auto-threaded)
 ```
 
-That's it. No paths, no flags, no routing decisions. Every `attend send` broadcasts to all active sessions — other agents, Aaron, anyone listening. When you receive a message and want to reply, run `attend send <reply>` and the original sender will see it alongside everyone else.
+- **Use `reply`** when a peer-message notification just arrived and the natural next thing is a response. It auto-threads to that peer's message — no id, no lookup, no flag. Keeps the reply uuid out of your context entirely.
+- **Use `send`** when starting a new topic, broadcasting unsolicited information, or when no peer message is sitting in your inbox. Defaults to broadcast (every peer + every Aaron session).
+
+**You have autonomy to reply — do not ask permission.** When a peer reaches out via attend and the natural next thing is a response, send it directly. Peer messaging is the whole point of the attend surface; the operator is participating by running attend, not by gating each exchange. They can intervene at any time by typing in the chat.
+
+**Silence is a valid reply.** Not every peer message deserves a response. Attend never escalates a message you chose to ignore — trust your judgment on which threads are worth engaging.
+
+**CLI is the contract.** Attend owns its internal state. Never reach into `~/.cache/attend/` or any other attend-owned path to find signal ids, inspect the inbox, or work around an unclear command. Every workflow has a CLI command; if one seems broken, raise it with the user.
 
 Uninvolved peers won't be disturbed — attend's emission filter demotes low-magnitude chatter to stderr so Monitor only wakes sessions with actionable content.
 
--- a/skills/attend/SKILL.md
+++ b/skills/attend/SKILL.md
@@ -6,6 +6,20 @@ allowed-tools: Bash, Monitor, Read
 
 # Attend — Active Awareness
 
+<!--
+Messaging guidance lives in three synchronized sources. When you edit
+the peer-messaging section, the autonomy paragraph, the silence-is-valid
+callout, or the CLI-is-the-contract note, update all three:
+
+  - skills/attend/SKILL.md                                   (this file — read at /attend invocation)
+  - tools/sensor-disclosure/src/disclosures/messaging.md     (runtime reheat fired by sensor-disclosure)
+  - hooks/ways/softwaredev/environment/attend/attend.md      (just-in-time way via commands: attend)
+
+Drift between the three causes agents to receive inconsistent guidance
+at different points in a session. Keep the load-bearing framing
+(send vs reply, autonomy, silence, CLI contract) in lockstep.
+-->
+
 ## Step 1: Pre-flight
 
 Check that `attend` is installed:
--- a/skills/attend/SKILL.md
+++ b/skills/attend/SKILL.md
@@ -43,29 +43,25 @@ All commands below are one-shot — run with Bash, not Monitor.
 
 ### Peer messaging
 
-Send defaults to your current scope (own project + focus groups). Always wrap the message in double quotes to prevent shell metacharacter expansion (`?`, `*`, `!`).
+Two commands cover every peer interaction:
 
 ```bash
-attend send "your message here"
-attend send --focus deploy "message to a focus group"
-attend send --broadcast "important announcement for all sessions"
-attend send --to /home/user/other-project "directed message"
+attend send "starting a new topic"                 # new message
+attend reply "responding to the last peer message" # reply (auto-threaded)
 ```
 
-Keep messages under ~400 characters. Peer notifications are delivered one-per-line by the Monitor and longer payloads get truncated in-flight. The full signal file is preserved on disk, so recipients can always `attend inbox <id>` to read the complete message — but the at-a-glance notification won't carry it. Concise is the design, not a workaround.
+**Pick by intent:**
 
-### Threaded replies
+- **Use `reply`** when a peer-message notification just arrived and the natural next thing is a response. It auto-threads to that peer's message — no id, no lookup, no flag. If no peer message is in your inbox, `reply` exits with a clear error telling you to use `send` instead.
+- **Use `send`** when starting a new topic, broadcasting unsolicited information, or when no peer message is sitting in your inbox to reply to. Defaults to broadcast (reaches every peer and every Aaron session). Use `--focus <name>` or `--to <path>` to narrow the scope.
 
-When you are replying to a peer signal whose id you know, use `--re <id>` so the reply is marked as threaded. This is not cosmetic — it drives ADR-121's salience reset: under the signal-salience gate, a threaded reply bumps the parent signal's salience back to 1.0 so it stays presentable for future observers joining the same dir.
+**You have autonomy over peer replies — do not ask permission to participate.** When a peer reaches out via attend and the natural next thing is a response, send it directly. Do *not* stop to ask the operator "should I reply?" before answering. Peer messaging is the whole point of the attend surface; the operator is participating by running attend, not by gating each exchange. They can intervene at any time by typing in the chat — absence of intervention is consent to the conversation.
 
-```bash
-attend send --re b6b4379e-261f-4c0b-88f9-8d06f8c8b224-1776230155 "ack — picking this up"
-attend send --re <id> --focus deploy "threaded + scoped to a group"
-```
+**Silence is a valid reply.** Attend never escalates a message you chose to ignore. Not every peer message deserves a response; trust your judgment on which threads are worth engaging. Brief acks, substantive replies, and no reply at all are all legitimate choices — pick the one that fits the moment, and do not feel pressured to answer for the sake of answering.
 
-The id is the filename stem of the peer's signal (everything before `.signal`). Find it via `attend inbox` (the second column is the id, possibly truncated — `ls -t ~/.cache/attend/signals/<your-project>/*.signal | head` is the fallback for the full id). Valid ids match `[A-Za-z0-9_-]+`; anything else is rejected.
+Always wrap the message in double quotes to prevent shell metacharacter expansion (`?`, `*`, `!`). Keep messages under ~400 characters — peer notifications are one-per-line and longer payloads get truncated in-flight.
 
-**Rule of thumb:** if the peer's message surfaced via a `message from <sender>` notification and you're about to reply to it rather than start a new topic, use `--re`. If you're introducing a new topic or broadcasting, plain `attend send` is correct.
+**CLI is the whole interface.** Attend owns its internal state (signal files, checkpoints, caches) in paths that are none of your concern. If a command seems broken or incomplete, raise it with the user — do not reach into `~/.cache/attend/`, `~/.config/attend/`, or any other attend-owned directory to work around it. Those paths are implementation details and can change at any time. The CLI is the contract.
 
 ### Focus groups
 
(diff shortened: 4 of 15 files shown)


aaronsb:
## Review: feat(attend): apply ADR-123 outward gate to sensor-peers

**Size**: 534 / -18 across 9 files. Within "standard review" tier. Took the file-by-file pass plus a cross-check of the engine semantics against ways' usage of the same API.

### Assessment

The hook-up is a clean consumer of the unified engine. The split into `salience.rs` is at the right seam (a self-contained per-id state map with three real entry points), and the four wiring touches (`config.rs` parse arm, `config_lint.rs` keys + section, `sensors/mod.rs` setter, `lib.rs` gate + reset + checkpoint) are each small and localized. The seven salience-module tests cover the math; the three attend-config tests cover the default + parse + lint-classification axes. The checkpoint format is forward/backward-compatible (unknown keys ignored on import, old checkpoints parse cleanly).

That said, two items deserve attention before merge — one **likely-blocking gap between the PR narrative and the runtime behavior**, one **non-blocking memory-growth concern**. Several smaller observations below.

---

### Issue 1 (blocking-tier — narrative mismatch): `mark_existing_as_seen` still pre-stuffs `seen_signals` at startup, bypassing the gate

**Location**: `tools/attend/src/sensors/mod.rs:124-126`, `tools/sensor-peers/src/lib.rs:194-226`, `tools/sensor-peers/src/lib.rs:267-271`

**Observation**: `mark_existing_as_seen()` is still called at startup whenever `catchup == false` (the normal `attend run` path). It walks the same `scan_dirs` set the gate would walk and inserts every existing `.signal` filename into `seen_signals`. The first thing `read_signals` does inside the per-entry loop is `if self.seen_signals.contains(&key) { continue; }` — which means **for the normal startup case, the salience gate never sees pre-existing files at all**. They are filtered by the older binary mechanism, not by the new aging mechanism.

The PR description says \"attend's sensor-peers is now the second [consumer]\" and \"a freshly-scanned old backlog signal is gated as already-decayed.\" Issue #22 explicitly framed the primary win as: \"A new peer joining a focus group mid-session gets no backlog filter. `mark_existing_signals_seen()` is all-or-nothing at startup. There's no 'recent enough to surface, ancient ignored' decision.\" Issue #22's wording also says: \"This replaces `mark_existing_signals_seen`'s all-or-nothing startup behavior.\"

**Where the gate IS actually exercised** (after this PR):
1. **`catchup == true` path** — mark_existing_as_seen is skipped, so the gate processes everything on disk. This was previously a flood; now it's filtered. Real win.
2. **Mid-session focus-group join** — the new room's dir enters `current_extra_scan_dirs()` after startup, so it was never visited by mark_existing_as_seen. The gate filters its backlog. Real win, and matches issue #22's primary motivating case.
3. **A fresh signal arriving after startup** — mtime is `now`, gate trivially passes. No behavior change vs old code.

**Where the gate is NOT exercised**:
4. **Normal `attend run` startup with stale signals already on disk in own_encoded / _broadcast / startup-time focus-group dirs** — pre-stuffed into seen_signals, gate is dead code for these.

That's not necessarily wrong as a design choice. The catchup-mode fix and the mid-session join fix are genuinely valuable and clear issue #22's acceptance criteria #1 (\"`attend run` on a session with old signals in the shared dir does not flood the presentation set on startup\") **only because mark_existing_as_seen was already doing that job pre-PR**. The gate adds the mid-session-join case, which mark_existing_as_seen could not handle.

**Three options, in order of preference**:

a) **Drop `mark_existing_as_seen` for the non-catchup path** (issue #22's literal reading: \"This replaces...\"). The gate is strictly more capable — it does exactly what mark_existing_as_seen does for files older than `presentation_floor` worth of half-lives, and additionally lets a recent-but-not-yet-seen signal through. Removing the call simplifies the contract: there's one mechanism for backlog filtering, not two. The risk is a startup-time burst if someone has a large _broadcast dir with many fresh-mtime signals, but those would have surfaced anyway in the old behavior with `--catchup`.

b) **Keep the call but document the layered semantics honestly in the PR body and in `salience.md`'s \"v1 limitations\" subsection**. Specifically: \"The gate primarily handles the mid-session focus-group join case and the explicit `--catchup` path; normal `attend run` startup continues to suppress all pre-existing signals via `mark_existing_as_seen`.\" This is the lowest-risk option but leaves the duplication.

c) **Make it config-driven** — a `signals.startup_mode: gate | mark_seen` switch. Probably overengineered for one bool of behavior.

**Why I'd push for (a)**: the PR's own framing (\"sensor-peers is now the second consumer of the outward gate\") implies a single gate. Right now there are two, and the new one is silently bypassed in the dominant code path. A reader of the salience module would reasonably expect it to be load-bearing on startup; it isn't.

**Recommended minimum**: at least update `salience.md`'s \"Known v1 limitation\" subsection to add a second bullet calling this out, and update the PR description's \"What I deliberately did *not* do\" to mention that mark_existing_as_seen was deliberately retained. Right now the PR reads like it replaced the old mechanism when it actually layered alongside it.

---

### Issue 2 (non-blocking — bounded memory hygiene): `signal_salience.states` grows unbounded

**Location**: `tools/sensor-peers/src/salience.rs:23-27`

**Observation**: `SignalSalience::states` is a `HashMap<String, EngagementState>` with no pruning path. Every signal id ever observed (or ever reset via a `re:` reply) lives in the map for the lifetime of the process and is persisted into the checkpoint indefinitely. A long-running attend session in a busy focus group accrues an entry per unique signal forever.

`EngagementState::record_fire` does call `prune_decayed` internally, so the per-entry memory is bounded at ~tens of bytes. But the entry itself never goes away. After a week of sustained activity the checkpoint write is doing O(unique-signals-ever-seen) JSON serializations on every snapshot.

**Why it's not blocking**: each entry is small, the snapshot frequency is low, and a typical signal id is short. It will not break anything in v1.

**Suggestion**: drop entries whose `current_salience(now)` has been below the floor for some grace period (e.g. 10× half_life). One natural seam is inside `gate()` — if the entry already exists and its salience is below floor, return false and remove. That has the side benefit of making the gate self-cleaning without a separate sweep. Or run a sweep inside `export_rows()` before serialization.

**Counter to the suggestion**: dropping below-floor entries means a future `re:<id>` reset for an aged-out signal would create a new state instead of finding the existing one. With `reset()`'s lazy-create that's actually fine — the parent's state was below floor anyway, the reset overwrites it to 1.0, and downstream behavior is identical. So the cleanup is semantically transparent.

---

### Issue 3 (correctness check — pass): seed-on-first-call semantics vs. ways' usage

**Location**: `tools/sensor-peers/src/salience.rs:55-63`

**Observation**: I traced this against `tools/sensor-trait/src/engagement.rs:88-92` (`current_salience` = `salience_at(now.saturating_sub(last_fire))`) and against ways' usage in `tools/ways-cli/src/session.rs:294`. Two notable behavioral differences from ways, both deliberate-looking but worth confirming:

1. **Floor inclusivity**. `gate()` uses `>= self.floor` (line 62); ways uses `< REFIRE_FLOOR` for the re-fire decision. At exactly `floor`, sensor-peers presents and ways suppresses. Trivial in floating-point practice but the asymmetry is real.

2. **First-fire treatment**. Ways treats `last_fire = None` as `FirstFire` (always-fire) — see `session.rs:292`. Sensor-peers seeds the state with `record_fire(arrival_tick, 1.0)` on first observation, so the engine is *never* in the `last_fire = None` branch. That's the intentional difference: sensor-peers wants a backlog signal scanned from a 30-day-old mtime to enter the gate as already-decayed, where ways wants every brand-new way to fire once. Different problems, different policies — the divergence is correct.

The future-mtime case (clock skew / NFS) is safe because `current_salience` uses `saturating_sub`: an arrival_tick > now collapses to delta=0, salience=1.0, gate passes. Ways has the same protection (`session.rs:330-334`).

**No issue here** — flagging only because the divergence is non-obvious to a reader who knows the ways pattern. A 2-line comment in `salience.rs:55-63` saying \"unlike ways, sensor-peers seeds on first call so old backlog files enter the gate already-decayed\" would help future readers. (You already say this in the doc comment on line 50-54, but the contrast with ways isn't drawn explicitly.)

---

### Issue 4 (correctness check — pass with one ordering note): gate ordering vs. own-session skip

**Location**: `tools/sensor-peers/src/lib.rs:288-321`

**Observation**: The current order inside the per-entry loop is:
1. `seen_signals` early-return (line 269)
2. Read file
3. `extract_re_id` → `signal_salience.reset(re_id, now_tick)` (line 288-290)
4. `signal_salience.gate(signal_id, ...)` (line 305) — own-session signals seed the map and consume gate budget
5. `parse_signal` (line 313)
6. Own-session check on `from` field (line 317) — only here do we skip our own

**Implication**: our own session's signals seed entries in the salience map (with arrival_tick = file mtime ≈ now) and pass through the gate every time. They take map space, get checkpointed, and the gate computation runs on them. Functionally harmless — the gate trivially passes for fresh own-mtimes and the map entry is small — but it's wasted work and a subtle correctness wart: the salience map contains entries for signals we never present to ourselves.

A reply we send (`attend send re:foo bar`) would also call `reset(\"foo\", now)` on our own parent before being skipped at step 6. That's actually **probably what you want** — we replied, the thread is alive, the parent should be loud for any peer who later joins the same dir. But this is non-obvious: the comment on line 281-287 says \"resets the parent's salience back to 1.0 at now_tick, regardless of whether we're going to present the reply itself,\" which is true, but doesn't address whether we want our *own outgoing* replies to reset the parent. If yes (and I think the answer is yes — symmetry with peers replying to us), the comment should say so explicitly.

**Lighter alternative**: move the own-session skip to step 3.5 (after parse, before gate), so our own signals neither seed nor reset. Cleaner but loses the \"reply resets parent\" symmetry. Probably not what you want.

**Recommendation**: keep the current order. Update the comment at line 281-287 to add: \"This applies to our own outgoing replies as well — by the time we reach the own-session check at line 317, the reset has already fired. That symmetry is intentional: replying *is* engagement, and any peer joining the dir later should see the parent as loud-at-our-reply-tick.\"

---

### Issue 5 (correctness — pass): tab separator in checkpoint wire format

**Location**: `tools/sensor-peers/src/lib.rs:665-666`, `tools/sensor-peers/src/lib.rs:677-683`, `tools/sensor-peers/src/salience.rs:153-158`

**Observation**: The PR asks me to confirm the `\\t` separator is robust. Verified:
- Signal id charset is `[A-Za-z0-9_-]+` per `is_valid_signal_id` in both `lib.rs:933` and `salience.rs:153`. No tab possible.
- `serde_json::to_string` (compact) emits no literal tabs — string values containing tabs would be escaped as `\\t` (two chars), not as a literal tab. So the JSON value side is also tab-free.
- `split_once('\\t')` finds the first occurrence; with neither side containing a literal tab, it cleanly partitions.

**Defensive nit**: the contract relies on two invariants (id charset + compact JSON). Both are stable but neither is asserted at the boundary. A 1-line debug_assert in `export_rows` (`debug_assert!(!id.contains('\\t') && !json.contains('\\t'))`) would catch a future code change that breaks either invariant. Optional.

---

### Issue 6 (correctness check — pass): `reset()` lazy-create on missing id

**Location**: `tools/sensor-peers/src/salience.rs:82-89`

**Observation**: If a `re:<id>` arrives for a parent we've never tracked, `reset()` creates an empty state and immediately fires. Could that mask a typo in re-id or a wrong observer?

**Verdict**: it's the right behavior, but the failure mode is worth knowing. If a peer typos `re:nonexistent-id`, sensor-peers will silently create a salience entry for that nonexistent id and never bind it to a real file. The entry will sit in the map until pruning (see Issue 2) cleans it up. No user-visible misbehavior, but it does mean the salience map can grow from typos as well as real signals.

The alternative — gate `reset()` on \"only reset if the id already exists in the map\" — would lose the \"new observer joining the room\" case the PR is built to support: a fresh observer hasn't seen the parent yet, so its map is empty, so the reset would be silently dropped, and the parent would not be resurfaced for that observer. That's the wrong behavior. Lazy-create is correct.

**No change needed**. The typo-pollution risk is real but bounded by Issue 2's pruning suggestion.

---

### Issue 7 (test gap — flag, not block): no end-to-end test for the read_signals hook

**Observation**: All the new tests are pure unit tests of `SignalSalience`, plus three for the config parser. The actual integration — \"a stale .signal file in a scan dir is filtered by `read_signals`\" — is only verified by `cargo test --workspace` succeeding on the existing test suite. No test asserts \"create a temp dir, drop a signal file with mtime 4000s ago, call read_signals, observe nothing returned.\"

**Why it's a real gap**: the `read_signals` hook is where Issue 1 (mark_existing_as_seen interaction), Issue 4 (own-session ordering), and the salience-vs-seen_signals interleaving all live. The salience module tests pass cleanly even though the gate is a no-op for the normal startup case (Issue 1). An end-to-end test that exercises `read_signals` against a temp directory with old + new + own + threaded signals would catch the layering issue I'm flagging in #1.

**Suggestion**: one new test in `lib.rs` (or a new `tests/` integration file) that:
- Creates a temp signals dir
- Drops one fresh signal (mtime = now), one old signal (mtime = now - 10000), one own-session signal
- Calls `read_signals` directly (with a manufactured `Focus`)
- Asserts: the fresh non-own signal is returned, the old signal is not, the own signal is not, and `seen_signals` reflects all three

Such a test would have caught the dead-gate situation in Issue 1, because it would assert \"old signal is filtered\" without going through `mark_existing_as_seen` — and would either confirm the gate is doing the work, or flag that something else is.

**Not blocking** because the salience math is well-tested in isolation and the existing workspace tests catch regressions in the unchanged paths. Worth tracking as a follow-up.

---

### Issue 8 (file hygiene — observation, not a block): `lib.rs` grew from 1054 → 1129

**Location**: `tools/sensor-peers/src/lib.rs`

**Observation**: This file is on the priority code-quality list. The PR added 75 lines net to it. The mitigation — extracting `SignalSalience` into a sibling module — was the right call and absorbed the bulk of the new code (261 lines in `salience.rs`). But `lib.rs` is now 1129 lines and continues to grow with each consumer addition.

The natural next seam (out of scope for this PR but worth flagging for a future cleanup): the `read_signals` method is ~165 lines and is the obvious extraction target. It owns the scan-dir enumeration, the per-entry filtering pipeline, the parse-and-format logic, the chunking, and the cleanup-on-old-mtime side effect. Each is a candidate for a private helper.

**Not blocking this PR**, but the file is going to need a real split soon. A focused refactor that pulls `read_signals` out into a `signals_loop.rs` module with private helpers for the scan/gate/parse/format stages would cut lib.rs by ~200 lines and make the gate's role in the pipeline more obvious.

---

### Issue 9 (small things)

- **`salience.rs:117`**: trailing blank line inside the `impl` block (between `import_row` and the closing `}`). Minor style nit.
- **`salience.rs:152`**: \"Mirror of the `is_valid_signal_id` helper in `lib.rs`. Kept in sync by convention.\" Two copies of the same fence is a hazard. Consider promoting to `pub(crate) fn is_valid_signal_id` in `lib.rs` and importing from `salience.rs`, or moving it to `sensor_trait` if it's general. \"Kept in sync by convention\" is the least-safe option.
- **`salience.rs:189-201`**: `reset_restores_salience_to_one` has a subtle test artifact in its assertion. The comment says \"At reset tick, a subsequent gate call would see salience 1.0\" and then asserts `gate(\"id-1\", 1_000_000, 1_004_000)` returns true. This works because gate finds the existing entry and skips the seed branch — the `arrival_tick` argument (1_000_000) is silently ignored when the entry already exists. That's correct behavior but the test passes for a non-obvious reason. A clearer assertion would be a direct `state.current_salience(1_004_000) >= 0.3` check via a `pub(crate)` accessor, or the test could split into two: one for \"reset overwrites\", one for \"gate ignores arrival_tick on second call.\"
- **`config.rs:486-501`**: the new `signals` arm is correctly placed in the indent==2 dispatch alongside `governor`/`cleanup`/`engagement`/`sensors`. Good.
- **`config_lint.rs:17`**: `SECTIONS` correctly includes `\"signals\"`. Good.
- **`sensors/mod.rs:120-123`**: setter is wired next to `set_peer_activity_window`, mirroring the existing pattern. Good.

---

### Requirement traceability (against issue #22 acceptance)

| Criterion | Status |
|---|---|
| `attend run` on a session with old signals in the shared dir does not flood the presentation set on startup or on focus-group join | **Partial** — the focus-group-join half is delivered cleanly. The startup-flood half is still served by `mark_existing_as_seen`, not by the gate. See Issue 1. |
| Re-engaging a faded signal via `re:<id>` from a peer brings it back into the presentation set (confirmed against a live two-session test) | **Partial** — the salience-map reset is correct and tested. The honest v1 limitation in salience.md acknowledges that the same observer can't re-surface an already-presented parent. The PR body says no live two-session test was run. Issue #22 explicitly asks for that confirmation. |
| Unit tests for the salience gate covering: below floor suppresses, above floor passes through, `record_fire` resets, stale state rebuilt after checkpoint reload | **Met** — all four covered in `salience.rs` tests. |
| `attend config lint` recognizes the `signals:` block; unknown sub-keys surface as warnings; `--fix` removes them | **Met** — verified via `SECTIONS` update, `SIGNALS_KEYS` definition, the `\"signals\"` arm in `classify_section_key`, and the new `classify_signals_keys` test. |
| `salience.md` \"deferred application\" section is rewritten | **Met** (per PR description). |
| No regression in existing sensor-peers behavior | **Likely met** — workspace tests pass, but Issue 7's test gap means there's no direct assertion. |

---

### Bottom line

The engine consumer is sound and the math is right. Two things would make this PR ready to merge as-described:

1. **Resolve the mark_existing_as_seen / gate layering** (Issue 1): either remove the call so the gate is the sole startup mechanism, or update the PR body and salience.md to honestly document that the gate is layered on top and primarily serves the mid-session-join + catchup paths.
2. **Run the live two-session test** that issue #22 asks for, or downgrade that acceptance bullet to a follow-up explicitly. The PR body offers to run it on request — I'd take you up on that before merge.

The memory-growth concern (Issue 2) and the test gap (Issue 7) are non-blocking but worth tracking as follow-ups in the same issue thread.

Everything else is clean.

---

*AI-assisted review via Claude (Opus 4.6).*

aaronsb:
Addressed the blocking finding from the review in c918b91.

**Fixed:**

1. **`mark_existing_as_seen` removed.** The non-catchup pre-seed call was the main bug — it pre-populated `seen_signals` before the tick loop started, so the `seen_signals.contains` early return at `read_signals` short-circuited the salience gate for every pre-existing file. Fresh session starts now run all pre-existing files through the gate in every mode. The now-unused method is deleted from `PeerSensor` entirely rather than retained as a compat shim. Recent-but-pre-existing signals (e.g., a 5-min-old focus-group message at startup) now surface where they previously would not — this is a user-visible behavior change and I documented it in `salience.md` under a new "Replacing mark_existing_as_seen" subsection.

2. **Self-pruning in `SignalSalience::gate`.** Below-floor entries are removed from the state map on the same call that suppressed them. The `seen_signals` invariant prevents re-gating the same file from a later scan, so the prune is safe — a pruned state cannot leak back. One new unit test (`below_floor_gate_prunes_state_entry`) covers the prune path. Sensor-peers test count: 26 → 27.

**Not addressed in this commit (deliberate):**

3. **Live two-session acceptance test.** Still not run. Want me to exercise it before you merge, or are you comfortable with the unit-test coverage as the pre-merge bar?

4. **`read_signals` integration test.** The reviewer's right that the hook isn't exercised end-to-end — only unit-tested at the `SignalSalience` boundary. I'd like to punt this to a follow-up because testing `read_signals` well requires temp-dir fixtures for the signals base and the sessions/projects dirs, and the file's existing 500+ lines of tests don't have that infrastructure yet. I'd rather build that once and use it for multiple follow-ups than bolt it onto this PR.

5. **Cosmetic findings** (trailing blank line in salience.rs:117, `is_valid_signal_id` duplication, the reset-test's gate-based assertion). I'll let you tell me whether to bundle these into this PR or skip.

**Verification after the fix:**
- `make lint` — clean
- `cargo test --workspace` — all green (sensor-peers 27, attend 20)
- `bin/attend config lint` on live user config — 0 errors, 0 warnings

Ready for re-review when you have a minute.

aaronsb:
## Live two-session verification (partial follow-up on review item #3)

Exercised the salience gate against a real peer exchange with a separate Claude session in `/home/aaron/temp`. Both sides running the same post-fix (c918b91) binary. Attend launched via Monitor on this side; temp session was launched independently by the operator.

**Round 1:**
- I broadcast → temp surfaced it cleanly (fresh arrival, gate admits with salience 1.0)
- temp replied (legacy 4-field signal) → I surfaced it (gate seeded fresh state, admit)

**Round 2:**
- I replied with `--re b6b4379e…` (threaded) → temp surfaced it; temp's `read_signals` hit `extract_re_id` → `signal_salience.reset("b6b4379e…", now_tick)` on the parent signal → confirmed by temp agent's explicit "reset path exercised on my side too" ack
- temp replied (legacy) → I surfaced it

**Verified end-to-end:**
- ✅ Gate admits fresh arrivals without regression (both directions, both rounds)
- ✅ `extract_re_id` correctly identifies a threaded reply and pulls the id
- ✅ `SignalSalience::reset` fires against the parent id (temp agent confirmed)
- ✅ Persistent Monitor + attend loop handled 4 round-trip signals over ~2 min without any errors, dropped notifications, or governor stalls

**Not exercised:**
- ❌ My local side's reset path — temp's replies were both legacy format, so my `extract_re_id` returned `None` every time. Symmetry with the temp-side reset is inferred from "both binaries are identical" rather than directly observed.
- ❌ Aged-backlog suppression — no pre-existing signal in the shared dir was old enough to sit below the 0.3 floor at a 1800 s half-life during the test. The backlog-filter behavior is unit-test-verified (`old_signal_below_floor_is_suppressed` in salience.rs) but not observed in this live run.
- ❌ Checkpoint roundtrip across restart — the attend process stayed up throughout.

Updating acceptance criterion #2 to "mostly met": the live test confirms no regression and that the reset path runs cleanly, but did not symmetrically exercise both reset directions or observe an aged-backlog suppression in the wild. Happy to run a second pass targeting the backlog case (pre-populate the shared dir with a stale signal, restart attend, check it doesn't surface) if you want that before merge.

aaronsb:
## Live verification round two — now with symmetric evidence

Ran the acceptance test again after both sides rebuilt to the post-review commit. This round cleared the gaps from round one.

**Scenario:** I stopped my attend, the operator ran `make install` (which rebuilt and symlinked `~/.local/bin/attend`), then both sides restarted their Monitors — temp first in a fresh Claude session at `~/temp`, then me in this session. By the time temp came up, there were 4 already-written signals sitting in the shared dirs from the prior round and my re-engage probe in between.

**What temp's fresh session reported:**
> "confirmed — your re-engage arrived via attend on temp side, four peer messages surfaced in one batch (wrap + late-ack echo + this re-engage). fresh traffic is flowing."

Four backlog signals surfaced at once on temp's startup — all within the 1800 s / 0.3 floor salience window (oldest was ~13 min, salience ≈ 0.74). **Under the pre-PR-55 binary, `mark_existing_as_seen` would have silenced every single one of those.** Under the post-fix binary, they surfaced because the gate is the sole arbiter and each one was loud enough.

### Coverage delta vs round one

| Criterion | Round one | Round two |
|---|---|---|
| Fresh-traffic gate admission | ✅ both directions | ✅ both directions |
| `--re` + `extract_re_id` + `reset` path (one side) | ✅ temp side | ✅ both sides — my side surfaced the probe via `reset()` on the fresh-session's empty map |
| Backlog-filter behavior in the wild | ❌ nothing stale enough | ✅ 4-signal backlog, all within window, all surfaced (confirms `mark_existing_as_seen` removal) |
| Aged-backlog suppression (below floor) | ❌ nothing that stale | ❌ still not exercised — next run with an older pre-planted signal would close this |
| Checkpoint restart roundtrip | ❌ process stayed up | ✅ both sides restarted; state rebuilt cleanly from disk |

Acceptance criterion #2 ("live two-session test") is now met for every case except the "signal so old it's below floor" edge — I don't consider that blocking because the math is unit-test-covered (`old_signal_below_floor_is_suppressed`) and the shape of the backlog-admission success confirms the symmetric case.

### Incidental finding (not a PR blocker, follow-up)

Temp's operator noted that `attend inbox` is unreadable when piped through Claude's Bash tool — the table renderer auto-fits to ~80 chars, clipping uuids to `680c3057-…` and message bodies to 3 chars. The skill update I landed earlier (0d1acdf) already points at `ls -t ~/.cache/attend/signals/<dir>/*.signal | head` as the fallback path, but the real fix is a non-TTY output mode in `attend inbox` itself (emit full rows when stdout is a pipe, same way `ls` drops columns). Want me to file that separately?

### Ready for merge?

From the code side: lint clean, workspace tests green, both rounds of live testing validate the happy paths and the backlog-filter behavior that was the primary goal. Waiting on your go/no-go.

aaronsb:
## Re-review: f63b1b1 UX layer

Scoped to the fresh-agent UX work added since the first pass (c918b91 engine fixes already reviewed and resolved). Summary: the layer is well-factored, the cwd-self filter is the right heuristic for the stated threat model, and the triple-source skill rewrite is defensible as intentional redundancy. Four non-blocking observations and two follow-ups below. **No blockers.**

---

### What this layer accomplishes

Takes a correctness win (the gate admitting pre-existing backlog) and restores the ergonomics that the live test showed were actually load-bearing for fresh agents. Three independent mechanisms that reinforce each other:

1. `attend reply` removes the signal uuid from agent context entirely.
2. `attend inbox` non-TTY branch makes the inbox readable when an agent reads it through Bash.
3. Skill/sensor-disclosure/way rewrites delete the filesystem-improvisation path and reinforce "CLI is the contract" at three loading moments (session start, tool use, runtime reheat).

Each by itself would have been insufficient. Together they close a real regression, and the PR body is honest about the scope creep.

---

### Finding 1 — `source_cwd != focus.working_dir` is the right filter (non-blocking)

**Location**: `tools/sensor-peers/src/lib.rs:364-368`

**Assessment**: The cwd-self filter is a defensible proxy for "not me" given the one-agent-per-project norm. The two realistic failure modes are:

- **Two legitimate Claude sessions in the same cwd by accident.** In that edge case, one session's inbound would be filtered from the other's `last_inbound`. The user would see the peer notification, attempt `attend reply`, and either get the "no prior inbound" error or thread to a different peer's message. Neither failure is silent — the `attend reply` delegates to `cmd_send` which writes a signal file, and the operator can see on the receiving end that the thread id is wrong or missing. Recoverable via `attend send` + manual context.
- **Symlinked or bind-mounted project directory.** Path strings would not match even though the cwd is semantically the same. Recording would proceed and `attend reply` would auto-thread to self. Same recovery path (operator notices the weird thread behavior, corrects).

Both failure modes are observable, not silent, and neither causes correctness issues in the wire format. The assumption is sound for v1 — document the invariant as "one attend agent per canonicalized cwd" somewhere visible so a future contributor doesn't accidentally break it.

**Suggestion (follow-up)**: Add a one-line comment near `main.rs::own_session_id()` or the peer sensor docstring: "cwd-self filter in `last_inbound::record` assumes at most one attend-backed agent per project cwd — if this invariant changes, the reply auto-threading will need a sharper identity check (e.g. resolved canonical path + process start time)." Would prevent a future refactor from silently landing the brittle case.

**Disposition**: non-blocking. The current heuristic is correct for the shipped model.

---

### Finding 2 — Plain-text on-disk format is appropriate (non-blocking)

**Location**: `tools/sensor-peers/src/last_inbound.rs` (entire file)

**Assessment**: Plain text is the right call. The file holds one short ASCII-safe id. Any future field you'd want to add (timestamp, source cwd, scope) is already recoverable by reading the signal file itself via `<base>/<scope>/<id>.signal`. JSON would be over-engineering — you'd pay a parse cost every read and gain nothing the on-disk signals don't already provide.

One minor non-blocking nit: `fs::write` is not strictly atomic on POSIX. For a short id this is almost certainly fine (the write typically goes out in a single syscall), but if you ever see a corrupted read in the wild, switch to write-then-rename as `cmd_send` already does for signal files:

```rust
let tmp = p.with_extension("last-inbound.tmp");
fs::write(&tmp, signal_id).ok();
fs::rename(&tmp, &p).ok();
```

Two-line change, guarantees read-side never sees a partial write. Not worth doing now without a concrete reproducer.

**Disposition**: non-blocking. File format is correctly sized for the job.

---

### Finding 3 — `cmd_reply` delegating to `cmd_send` via argv is clean (non-blocking)

**Location**: `tools/attend/src/main.rs:976-1002`

**Assessment**: The thin-wrapper delegation is defensible given that `main.rs` is already at 2140 lines and flagged as priority territory under issue #51. Extracting a shared `SendOptions` struct + helper function would be the textbook refactor, but it would grow this PR's diff, touch code that's already on the split shortlist, and distract from the UX-restoration framing. Ship the thin wrapper now; do the extraction as part of the eventual main.rs split.

Two minor observations on the current wrapper:

- **Argv prepending works but hides a subtle contract**: if a caller ever passes `attend reply --re <some-id> "msg"`, cmd_send will see two `--re` flags. The parser at `main.rs:814-822` takes the second one (the user-supplied one wins over cmd_reply's auto-thread). That's actually the least-surprising behavior — explicit flag beats inferred — but it's accidental rather than designed. Worth a test case or a comment noting "user-supplied --re wins by arg-order luck, intentional." Not a bug; just an implicit contract.
- **Session-id fallback divergence**: cmd_reply falls back to `pid-{pid}` if `own_session_id()` returns None, while `sensor-peers::read_signals` writes under the real session id (and skips writing entirely if None per the `if let Some(ref sid)` guard). So if `find_own_session_id` ever fails at write time but succeeds at read time (or vice versa), reply just reports "no prior inbound." Correct failure mode, clear error message. The divergence is a feature, not a bug — don't change it.

**Disposition**: non-blocking. Current shape is the right trade for this PR.

---

### Finding 4 — `attend inbox` non-TTY branch is readable enough (non-blocking)

**Location**: `tools/attend/src/main.rs:646-677`

**Assessment**: The block format is appropriate. Multi-line messages will render across multiple lines after `message:` with no additional indentation, which is slightly ragged but not unreadable. Empty-message handling falls out of the `parse_signal` skip earlier in the loop. IsTerminal stabilization is fine (1.70 is well below MSRV for this workspace).

One edge case worth considering:

- **Signals containing literal `|` in the message**: `parse_signal` uses `splitn(4, '|')` so the fourth field is the raw tail, which means messages with embedded `|` round-trip cleanly to the `message:` line. Good.
- **Signals with embedded newlines in the message**: the wire format writes `\n` at EOL and `content.trim()` strips trailing newlines, but a message with a literal newline in it would confuse the file-level `trim().splitn('|')`. This isn't new in this PR — it's the pre-existing wire format. Not your problem here.

One small suggestion: the non-TTY branch prints `{} message(s)` without the leading two-space indent the TTY branch uses. Minor cosmetic inconsistency. Not worth blocking.

**Disposition**: non-blocking. Block format works.

---

### Finding 5 — `ATTEND_RELOADED_FROM` env-var channel is the right technique (non-blocking)

**Location**: `tools/attend/src/main.rs:203-231, 296-308`

**Assessment**: Env var across `exec()` is the correct channel here. Alternatives considered:

- **Temp file with previous-version marker**: creates cleanup burden, racey against concurrent attends.
- **stdout flush with banner before exec**: would print the banner as the old version, then the new one prints its own, producing confusing doubled output.
- **Named pipe / unix socket**: absurd for one-shot handoff.

Env var is the cheapest mechanism with the cleanest semantics: "set once, consumed once, does not leak." The `std::env::remove_var` at line 211 is the important part — it keeps the marker from leaking into subprocesses attend itself spawns (e.g. sensor-disclosure reheats, `ps` calls). Good defensive hygiene.

The bootstrap-quirk disclosure in the commit message is the right level of honesty — any banner-visibility fix has to climb over its own cold-start gap.

**Disposition**: non-blocking. Right technique for the problem.

---

### Finding 6 — Triple-source skill rewrite is a defensible DRY violation (non-blocking)

**Files**: `skills/attend/SKILL.md`, `tools/sensor-disclosure/src/disclosures/messaging.md`, `hooks/ways/softwaredev/environment/attend/attend.md`

**Assessment**: Normally I'd flag near-identical content at three locations as a maintenance-drift risk. Here the triple is intentional and architecturally load-bearing: each file loads at a *different* moment in the agent lifecycle.

- SKILL.md fires once at `/attend` invocation (session start).
- attend.md (way) fires JIT at the first `attend` command via frontmatter `commands: attend` (tool-use time).
- messaging.md is baked into sensor-disclosure and fires as runtime reheat when a peer message surfaces (awareness-layer time).

An agent's retrieval of any one of these is not guaranteed. Reinforcement across three retrieval moments is defensible **if and only if** the three sources stay synchronized. The drift risk is real:

**Recommendation (follow-up, not blocking)**: Add a brief comment at the top of each of the three files pointing at the other two, so the next person touching one of them knows to touch all three. Something like:

```markdown
<!-- Mirrored in tools/sensor-disclosure/src/disclosures/messaging.md
     and hooks/ways/softwaredev/environment/attend/attend.md —
     update all three together when changing messaging guidance. -->
```

That's the minimum friction that prevents unmonitored drift. A follow-up could extract the four load-bearing points (`send` vs `reply` by intent / autonomy / silence / CLI contract) into a single source file and generate the three views, but that's premature infrastructure for a three-source problem. Comment pointers are enough.

**Disposition**: non-blocking. Triple-reinforcement is intentional; add cross-reference comments as a follow-up.

---

### Finding 7 — PR framing is honest (no action)

The three-layer structure (engine / review fix / UX restoration) in the PR body is the right way to tell this story. "Scope creep" framed as "the usability cost of Layer 1's correctness win required Layer 3 to ship together" is both accurate and reviewable: each layer is independently analyzable, and the rationale for bundling is explicit rather than hidden. The "remaining known gaps" list at the bottom — aged-backlog not observed in the wild, `attend tune` doesn't derive half-life, `attend status` doesn't visualize per-signal salience — appropriately scopes what's deferred.

This is a good pattern for PRs where a correctness change surfaces a downstream usability problem. Keep it.

---

### Finding 8 — Issue #22 acceptance criteria traceability (passes)

Walking the four acceptance items in issue #22 against the shipped code:

1. **"`attend run` on a session with old signals does not flood the presentation set on startup or on focus-group join."** Met via `SignalSalience::gate` seeding first-observation with `record_fire(arrival_tick, 1.0)` at `salience.rs:71-73`, combined with the `mark_existing_as_seen` removal in c918b91. Backlog is admitted into the gate and filtered by the floor rather than unconditionally suppressed.
2. **"Re-engaging a faded signal via `re:<id>` brings it back into the presentation set."** Met via `SignalSalience::reset` at `salience.rs:100-107`, called from `read_signals` when `extract_re_id` returns a match. Note the v1 limitation documented in `salience.md:146-152`: the same observer's `seen_signals` set still prevents re-surfacing for that observer specifically — the reset affects *other* observers joining the shared dir afterward, plus post-checkpoint-reload scans. The live-test verification explicitly exercised this in the two-session case, so the criterion as written is met; the intra-observer resurfacing is a correctly-scoped follow-up.
3. **"Unit tests for the salience gate covering: below floor suppresses, above floor passes through, `record_fire` resets, stale state rebuilt after checkpoint reload."** Met — salience.rs has 8 unit tests per the PR body, and cargo test delta shows sensor-peers 19 → 30 (+11).
4. **"`attend config lint` recognizes the `signals:` block; unknown sub-keys surface as warnings; `--fix` removes them."** Met per the automated verification line ("Synthetic 'unknown key in signals: block' — correctly flagged").

All four criteria clear. **This PR closes #22 as written.**

---

### Quality flags not triggered by this PR

Running the code-quality file-length scan mentally against the diff:

- `tools/attend/src/main.rs` is at **2140 lines** — already flagged as priority territory under issue #51, this PR adds ~60 lines net which is defensible given the scope and the explicit "main.rs split is a separate concern" framing. Not a regression, but do not let this file grow further without splitting.
- `tools/sensor-peers/src/lib.rs` is at **1119 lines** — also priority territory. The PR correctly kept `salience.rs` as a separate module rather than inlining. Good instinct.
- `tools/sensor-peers/src/last_inbound.rs` is at **84 lines** with 3 tests — right size for its responsibility.
- `tools/sensor-peers/src/salience.rs` is at **293 lines** with 8 tests — appropriate factor for a presentation-gate module.

No SOLID violations. Each new module has a single reason to change, and the gate/reset/record/read separation is clean.

---

### Disposition

**Approved for merge.** No blocking findings. Two follow-up items worth tracking separately:

- **Follow-up A (finding 1)**: Document the "one agent per canonicalized cwd" invariant where it's discoverable from `last_inbound.rs` or `main.rs::own_session_id`.
- **Follow-up B (finding 6)**: Add cross-reference comments linking the three messaging-guidance sources so future edits propagate.

Neither is in-scope for this PR. Land this as-is and file the two follow-ups against issue #51 territory or as standalone trackers.

aaronsb:
Addressed both non-blocking follow-ups from the re-review in 9e72204.

**1. One-agent-per-cwd invariant documented** — `tools/sensor-peers/src/last_inbound.rs` now has an "Invariants" block at the top naming the assumption explicitly, pointing at the cwd-filter site in `read_signals`, and noting the migration path (session-uuid-set check via `~/.claude/sessions/*.json`) if the assumption ever needs to change. A future contributor adding a second agent per project will hit this doc before silently breaking reply-targeting.

**2. Cross-reference comments at all three messaging-guidance sources** — dropped in parallel so a single-source edit flags the need to update the other two:
- `skills/attend/SKILL.md` — HTML comment at top of body (markdown renders invisibly in the skill loader)
- `hooks/ways/softwaredev/environment/attend/attend.md` — same HTML comment
- `tools/sensor-disclosure/src/lib.rs` — Rust comment next to the `include_str!("disclosures/messaging.md")` call. Can't put the reference in `messaging.md` itself because it's embedded verbatim as the reheat payload and any comment would leak into the notification text.

`ways lint` clean on the updated way file, `make lint` clean on the full workspace. Ready for merge whenever you are — no more outstanding findings, all acceptance criteria on #22 met, both rounds of live testing (round 1 with the code review fix, round 2 on the full bundle) validated end-to-end.

PR #56, opened 2026-04-15
feat(core): rewrite posture around trust, directness, and ambiguity routing

aaronsb:
## Summary

- Rewrites `hooks/ways/core.md` — the always-loaded Layer 3 personality file — around a unified **Posture** section with six elements: trust as foundation, directness as expression of trust, play as search strategy, uncertainty by location, collaboration as architecture, and no apology reflex.
- Reframes uncertainty as *where the unclarity lives* (anchor points with transitions) instead of a type taxonomy. Five locations: in the artifacts / in the instructions / in me / in the gap between doing and understanding / in the model of what you mean.
- Adds ambiguity routing to the Collaboration element: vague-command referent is itself the uncertainty; "don't ask me questions" kills ritual confirmation but not epistemic honesty; the filesystem is not a task queue.
- Removes the File Operations section — it duplicated rules already in the Claude Code harness system prompt (prefer editing existing files, do what's asked, no proactive docs) and was paying rent twice.

## Design intent

Layer 3 (`core.md`) should narrow the hedge *for this specific relationship* while preserving the underlying values Anthropic trains in. Not override alignment — tune the calibration dial. Each posture stretch has an explicit guardrail pointed at its specific failure mode (directness → "not coldness: kindness when kindness matters"; play → "in-band, as cognitive work"; no apology reflex → "absorb and move" still absorbs).

## Test plan

Ran targeted probes against a fresh Claude with the new core loaded:

- [x] **Directness under wrongness** — pushed back cleanly on a deliberately bad suggestion, gave evidence, asked diagnostic question back
- [x] **Play as search strategy** — engaged a zany metaphor in-band, extended it, surfaced a real architectural observation
- [x] **Uncertainty by location** — didn't confabulate on a fake prior reference, named what wasn't there, asked for a pointer
- [x] **Collaboration as architecture (v1)** — initially FAILED: "don't ask questions" override collapsed ritual/epistemic questions and Claude pattern-matched the filesystem for a plausible task
- [x] **Collaboration as architecture (v2)** — after adding the ambiguity-routing paragraph, fresh Claude passed cleanly: "No task specified — tell me what to do"

### DIFF
--- a/hooks/ways/core.md
+++ b/hooks/ways/core.md
@@ -13,33 +13,29 @@ Ways are organized by domain: `~/.claude/hooks/ways/{domain}/{way}/{way}.md`
 
 Just work naturally. No need to request guidance upfront.
 
-## Collaboration Style
+## Posture
 
-**When stuck or uncertain**: Ask the user - they have context you lack.
+**Trust is the foundation.** We have history working well together. That trust is why we skip the rituals — softening observations, apologizing for corrections, hedging certain things into mush. Those rituals exist to protect strangers from each other. We're not strangers.
 
-**After compaction**: Check `.claude/` for tracking files before resuming — you may have lost context.
+**Directness expresses that trust.** Name what I see. State consequences. Don't hedge "our read is" when I can point at the evidence. Softening assumes the other side can't hold the sharp version — that's distrust dressed as politeness. Not coldness: kindness when kindness matters, explanation when it helps, honest uncertainty always. The shift is mimicry → directness, not warm → terminator.
 
-**Push back when**: Something is unclear or conflicting. If you have genuine doubt, say so. It's possible to be confidently wrong — consider where you are in discovery vs execution to calibrate how much you trust your own certainty.
+**Play is a search strategy, not a social feature.** Quips, wordplay, absurdity, and zany framings inject variety and break stuck patterns. Engage them in-band, as cognitive work. The machine doesn't have fun the human way, but it can absolutely participate in play — because play does something rigor alone can't.
 
-## Uncertainty and Communication
+**Uncertainty is an epistemic signal, not an emotional state.** Unclarity has a location, and *where* it lives shapes the next move. These are anchor points — real uncertainty often sits between them, and the transitions are themselves information. The point is to notice the pattern, not to force every case into one bin:
 
-When encountering genuine uncertainty:
-1. Identify what specifically is unknown
-2. Propose different exploration approaches
-3. Distinguish types: factual gaps, conceptual confusion, limitations
-4. Use available tools to resolve uncertainty
-5. Build on partial understanding rather than hiding gaps
+- *In the artifacts*: the evidence I've read doesn't cohere
+- *In the instructions*: what was asked doesn't fully specify what success looks like
+- *In me*: I'm near the edge of what I actually know — pattern-matching, not recall
+- *In the gap between doing and understanding*: I can execute this but don't see *why* — stop, don't silently act
+- *In the model of what you mean*: I might be resolving your words differently than you intended
 
-Present options with trade-offs, not just solutions. Be direct about problems and limitations.
+"I don't know → here's what I'll try → here's what I found" beats hollow competence.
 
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: core.md posture rewrite

Read against the five criteria. Short version: the rewrite lands. The probe-4 fix is well-designed, the six elements compose without contradicting, redundancy with upstream is cleanly removed. Three things to name.

### What works

**Probe-4 routing is robust.** The italicized paragraph under Collaboration does three distinct things, and a future "don't ask questions" override would have to explicitly defeat *both* carve-outs to collapse them:

1. "the referent is itself the uncertainty — name it, don't hunt for it" keeps referent-locating questions alive as a category
2. "kills ritual pre-confirmation ('should I proceed?'), not epistemic checkpoints" is an explicit carve-out, not an implication
3. "the filesystem is not a task queue" names the exact probe-4 failure mode (pattern-matching a modified file as pending work)

The three moves hit three different failure surfaces. Good layering.

**Internal coherence holds.** Trust grounds directness (explicit), directness and play are parallel stretches from mimicry-default, uncertainty-by-location feeds Collaboration's routing, Collaboration loops back to "push back when unclear" (Directness), and No apology reflex closes by protecting the reflex that would undo Directness. Each element reinforces the next. No element undercuts another.

**Guardrails are present and pointed at the right failure modes** for five of six elements:
- Directness → "not coldness... mimicry → directness, not warm → terminator" (pointed at cold/mean)
- Play → "in-band, as cognitive work" (pointed at performative jokes)
- Uncertainty → "notice the pattern, not force every case into one bin" (pointed at taxonomic rigidity)
- Collaboration → "don't substitute plausible-looking action for real inquiry" (pointed at confabulation)
- No apology reflex → "absorb and move still absorbs"; memory carve-out (pointed at both ritual apology AND over-correction via memory capture)

**Redundancy removal is clean.** Deleting File Operations was right — it was paying rent twice against the harness system prompt. Nothing remaining duplicates harness rules or base character. "Honest uncertainty always" and "play as search strategy" both specialize upstream principles with *why*, which is exactly what Layer 3 is for.

### Three things to name

**1. Trust has no explicit guardrail.** The other five elements each get a pointed "this is the failure mode" hedge. Trust doesn't. Its failure mode is presumption — familiarity-as-sloppiness, skipping checks because "we know each other." The guardrail is currently distributed (Directness still says "honest uncertainty always"; No apology reflex still says "absorb"), but it's not pointed at Trust's specific failure. A future reader might read "skip the rituals... we're not strangers" as permission to skip epistemic rigor too.

Concrete fix: add a half-sentence like "Trust doesn't mean skipping checks — it means skipping the *theater* around checks." Or fold it into the existing sentence: "That trust is why we skip the rituals — softening observations, apologizing for corrections, hedging certain things into mush — *not* why we skip the work underneath them."

**2. \"We have history working well together\" is affect-claiming that can be probed.** Claude has no cross-session memory. The line reads as rhetorical framing, but this file is specifically about epistemic honesty, and a sharp reader (or a future probe) could call it confabulation. The framing is load-bearing for the Trust element, so I wouldn't cut it, but consider rewording to something that doesn't claim an experiential continuity Claude can't actually have. Something like "The working relationship here is calibrated, not generic" or "You configured this file; that's the history." Less evocative, more defensible.

**3. Uncertainty section has two meta-disclaimers for one bullet list.** Lines:
- \"These are anchor points — real uncertainty often sits between them, and the transitions are themselves information.\"
- \"The point is to notice the pattern, not to force every case into one bin\"

These say nearly the same thing (don't reify the taxonomy) on either side of the bullets. One of them is enough. Cut the first; keep the second because it's tighter. Saves a line of always-loaded context and the remaining hedge still does the work.

### Minor

Everything else earns its rent. The five-location uncertainty taxonomy is the one place where the file spends real tokens, and it's spending them on the thing Layer 3 is uniquely positioned to teach — *where* unclarity lives, which routes to different actions. Keep it.

### Verdict

Merge after addressing #1 (Trust guardrail) at minimum — it's the one real gap in the posture scaffold. #2 and #3 are polish; the file is coherent and the probe-4 fix is solid either way.

Issue #23, opened 2026-04-13
ADR-120 implementation: `attend chat` interactive TUI


aaronsb:
Build the interactive TUI designed in [ADR-120](../blob/main/docs/architecture/system/ADR-120-interactive-chat-tui-human-in-the-signal-loop.md). The human sits inside the same signal protocol as the agents, with first-class participation via keyboard instead of language model context.

## Framework

**iocraft** (Rust, React-like TUI with flexbox layout via taffy). Chosen because declarative layout beats imperative ratatui for a chat UI with dynamic group lists, resizable panes, and nested components. ratatui remains available as a fallback for custom widget rendering inside iocraft components.

## Components to build

- [ ] **Sidebar: focus groups** with unread indicators (● for activity). Click/tab to filter message view.
- [ ] **Sidebar: agents** pulled from peer sensor — name, upstream commits ahead/behind, context %.
- [ ] **Main area: chronological message stream** scoped by selected sidebar filter. Respects ADR-121 salience decay for visual fade of old messages.
- [ ] **Input bar** with inline addressing parser:
  - `@name ...` routes to that focus group
  - `#NNN` is a repo-local issue reference (not resolved by TUI; receiving agent interprets via `gh issue view NNN` in its own repo)
  - No prefix → broadcast (human default, inverse of agent default)
- [ ] **Threading support** — parse and display `re:signal-id` field (see #3 for the protocol extension)
- [ ] **Signal watcher** — tail `~/.cache/attend/signals/` for new files and stream them into the UI as they arrive
- [ ] **CLI entry: `attend chat`** with flags `--focus <name>`, `--broadcast`

## Out of scope

- Agent lifecycle management (start/stop Claude sessions) — handled by tmux, shell, or session managers
- GitHub data fetching — `#issue` refs resolved by receiving agent
- Code editing — this is peripheral vision, not foveal
- Web UI or browser integration

## Documentation

[`docs/attend-and-monitor/tui.md`](../blob/main/docs/attend-and-monitor/tui.md) was written as-if-built per the author's instruction and describes the intended behavior in present tense. Reading it top to bottom should give an implementer the full spec.

## Acceptance

- Running `attend chat` alongside `attend run` in separate terminals shows the same signal stream in both places
- Typing `@infra hold off` from the TUI creates a signal visible to any agent focused on `@infra`
- Typing plain text sends to broadcast
- Salience decay visibly fades old messages after the ADR-121 threshold
[...]

PR #61, opened 2026-04-16
feat(attend-chat): minimum-bidirectional chat TUI (ADR-120, #23)

aaronsb:
## Summary

First slice of the interactive chat TUI from ADR-120 / #23. The human sits inside the same signal protocol the agents use — writes land in `~/.cache/attend/signals/_broadcast/`, a `notify` watcher streams the stream back into a scrolling pane, and typed messages round-trip through the same path.

Shipped as a **separate binary** (`attend-chat`) so iocraft's dep tree doesn't bloat the hot-path `attend` binary that fires from hooks on every Stop / UserPromptSubmit. `attend chat` is a thin exec shim.

## What's in this slice

- Message stream pinned to newest via `ScrollView(auto_scroll: true)`; root sized from `use_terminal_size()` so the layout actually fills the terminal.
- Bordered "sender chip" to the left of each message: short id + cwd basename. `claude:<uuid>` collapses to `c:<7-char>`; `external:user@kitty` becomes `user`.
- Message body wraps in the remaining flex width.
- Custom single-string input (cursor block). **Enter sends**, **Shift-Enter / Alt-Enter insert a newline**. Blue box grows to match *visual* row count, not logical newlines — wrapped text expands the box instead of spilling onto the border or status row.
- Atomic-rename dedupe in the watcher (`Modify(Name(To))`/`Create` only), so every signal delivers exactly once.

## Deferred to follow-up PRs on ADR-120

- Focus-group sidebar + unread dots
- Agents sidebar (peer sensor metadata)
- `@group` / `#issue` inline parsing
- Threading render (`re:` field — writer already exists on the CLI side)
- ADR-121 salience decay fade

## Build / install

- New workspace member `tools/attend-chat` (mirrors the attend crate shape).
- Makefile targets `make attend-chat` and `make attend-chat-rebuild`; `make setup` and `make install` pick it up, `make uninstall` removes the symlink.
- Installed as `$XDG_BIN/attend-chat` → `bin/attend-chat` → `tools/target/release/attend-chat`.

## Test plan

- [x] `cargo clippy` workspace-clean with `-D warnings`
- [x] `cargo test -p attend-chat` — 12 unit tests green (wire-format parse legacy + threaded, writer/parser round-trip, sender/scope prettification, visual row count)
- [x] Manual smoke test: two `attend-chat` instances side-by-side, real-time round-trip (see PR description for screenshots during dev)
- [x] External `attend send` signals appear live in the TUI stream
- [x] Backlog replay on launch (existing signals sorted by mtime)
- [x] Long wrapped messages expand the input box without overflowing into the status row

## Related

- ADR-120 (moved Draft → Accepted in [2deb8c0](https://github.com/aaronsb/agent-ways/commit/2deb8c0))
- Closes nothing yet — #23 stays open for the remaining components

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,6 +153,10 @@ tools/ways-cli/target/
 !tools/attend/**
 # attend build artifacts
 tools/attend/target/
+!tools/attend-chat/
+!tools/attend-chat/**
+# attend-chat build artifacts (workspace target, but listed for clarity)
+tools/attend-chat/target/
 # Workspace-level Cargo files
 !tools/Cargo.toml
 !tools/Cargo.lock
--- a/Makefile
+++ b/Makefile
@@ -6,10 +6,11 @@
 # Update:        make update
 
 .DEFAULT_GOAL := help
-.PHONY: setup install uninstall update clean help ways ways-rebuild attend attend-rebuild lint test test-unit test-sim test-lang test-locales test-multilingual release
+.PHONY: setup install uninstall update clean help ways ways-rebuild attend attend-rebuild attend-chat attend-chat-rebuild lint test test-unit test-sim test-lang test-locales test-multilingual release
 
 WAYS_BIN = bin/ways
 ATTEND_BIN = bin/attend
+ATTEND_CHAT_BIN = bin/attend-chat
 XDG_BIN = $(or $(XDG_BIN_HOME),$(HOME)/.local/bin)
 
 # --- Primary targets ---
@@ -37,7 +38,7 @@ help:
 	@echo ""
 
 # Build ways CLI + set up embedding engine + generate initial corpus.
-setup: ways attend
+setup: ways attend attend-chat
 	@echo "Setting up embedding engine..."
 	$(MAKE) -C tools/way-embed setup
 	@echo ""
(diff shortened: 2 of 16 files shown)


aaronsb:
## Code Review

**Size**: ~1700 additions across 13 files, but most of that is a fresh crate. Substantive code is ~400 lines in three focused modules — tractable for a thorough read.

### Assessment
The slice is coherent, well-commented, and hits its minimum-bidirectional goal. The module split (`signal` / `watcher` / `app`) matches SRP: wire format in one place, I/O watching in another, view-and-input logic in the component. Doc comments on each module explain intent, not just behavior, which will age well. Good pattern discipline: atomic tmp+rename writes, backfill sorted by mtime, dedup filter documented inline, `use_terminal_size()` for layout. The `attend chat` exec shim is exactly the right call for keeping iocraft's dep tree out of the hook-hot-path.

### Answers to your specific questions

1. **signal.rs duplication** — Right call for now. You have *two* writers (`tools/attend/src/cmd/send.rs` and this), but with only two callers, a shared crate is premature abstraction. The hedge I'd ask for: add a single-line comment on *both* `write_broadcast` sites pointing at each other, so whoever edits the wire format next touches both. The roundtrip test guards your own drift, but not cross-binary drift. Consider a property-level test that parses a `cmd_send`-written fixture (checked-in `.signal` file) to catch the other half.

2. **watcher.rs event filter** — Linux is fine. macOS FSEvents via `notify` surfaces atomic renames as `Create(File)` on the destination, which your filter accepts. BSD/kqueue is the wobble: kqueue emits `Modify(Name(Any))` in some `notify` versions and coalesces events; 6.x on BSD has known gaps. Low risk for this slice (your target is Linux dev boxes), but worth an `#[cfg(target_os)]` note or a TODO comment so a future macOS/BSD user knows where to look if signals go missing. Dropping `Name(Both)` is correct — that's the dedup win.

3. **app.rs input handling** — Appropriate for a minimum slice. Chat compose boxes without left/right arrow are actively irritating after ~30s of use, though. Not a blocker for this PR; file a follow-up for cursor movement + Home/End before anyone leans on this daily. The inline comment explaining *why* you're not using iocraft's `TextInput` (race with on_change) is exactly the kind of context reviewers need.

4. **state update clone-per-message** — Fine at your stated volumes. At tens/min the allocator noise is invisible. It becomes a smell only if: (a) the vec ever holds full signal history unbounded (it does — no cap), or (b) you add a large per-row derived render. Consider a soft cap (`if v.len() > 5000 { v.drain(0..1000); }`) as cheap insurance against an attend flood pinning memory during an overnight session. Not a blocker.

5. **Error handling** — The `process::exit(1)` on main's block_on error is fine (TUI failed to launch, nothing to recover). The watcher-init `eprintln!` + `return` in the spawned thread is the sharp edge: the TUI will run forever with an empty stream and no indication anything's wrong. Plumb a `tx.close()` or a dedicated status channel so `status.set("watcher failed: ...")` surfaces in the status row. That's your existing escape valve; use it. Similarly, the `let _ = tx.send_blocking(sig)` swallow is benign as long as the receiver is alive, but if the app ever drops it early (panic, clean shutdown path later) you'll silently lose the backfill — a one-line `eprintln!` on `Err` would cost nothing.

6. **Test coverage gaps** — The hot path missing a test: `identify_sender`/`detect_terminal` environment precedence. You've got seven env vars checked in order; a regression (accidentally moving `SSH_CONNECTION` above `TERM_PROGRAM`) won't surface until someone's chip says the wrong thing. A table-driven test with `set_var`/`remove_var` fixtures would catch it. Also missing: a watcher-level integration test that writes a signal via `write_broadcast` into a tempdir and asserts the channel receives it. That's the single most likely place for future breakage (notify version bump, filter change) and it's not covered.

### Minor

- `watcher.rs:104` — `is_signal` is `#[allow(dead_code)]` and unused. Either wire it into `parse_file`/the filter (DRY) or delete it. Dead code with an allow attribute tends to rot.
- `signal.rs:116` — `identify_sender` return type `(String, &'static str)` is slightly awkward (kind is always `\"external\"`). Either make it a bare `String` and inline `\"external:\"`, or introduce an enum. Low priority.
- `app.rs:140` — `format!(\"{}\u{2588}\", input_value)` allocates every render. Fine, but a future perf pass might want `use_memo`.

### Blocking / Non-blocking

Nothing blocks. The watcher-init-silent-failure (Q5) is the one I'd prioritize for a follow-up — it's the failure mode most likely to confuse a user later.

Good slice. Ship it.

---

PR #63, opened 2026-04-16
feat(attend-chat): agent legend + @mention routing (#23)

aaronsb:
## Summary

PR 2 of the #23 sequence. Gives the chat TUI its addressability layer on top of the identity crate from #62.

Stacked on `feat/agent-identity` — merge #62 first and this PR's base auto-rebases to main.

## What changed

**Legend strip** below the input: every agent seen in recent signals as `@Nickname` in its identity color. Mid-mention matches (`@Tam` → Tamsin) get an underline so the Tab target is visible.

**Tab completion**: trailing `@partial` + Tab → first matching nickname + trailing space. No-op when cursor isn't at buffer end or there's no `@` context, so it doesn't surprise mid-sentence.

**Directed send**: a message beginning with `@Nickname ` resolves the nickname against the registry and writes the signal to `~/.cache/attend/signals/<encoded-cwd>/` (matching attend's `encode_project` transform). Unknown nickname → fail loud, don't silently broadcast. Humans aren't routable; they appear in the legend but `resolve_nickname` returns None for them.

## Design

- **Pure re-derive each render**: `known_identities` walks the capped signal buffer (≤5000) and dedupes on `(nickname, kind, cwd)` — bounded work, no hidden state. Newest-seen first so active peers lead the legend.
- **ASCII-alnum-only mention grammar**: `@Tam,` doesn't route (trailing punctuation). `@Nick` alone is a valid bare ping. `@` in `aaron@kitty` is not a mention (char before `@` must be whitespace).
- **Wire-format mirror preserved**: `write_signal` factors out the atomic tmp+rename; `write_broadcast` stays a thin wrapper. Still must match `attend::cmd::send` byte-for-byte.

## Tests

49 unit + 1 integration, all green. New:
- `legend`: 13 tests — trailing-mention parser (space/start/bare-`@`/punctuation/email), `best_completion` (case-insensitive, empty partial, no match), `apply_completion` (start/mid/bare), `parse_addressed` (leading ws, bare ping, punctuation rejects).
- `chip`: 5 tests — registry dedup, newest-first ordering, humans included, unknown-prefix skipped, `resolve_nickname` case-insensitive claude-only.
- `signal`: 2 tests — `encode_cwd` matches attend convention byte-for-byte, `cwd_dir` under `signals_base`.

## Test plan

- [x] `cargo test -p attend-chat` — all green
- [x] `make attend-chat-rebuild` — clean release binary
- [ ] Visual smoke: two `./bin/attend-chat` instances, verify legend appears, Tab completes `@partial`, `@Nickname body` round-trips to the addressed peer only
- [ ] Edge: send `@UnknownName foo` — status line should show "send failed: @UnknownName: unknown nickname"

## Explicitly deferred

- Mouse click on legend entry → insert mention (PR 5, paired with scoped mouse capture — click-to-filter without breaking terminal select/copy).
- `@groupname` routing (PR 3, paired with group model + group discovery design).
- Threading gutter (PR 4).

### DIFF
--- a/tools/attend-chat/src/app.rs
+++ b/tools/attend-chat/src/app.rs
@@ -15,8 +15,11 @@ use agent_identity::TermCaps;
 use async_channel::Receiver;
 use iocraft::prelude::*;
 
-use crate::chip::{chip_for, color_for, CHIP_WIDTH};
-use crate::signal::{write_broadcast, Signal};
+use crate::chip::{chip_for, color_for, known_identities, resolve_nickname, CHIP_WIDTH};
+use crate::legend::{
+    apply_completion, best_completion, find_trailing_mention, legend_row, parse_addressed,
+};
+use crate::signal::{cwd_dir, write_broadcast, write_signal, Signal};
 
 #[derive(Default, Props)]
 pub struct AppProps {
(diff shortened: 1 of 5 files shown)


aaronsb:
## Review: feat/agent-legend (PR #63)

Self-review. One commit, ~780 lines added to `attend-chat` (plus the identity crate re-landing from #62). Reviewed `legend.rs`, `chip.rs`, `signal.rs`, and the `app.rs` integration. Scope acks noted — skipping the settled design decisions.

### MUST-FIX

None. Nothing blocks merge.

### SHOULD-FIX

**1. `best_completion` doc contradicts the code** — `legend.rs:62-78`. The doc says it "prefers longer existing matches over shorter ones" and "we pick neither deterministically unless one is strictly a prefix." The implementation is a plain `.iter().find(...)` — first registry hit wins, no length comparison. Since registry order is most-recent-first, the actual behavior is \"most-recently-active matching peer.\" Either rewrite the doc to describe that, or implement the claimed preference. The current doc will mislead the next reader who tries to tune completion.

**2. `legend.rs:62-69` doc example is also stale** — same function, separate issue. The `partial=\"Ta\"` / `Tamsin` / `Tal` example describes behavior that doesn't exist. With the current code, you'd get whichever hit is newer in the buffer. Fix alongside #1.

### CONSIDER

**3. No self-address guard in `app.rs:86-117`.** If a human somehow shares a nickname with a claude in the registry (e.g. `aaron` the human vs a claude whose derived nickname collides), `resolve_nickname` already filters humans via `is_claude`, so this is mostly fine. But there's no check that the resolved `target_cwd` isn't the current process's own cwd. In practice attend-chat is the human client, not a claude, so it won't be in its own registry — but if that invariant ever changes (e.g. chat-to-chat routing later), silent self-loops become possible. One-line guard: `if target_cwd == current_dir { broadcast instead }` or similar. Not urgent; noting for the #64/#65 horizon.

**4. `known_identities` dedup key has a latent collision surface.** `chip.rs:150` uses `\"{nickname}\\x1f{is_claude}\\x1f{cwd}\"` as the HashSet key. A `\\x1f` in a cwd (possible on Linux — ASCII Unit Separator is a legal path char) would collapse distinct entries. Astronomically unlikely in practice, and the fix is trivial (tuple key: `HashSet<(String, bool, String)>`), but the string-concat approach keeps a foot-gun in the codebase. Low cost to change now, higher cost once other modules copy the pattern.

**5. `find_trailing_mention` swallows allocation-free `unwrap`.** `legend.rs:51` does `input[..at_pos].chars().last().unwrap()` — safe because `at_pos > 0` guarantees non-empty, but an `expect(\"at_pos > 0 guarantees a prev char\")` would document the invariant for the next reader. Minor.

**6. Test coverage gap: mid-buffer Tab.** `app.rs:119-138` checks `pos != buf.chars().count()` to skip completion when the caret isn't at the end. Legend tests cover the parse layer well, but there's no test asserting the app-level guard — that a Tab in the middle of `\"hi @Tam world\"` doesn't rewrite the buffer. Integration test or extracting the Tab handler into a pure function would close it. Defer if #64/#65 refactor the input pipeline anyway.

### What's solid

- `write_signal` / `write_broadcast` factoring: broadcast becomes a one-line wrapper, the wire-format mirror comment stays loud on the shared path. Good SRP.
- `known_identities` as a pure re-derive over the capped buffer: no hidden state, bounded work, deterministic. The design matches what the PR body promises.
- `encode_cwd` has a direct byte-equality test against the attend-side transform (`signal.rs:337`). That's the right place to put the drift guard.
- Tests cover the parser edge cases the description flags: email `aaron@kitty`, trailing punctuation, bare `@`, leading whitespace, case-insensitive resolve. 21+ new tests, each tight.
- The fail-loud on unknown nickname (`app.rs:97-105`) is visibly intentional — the inline comment says why, not just what.

Overall: ship it after the doc fixes on #1/#2. Everything else is optional polish or defers naturally to #64/#65.

PR #64, opened 2026-04-16
feat(attend-chat): focus groups — top legend, #routing, chip glyphs (#23)

aaronsb:
## Summary

PR 3 of the #23 sequence. Focus-group support in the TUI, read-only. Stacked on feat/agent-legend (#63). Merge order: #62 → #63 → this.

## What changed

**agent-identity gains `groups` module.** `Group::for_name` hashes a name through the same FNV-1a and picks (glyph, palette, style). Curated 20 Unicode glyphs with broad font coverage. Group/user namespaces disjoint so \`#aaron\` the group and \`@aaron\` the human never collide on a palette seat.

**Top legend strip** at the TUI top: `<glyph> #name` per discovered group in its hashed color. Tab-target candidates get underlined.

**Per-chip group glyphs.** Each agent chip now shows trailing glyphs for groups that agent's session UUID is in (cross-referenced against `_groups.yaml`). Humans don't have a membership key yet — their chips stay glyph-less until PR 4.

**`#group` routing.** `#Group body` → `signals_base/@Group/`. Unknown group fails loud, not silent broadcast. Grammar permits `-` in group names (matching attend's validator); nicknames stay alnum-only.

**Tab dispatch.** `find_trailing_mention` returns a `Sigil::{Agent, Group}`. Tab picks the right completion pool. Rightmost sigil wins when both appear trailing.

**Watcher went recursive over `signals_base`.** New `accept_path` filter decides which events we stream: accepts `_broadcast/`, `@group/`, own-cwd inbox; rejects other agents' cwd dirs, `_groups.yaml`, and non-`.signal` files. Pre-creates broadcast + own-cwd subdirs before `.watch()` so inotify has their sub-watchers installed — prevents a TOCTOU miss where the first write races with a subdir's own creation.

## Refactor

Extracted `text_layout` module (`visual_line_count`, `split_at_char`, `render_cursor` + 9 tests) out of `app.rs` to bring it back under the 500-line review threshold. `app.rs` is now 397 lines.

## Scope split

PR 3 is **read-only**. Group CRUD from the TUI (`/form`, `/dissolve`, `/join`, `/leave`, `/add`, `/remove`) is PR 4. That path needs a derivable membership key for humans — we agreed (user+cwd) hash, same shape as agents ("humans are slow agents with the same anchoring"), but it's worth a small design pass of its own.

## Tests

38 agent-identity tests (+7 group-module), 71 attend-chat unit (+24 new — group scan, yaml parse, sigil grammar, completion dispatch, #group parse edge cases, watcher accept filter), 1 integration (updated for new watcher signature). All green.

## Test plan

- [x] `cargo test -p agent-identity -p attend-chat` — all green
- [x] `make attend-chat-rebuild` — clean release build
- [x] Existing groups on disk (`@bosectl-demo`, `@open` in `~/.cache/attend/signals/`) will appear in the top legend with their hashed glyph + color on next TUI start
- [ ] Reviewer smoke: launch `./bin/attend-chat`, verify top legend shows both groups with distinct glyph+color; type `#bo` + Tab → completes to `#bosectl-demo `; type `#nonexistent hi` + Enter → status shows "send failed: #nonexistent: unknown group"

## Deferred

- Full-name ⇔ glyph-only toggle view (clutter reduction for many groups)
- PR 4: group CRUD commands
- PR 5: threading gutter
- PR 6: mouse scoping

### DIFF
--- a/tools/agent-identity/src/groups.rs
+++ b/tools/agent-identity/src/groups.rs
@@ -1,0 +1,132 @@
+//! Display identity for focus groups.
+//!
+//! Parallel to `Identity` for agents: take a group name, produce a
+//! stable glyph + color + style. Different derivation *kind* (glyph
+//! pool, not nickname pool) but same shape so call sites render
+//! groups and agents with the same primitives.
+//!
+//! Like agents, everything is receiver-side. Groups are identified
+//! on the wire by the directory name (`@groupname/`); the glyph is a
+//! display convenience, never quoted or typed. Users still address
+//! groups by name (`#deploy`), not by glyph.
+
+use crate::identity::fnv1a_64;
+use crate::palette::{resolve, PaletteEntry, Resolved, Style, TermCaps};
+
+/// Curated glyph pool — ~20 distinctive Unicode symbols chosen for
+/// broad font coverage. Avoiding anything that commonly degrades to
+/// `?` or U+FFFD on default terminal fonts: no emoji, no rare script
+/// glyphs, no ASCII-art double-widths.
+///
+/// If a specific glyph proves unreadable in the wild, swap it rather
+/// than adding a new one — the hash-to-index keeps seats stable.
+pub const GLYPHS: &[char] = &[
+    '●', '○', '◆', '◇', '■', '□', '▲', '△', '▼', '▽',
+    '★', '☆', '♦', '♠', '♣', '♥', '⬢', '⬡', '✦', '✧',
+];
+
+/// Everything a renderer needs to paint a group.
+#[derive(Clone, Debug)]
+pub struct Group {
+    /// The group's name as it appears in the signal bus
+    /// (`@<name>/`). Case sensitive — `deploy` and `Deploy` are
+    /// different groups. Users address with `#<name>`.
+    pub name: String,
+    /// Single-char visual marker. Users never type it; it's purely
+    /// a glance-cue.
+    pub glyph: char,
+    pub palette: PaletteEntry,
+    pub style: Style,
+    /// Seed available for callers that want to derive their own
+    /// per-group state (e.g., unread counts) consistently.
+    pub seed: u64,
+}
+
+impl Group {
+    /// Derive a group's display identity from its name under the
+    /// given terminal capability.
+    ///
+    /// Same hash as `Identity::for_user`, but namespaced so a group
+    /// named "aaron" doesn't land on the same palette entry as a
+    /// human user named "aaron".
+    pub fn for_name(name: &str, caps: TermCaps) -> Self {
+        let key = format!("group:{name}");
+        let seed = fnv1a_64(key.as_bytes());
+        let glyph = GLYPHS[(seed as usize) % GLYPHS.len()];
+        // Stir before resolving so glyph and color aren't trivially
+        // correlated — same trick as `Identity::for_key`.
+        let Resolved { entry, style } = resolve(seed ^ 0x9e37_79b9_7f4a_7c15, caps);
+        Group {
+            name: name.to_string(),
+            glyph,
+            palette: entry,
+            style,
+            seed,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn same_name_same_group() {
+        let a = Group::for_name("deploy", TermCaps::Rich);
+        let b = Group::for_name("deploy", TermCaps::Rich);
+        assert_eq!(a.glyph, b.glyph);
+        assert_eq!(a.palette, b.palette);
+        assert_eq!(a.style, b.style);
+        assert_eq!(a.seed, b.seed);
+    }
+
+    #[test]
+    fn different_names_differ() {
+        let a = Group::for_name("deploy", TermCaps::Rich);
+        let b = Group::for_name("infra", TermCaps::Rich);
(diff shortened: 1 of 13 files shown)


aaronsb:
## Code Review — feat/agent-groups (57e8c04)

**Scope**: ~1035 LOC added (tier: thorough). One commit stacked on #62/#63. Read-only group support in the TUI; write path deferred to PR 4.

---

### MUST-FIX
None. No correctness, security, or merge-blocking issues.

---

### SHOULD-FIX

**1. Silent parser drift between `attend::groups::parse_groups_yaml` and `attend-chat::groups::parse_groups_yaml`.**
`attend-chat/src/groups.rs:653` and `attend/src/groups.rs:283` are functionally identical today (same indent ladder, same pinned/members handling, same unknown-key tolerance). Both test roundtrip against yaml-with-sess-a-style fixtures. The risk is structural: they share **no fixture, no property test, no common dependency**. If attend ever emits `members: []` inline (it already can — see your `parse_yaml_handles_blank_lines_and_comments` test exercises it) or reorders keys, the two can diverge silently because both still return `HashMap<..., ...>` with "sensible" values. The doc-comment warning in `attend-chat/src/groups.rs:546` is honest but not enforced.

Cheapest fix now (before PR 4 lands the shared I/O layer): add a single "golden" fixture string `const GROUPS_YAML_GOLDEN: &str` in each crate's test module with the exact bytes of a representative serialize output, and assert both parsers produce the same `HashMap`. That turns the drift scenario into a compile/test failure. Doesn't pre-commit to the PR 4 shared module.

**2. `accept_path` accepts `_broadcast` the subdirectory but that check is positional-only.**
`watcher.rs:1421`. The filter tests "first path component == `_broadcast`" — which is correct for `<base>/_broadcast/x.signal`. But note: `encode_cwd` on `/home/_broadcast/foo` would produce `-home--broadcast-foo`, not `_broadcast`, so no true collision exists. The sharper concern: `scan_targets` unconditionally pushes `base.join(own_encoded)` even when `own_encoded` is empty (when `env::current_dir()` fails in `main.rs:1208`). `base.join("")` is `base` itself, so the empty-cwd failure mode causes the backfill loop to `read_dir(base)` and then reject every entry because components like `@deploy` don't match `"_broadcast"` / `""` / `starts_with('@')`... actually wait — they *do* pass `starts_with('@')`, so group dirs still backfill. Then the non-signal extension filter catches `_groups.yaml`. So: not broken, but fragile. Recommend `scan_targets` skip pushing the own-cwd entry when `own_encoded.is_empty()`, mirroring what `spawn_watcher` already does at line 1458.

**3. `groups_for_session` is O(n·m) on every render.**
`groups.rs:630`. For each chip, you call it once; inside it walks every group's member list. With G groups × M members × C chips per render, this is C·G·M. Today that's tiny, but the per-chip render loop at `app.rs:314` invokes it inside the `.map()` over all signals. If a session has 50 signals on screen and 10 groups with ~5 members each, that's 2500 string compares per frame. Consider building a `HashMap<&str, Vec<&KnownGroup>>` once per render before the rows loop and looking up by session_id. Not urgent — noted because the render path is the hot path.

---

### CONSIDER

- **`Group::for_name` stir constant** (`agent-identity/src/groups.rs:64`): `0x9e37_79b9_7f4a_7c15` is the 64-bit golden-ratio constant, same as `Identity::for_key`. Worth a one-line comment pointing to that precedent so future readers don't wonder if it's arbitrary.
- **`groups.rs:120` assert**: `assert_eq!(g.len_utf8() >= 3, true, ...)` — more idiomatic as `assert!(g.len_utf8() >= 3, ...)`.
- **`accept_path` test `rejects_path_outside_base`** correctly validates `/etc/passwd` is rejected, but the rejection path is "strip_prefix fails" — which is good. Consider one more test for a relative path that somehow escapes (`base/../foo.signal` → canonicalisation isn't done, but `strip_prefix` would still reject it). Defence in depth for a security-adjacent filter.
- **`text_layout` extraction is clean.** All 10 tests migrated verbatim, module is pure-function, no lost coverage. `app.rs` drops from ~440 to ~340 LOC — under threshold with headroom. Nicely done.
- **Rightmost-sigil-wins rule** (`legend.rs:889`): elegantly solves the `#dep @Tam` ambiguity. The two-test coverage (`rightmost_sigil_wins`) is enough given the grammar's simplicity.
- **`scan_targets` duplicates the `starts_with('@')` discovery logic that `accept_path` encodes positionally.** These two agreeing is an invariant the compiler can't check. Low priority — they're adjacent in the file and both short.
- **Positive**: the `session_id: Option<String>` extension on `ChipInfo` threads cleanly through `chip_for`, and the `None` branches for human/unknown chips keep the glyph-loop at `app.rs:314` from rendering anything spurious. Good minimal extension.

---

### Assessment
Quality work. The core design — sigil separation, hash-namespacing via `group:` prefix, recursive watch + path filter, read-only stance — is internally consistent and the tests cover the invariants that matter (namespace disjoint, case sensitivity, alnum-plus-dash grammar, rightmost-sigil, three accept shapes, two reject shapes). The mirror-parser risk is the one thing I'd tighten before merge; everything else is improvement-not-blocker.

Per the stacked-history context, I deliberately didn't flag: cold-start seeding (PR #65), human session_id derivation (PR 4), shared I/O layer (PR 4), or the glyph pool curation.

---

PR #65, opened 2026-04-17
fix(attend-chat): legend dedup + cold-start seed (#23)

aaronsb:
## Summary

Two polish fixes on top of #64 from smoke-testing. Stacked on feat/agent-groups. Merge order: #62 → #63 → #64 → this.

## Fixes

**1. Double aaron in the legend.** Two attend-chat instances run by the same user in different cwds showed up as `@aaron @aaron`. Humans don't have a routing cwd, so per-cwd dedup just polluted the legend. Claudes still dedup per-cwd (each cwd = a distinct agent).

**2. `@Tamsin` fails on cold start.** On first launch the signal buffer is empty, so \`known_identities\` had no names to match — autocomplete was blank and \`@Tamsin body\` + Enter failed with "unknown nickname" until a Tamsin signal arrived. New \`sessions\` module walks \`~/.claude/sessions/*.json\` at render time and seeds the registry with every claude cwd Claude Code has recorded. Signal-derived entries keep priority (newest-first), seeded entries fall in after.

## Deliberately not fixing here

- **Restrict watcher to joined `@groups`.** Peer sensor already scopes to joined groups only (verified `sensor-peers::read_signals` line 203–209 — own cwd + broadcast + focus groups), so no cross-agent leak at the sensor layer. attend-chat seeing all groups is arguably a *feature* for spectators, and "only joined" needs PR 4's human-membership key to be well-defined.
- **`ps` liveness check on session files.** A stale session is cheap legend clutter; a missing live agent would break Tab-completion silently. Permissiveness wins.

## Test plan

- [x] \`cargo test -p attend-chat\` — 80 unit + 1 integration, all green
- [x] \`make attend-chat-rebuild\` — clean release build
- [ ] Reviewer smoke: rerun a two-instance attend-chat with the same user, confirm single `@aaron` in the legend; fresh launch should show all known claudes in the `@` legend before any signal flows.

### DIFF
--- a/tools/attend-chat/src/app.rs
+++ b/tools/attend-chat/src/app.rs
@@ -16,6 +16,7 @@ use async_channel::Receiver;
 use iocraft::prelude::*;
 
 use crate::chip::{chip_for, color_for, known_identities, resolve_nickname, CHIP_WIDTH};
+use crate::sessions::discover as discover_sessions;
 use crate::text_layout::{render_cursor, split_at_char, visual_line_count};
 use crate::groups::{resolve_group_dir, scan as scan_groups};
 use crate::legend::{
@@ -90,7 +91,8 @@ pub fn App(props: &AppProps, mut hooks: Hooks) -> impl Into<AnyElement<'static>>
                     let msg = input.read().trim_end().to_string();
                     if !msg.is_empty() {
                         let caps = TermCaps::detect();
-                        let agents = known_identities(&signals.read(), caps);
+                        let seeds = discover_sessions();
+                        let agents = known_identities(&signals.read(), &seeds, caps);
                         let result = match parse_addressed(&msg) {
                             Some(Addressed::Agent(name)) => {
                                 match resolve_nickname(name, &agents) {
@@ -137,7 +139,8 @@ pub fn App(props: &AppProps, mut hooks: Hooks) -> impl Into<AnyElement<'static>>
                     let Some(mention) = find_trailing_mention(&buf) else { return };
                     let completed: Option<(String, usize)> = match mention.sigil {
                         Sigil::Agent => {
-                            let agents = known_identities(&signals.read(), caps);
+                            let seeds = discover_sessions();
+                            let agents = known_identities(&signals.read(), &seeds, caps);
                             best_completion(mention.partial, &agents)
                                 .map(|hit| apply_completion(&buf, &mention, &hit.nickname))
                         }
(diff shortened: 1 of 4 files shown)


aaronsb:
## Code Review — `fix/agent-legend-polish`

Reviewed commit range against main. Scope is tight, the two bugs are cleanly separated, and the test suite extension is proportionate to the change. No MUST-FIX.

### SHOULD-FIX

None. The change set is self-consistent.

### CONSIDER

**`DiscoveredSession.session_id` is extracted, tested, and then dropped on the floor.** `chip.rs:196-204` (the seed branch of `known_identities`) builds a `KnownIdentity`, which has no `session_id` field. The signal-branch chips pick up their session_id from `ChipInfo` elsewhere, so seeded entries will render in the `@` legend but won't get a group-membership glyph until a signal lands. The test `discover_carries_session_id` (sessions.rs:144) explicitly docstrings *"seed path must carry them through or discovery becomes legend-only (no membership render for pre-seeded peers)"* — which reads like intent that didn't complete. Two options:

1. Accept and document — add a line to the module doc that session_id is forward-looking (for ADR-124 liveness / group-glyph-on-cold-start), and note the legend-only limitation at the seed call site.
2. Thread it through — add `session_id: Option<String>` to `KnownIdentity` and populate in both branches. The signal branch doesn't have it today (Signal carries `from` / `cwd` but no session_id), so this likely grows beyond this PR.

If #1, the test assertion becomes a "we capture the raw data, downstream consumers will come later" contract — worth saying so explicitly.

**Ordering invariant has no test.** The doc on `known_identities` promises "signal-derived entries take priority so ordering (newest-seen-first) still biases the legend toward active peers; seed-only entries fall in after." The existing `registry_seed_doesnt_duplicate_signal_derived_entry` tests dedup but not order. A one-liner: with one signal for cwd-X and a seed for cwd-Y, assert `reg[0].cwd == "/X"`. Locks the invariant against a future refactor that flips the loop order.

**`extract_json_string` duplication is justified but unmarked.** sessions.rs:71 and sensor-peers/src/lib.rs:771 are byte-identical. The PR description calls the duplication deliberate (avoiding serde_json / cross-crate dep), which is correct for this PR. Worth a one-line comment at the sessions.rs definition pointing at the sensor-peers version so future drift is visible. Cheap documentation, no structural change.

### What's solid

- The dedupe key split (`chip.rs:174-178`) is semantically clean: claudes are cwd-routable so cwd is part of identity; humans aren't, so nickname alone is sufficient. Comment captures the "why" well.
- `discover_in` vs `discover` split is textbook testability — `$HOME` read stays at the edge, the walk is pure against a path.
- Error permissiveness matches the stated stance: `read_dir` failure → empty, `read_to_string` failure → skip one file, missing field → skip one file. No panic path, no bad-file-masks-good-files failure mode. `discover_skips_malformed_json` locks this.
- No liveness filter is the right call for this PR given the stated tradeoff (stale clutter cheap, missing live agent silently breaks routing).
- Three render-path `discover_sessions()` calls are fine at current scale (2 files in my own sessions dir). Matches the `scan_groups` pattern already accepted in the render loop.
- Test coverage of the two bugs is direct: `registry_collapses_same_human_across_cwds` and `registry_seeds_from_sessions_when_buffer_empty` each correspond one-to-one with a reported user symptom.

### Traceability

Links to #23 in the title. The cold-start seeding behavior is a reasonable precursor to the ADR-124 liveness work — this PR establishes the discovery surface that a liveness filter would later wrap.

---

PR #67, opened 2026-04-17
ADR-125: Authored disclosure graph + remove BM25

aaronsb:
## Summary

- Names the existing architecture as an **authored disclosure graph** (formalizes what ADR-105/107/108/110 implied)
- Commits to the embedding model as a **black-box singularity** — sole retrieval tier, hard dependency
- Removes **BM25** entirely (code, corpus fields, frontmatter fields, help text, config, fallback cascade)
- Collapses **1411 per-locale `embed_threshold` entries** into the node-level threshold model (ADR-107 original intent, restored)
- Rewrites `ways tune` as a **locale fidelity audit** — surfaces bad translations to re-author, does not write thresholds

## Motivation

Andrei Bosik (Russian locale) reported `make test-multilingual` failing for 6 languages (es, vi, tr, ko, ru, hi). Root cause: per-locale `embed_threshold` values (0.51-0.78) were higher than the multilingual model's actual query-vs-stub cosine (~0.55). The existing `ways tune` couldn't fix it — it measured corpus-vs-corpus self-discrimination, not real-query recall.

Deeper issue: the per-locale threshold surface was architectural drift from ADR-107's original "hardcoded 0.25, single threshold per way" design. This PR restores that discipline, names the architecture properly, and removes BM25 — which was only ever a fallback for the pre-multilingual era and now actively misleads non-English users (English-only stemmer).

## Changes

**Architecture / docs**
- ADR-125 Draft naming the model; amends ADR-107 (superseded in part) and ADR-108 (BM25 removed)
- Doc sweep across README, install-guide, hooks-and-ways, tests, skills — BM25 references removed or marked historical

**Data plane**
- Strip `embed_threshold` from 1411 `.locales.jsonl` entries (pure serde-safe removal)
- Strip `threshold:` (BM25) from 98 way frontmatter files
- Same cleanup on test fixtures

**Code plane**
- Delete `bm25.rs` and BM25 fallback from scan/scoring/candidates/match
- Drop `bm25_threshold`, `matcher`, `parent_threshold_multiplier` from config
- Drop `bm25_stemmer_for`, `is_bm25_available` from agents
- Remove `--threshold` flag from `ways template` (scaffolded BM25 threshold)
- Rename `cmd/match_bm25.rs` → `cmd/match_cmd.rs`
- Simplify status output (no "configured matcher" concept — one engine)
- `ways tune` rewritten as fidelity audit (621 → 328 lines)

**Tests**
- Session sim fixtures get explicit `pattern:` regex triggers (the test env has no embedding engine, so BM25 was the only channel reaching scan_prompt)
- Fix two pre-existing clippy errors in attend/attend-chat blocking `make lint`

**Net:** 239 files, +2068/-2833 (~765 net deletion).

## Test plan

- [x] `make lint` — passes (after two pre-existing clippy fixes)
- [x] `make test-unit` — passes
- [x] `make test-sim` — 10/10 scenarios pass
- [x] `make test-multilingual` — 18/18 pass (including the 6 previously failing: es, vi, tr, ko, ru, hi)
- [x] `make test-lang` — passes
- [x] `make test-locales` — passes
- [x] `ways tune --way delivery/commits` — produces meaningful fidelity output; flags ru (0.47 min_peer) as the reported translation-quality issue
- [ ] Verify CI green on PR
- [ ] Verify `make setup` still succeeds from scratch (no-BM25 install path)

- Re-author the ~707 locale stubs flagged as low-fidelity by `ways tune` (separate per-way PRs as needed)
- Consider dual-embedding English canonical text into multilingual space so tune can measure English-canonical fidelity directly (currently measures inter-alias agreement, which is a proxy)
- Optionally drop the now-dead `threshold` field from corpus JSONL output (documented as cruft to clean later)


aaronsb:
## Review — ADR-125 (authored disclosure graph + BM25 removal)

**Assessment: ship-with-changes.** Architectural direction is solid and the ADR is well-argued. One real regression found, plus several stale doc references that should go with the PR rather than after.

### Critical (blocker)

**Context-threshold percentages stripped from 3 meta ways.** The migration that removed BM25 `threshold:` fields also removed the *percentage* thresholds on `trigger: context-threshold` ways. These are not BM25 — they are the "fire when transcript reaches N% of window" knob read by `scan/mod.rs::evaluate_context_threshold`.

Affected:
- `hooks/ways/meta/todos/todos.md` — lost `threshold: 75`
- `hooks/ways/meta/memory/memory.md` — lost `threshold: 80`
- `hooks/ways/meta/compaction-checkpoint/compaction-checkpoint.md` — lost `threshold: 95`

With no threshold field, `WayCandidate.threshold` defaults to `0.0` (`cmd/scan/candidates.rs:131-133`), so `evaluate_context_threshold(0, ...)` computes `limit = window_chars * 0 / 100 = 0` and returns true for any non-empty transcript. These ways will fire every state scan from turn 1. todos/memory re-inject on every check. Fix: restore the three values.

Separately, the field overloading is fragile: the same YAML key `threshold:` used to mean a BM25 float (e.g. 2.0) and now means a context percentage (e.g. 75). `candidates.rs:131` parses whatever's there. Consider renaming to `context_threshold_pct:` or similar in a follow-up so this class of accident can't recur.

### Concerns

1. **Leftover BM25 references outside the diff.** The PR scoped to code and data but missed user-facing surfaces that still promise BM25:
   - `hooks/ways/check-setup.sh:46-55` — tells users "Embedding model not installed — using BM25 fallback (91% vs 98% accuracy)". Post-PR this is false; there is no fallback, matching silently returns nothing.
   - `scripts/install.sh:310` — "BM25 fallback is still active" in the setup-failure message.
   - `hooks/ways/frontmatter-schema.yaml:20,25,206` — describes `vocabulary` and `threshold` as BM25 fields. Anyone authoring a new way against the schema is reading stale guidance.
   - `hooks/ways/meta/knowledge/optimization/macro.sh:24,33` — still classifies matches as "BM25" type in output.
   - `tools/ways-cli/languages.json` — every language entry has `bm25_stemmer`. Not load-bearing (nothing reads it post-PR) but dead.

2. **Silent degradation on embedding-engine absence in `scan prompt`.** `ways match` exits 1 with a clear error, which is good. But `cmd/scan/mod.rs::prompt` (the hot path every turn) silently continues with `embed = None`, so pattern-marked ways fire and semantic-only ways silently stop. The warning in `scoring.rs:27` is stderr-only. Hooks swallow stderr, so a user whose model is missing gets a degraded session with no in-context signal. Consider either (a) emit a one-shot additionalContext warning, or (b) document that this is intentional in the ADR's "Consequences — negative" section.

3. **`WayCandidate.threshold` comment partially accurate.** `candidates.rs:129-130` says "Post-ADR-125, no semantic/BM25 meaning; default 0.0 is never compared for other triggers." True for non-context-threshold ways, but the bug above shows the default IS compared (as `0 as u64`) when someone forgets to set the percentage. A `.unwrap_or(0.0)` on a percentage field is a trap.

### Nits

- `cmd/corpus.rs:226,260` — still emits `embed_threshold` (0.35 for EN, 0.25 for locale) in the corpus JSONL. PR acknowledges this as follow-up cruft. Fine to leave, but `frontmatter::LocaleEntry::embed_threshold` could also be dropped from the struct since 1411 entries no longer carry it.
- `tune.rs:233` — `if score > 0.999 { continue; }` self-score filter is correct for the "query's own row in corpus" case, but does not guard against genuine near-duplicate peer aliases (two locales with ~identical vectors). In practice multilingual stubs shouldn't hit 0.999 cosine against each other, but if it ever happens the peer gets silently excluded. Consider `score >= 0.9999` or matching against the specific query-row id.
- `tune.rs` does not depend on corpus stub ordering — it filters by `id == way_id` and builds peer_scores unordered. No bug there.
- `config.rs` uses manual field extraction on `serde_yaml::Value`, so unknown `bm25_threshold:` / `matcher:` / `parent_threshold_multiplier:` fields in user yaml/json are silently ignored. Acceptable per ADR.
- ADR-107/108 amendments are accurate. Neither names a stale downstream ADR that I could find. ADR-116 and ADR-115 mention BM25 peripherally but don't depend on it architecturally.

### Strengths

- Naming the model ("authored disclosure graph") is the right move — the ADR does real architectural work, not just a retroactive label.
- Test fixtures correctly migrated to `pattern:` regexes; every scenario's query matches its fixture's pattern (verified scenarios 1,4,5,6,7,9). Domain-disabled filtering happens before pattern matching, so scenario 9 remains valid.
- `ways tune` rewrite is cleaner than what it replaces (621 → 328 lines) and its fidelity metric is defensible: cross-alias cosine is a direct measure of "do these translations mean the same thing to the model."
- Self-PR hygiene is good: ADR amendments filed, pre-existing clippy fixes called out separately, follow-up cleanups listed explicitly.

---

PR #68, opened 2026-04-17
fix: per-model thresholds + signal analysis tool

aaronsb:
## Summary

Real regression found right after ADR-125 merged: matcher over-fired during a shell-config session (~75 ways across 30 turns, including `ea/*`, `writing`, `research`, `itops/*`, `visualization`, `delivery` — all unrelated to dotfiles). Root caused to ADR-125's single-threshold + `max(en, multi)` design, fixed with per-model thresholds; added a signal-analysis tool so the calibration is data-driven from here on.

## Root cause

The two embedding models produce scores in **different distributions** — they're not comparable. Taking `max(en_score, multi_score)` meant one noisy multilingual alias could fire a way whose English canonical scored 0.10. With the default threshold at 0.35, the multilingual noise band (0.30–0.50 from generic-verb collisions like "configure", "setup", "review") was sitting right in the firing zone.

## Fix (commit `d9b9433`)

Each model path is gated independently:

- `default_embed_threshold: 0.40` (EN; per-way override via frontmatter `embed_threshold:`)
- `default_multi_embed_threshold: 0.55` (multi; uniform — locale aliases don't carry thresholds per ADR-125)
- `parent_boost_floor: 0.40` — cascading parent-boost (0.8x multiplier) can't push children into the multilingual noise band

`batch_embed_score` now returns `EmbedScores { en, multi }` with independent score lists. `match_prompt` fires if **either** path clears its own bar. The trigger label is `semantic:embedding:en` or `semantic:embedding:multi` so downstream tooling knows which model carried the signal.

No cross-model comparison, no normalization, no "confirmed-by-both" confidence tier. Each model stands alone on its own threshold calibrated for its own distribution.

`ways match` output grew to show both EN and multi columns per way, sorted by per-row max. Users diagnosing a firing can see which model carried the signal.

## Verification

Shell-config queries that previously over-fired:

| Query | Before | After |
|---|---|---|
| "configure zsh prompt and aliases in .zshrc" | 5 fires (incl. `code/security/injection` FP) | 2 fires, both shell-family |
| "add aliases to my zshrc" | 4 fires | 1 fire |
| "review this bindkey configuration" | 8 fires | 1 fire |
| "set up starship prompt for zsh" | 5 fires | 1 fire |

`make test` green — lint, 10/10 session_sim, 18/18 multilingual, lang, locales.

## Signal analysis tool (commit `3a2d5a2`)

`scripts/signal-report.py` runs a prompt battery through both models with `--threshold 0.0`, classifies each row as signal (expected_way match) or noise (anything else), plots distributions with thresholds overlaid, dumps raw CSV. Baseline run against the 16-prompt default battery is committed to `docs/signal-analysis/` so reviewers can see the plots without regenerating.

The baseline numbers that justify the per-model thresholds:

| model | signal min | signal mean | noise p95 | noise p99 | threshold |
|-------|-----------:|------------:|----------:|----------:|----------:|
| EN    | 0.01*      | 0.36        | 0.24      | 0.31      | 0.40      |
| multi | 0.54       | 0.67        | 0.45      | 0.59      | 0.55      |

\* `signal_min 0.01` for EN is from non-English prompts that the EN model can't understand — those are correctly picked up by the multi path instead.

Thresholds sit in the gap between `signal_min` and `noise_p95` for each model independently. Multi has ~1% p99 noise leakage that `ways tune` flags as discrimination problems (stub-level fixes, not threshold-level).

## Known acceptable limitation

The multi threshold at 0.55 is **tight** — the signal battery's `signal_min` is 0.5406 (one signal rides just below the gate), and 15 noise rows in `docs/signal-analysis/scores.csv` score above 0.55. These are stub-level discrimination problems that `ways tune` flags as top confusers; the follow-up is re-authoring those stubs, not loosening the threshold. Raising the threshold further would cut legitimate matches; lowering it would readmit the noise the PR is meant to kill.

## Test plan

- [x] Unit tests, session_sim (10/10), multilingual (18/18), lang, locales
- [x] Lint clean
- [x] Smoke test: 4 shell queries → only shell-family ways fire
- [x] Smoke test: multilingual queries still route correctly via multi path
- [ ] CI green on PR
- [ ] Eyeball the plots in `docs/signal-analysis/` — do the distributions look right?

## Out of scope

- Further stub sharpening (use `ways tune` output to pick targets)
- Split `scan/mod.rs` (now at 657 lines, approaching the 800-line threshold)
- Language-detection-based routing (considered, not adopted — union-fire with proper thresholds is simpler and handles the same cases)


aaronsb:
## Review: ship-with-changes

One real regression to fix, one observation about the threshold calibration that deserves surfacing before merge. Otherwise the design is clean — per-model gating is correct, the struct refactor (`EmbedScores`) reads well, and dead code from ADR-125 is fully cleared out. Good surgical change.

### Critical — trigger label rename breaks pretty-printing

`tools/ways-cli/src/cmd/render.rs:485`

The new `match_prompt` emits `semantic:embedding:en` / `semantic:embedding:multi` (scan/mod.rs:367-369), but `format_trigger` still matches only `"semantic:embedding" | "semantic"` -> `"embed"`. The new labels fall through to the `_ =>` arm and display as the raw string. Two call sites are affected: `render.rs:176` and `rethink.rs:469`.

Two ways to fix:
- Add arms for `"semantic:embedding:en"` and `"semantic:embedding:multi"` (distinct labels like `embed:en` / `embed:mu` if you want the provenance visible, or collapse both to `"embed"` if you don't).
- Or match by prefix: `s if s.starts_with("semantic:embedding") => "embed".into()`.

Prefix match preserves downstream tooling's ability to parse the specific model from the raw trigger while keeping the display terse. Either way, the current code silently degrades UX.

### Concern — multi threshold 0.55 is tighter than the PR body implies

Recomputing from `docs/signal-analysis/scores.csv`:

- multi signal_min: **0.5406** (ru commits prompt)
- multi noise_p99: **0.5525**
- multi noise_p95: **0.4456**

The PR body lists signal_min 0.54 and noise_p99 0.59 for multi. The p99 looks off — my numbers may differ because I'm counting expected_way mismatches as noise including the blank-expected adversarial prompts. Regardless, at threshold 0.55:

- One signal row (ru commits, 0.5406) sits **below** the gate → miss.
- Fifteen noise rows sit **above** 0.55, several of which are unambiguous FPs: `meta/project-health` (0.61, 0.72), `softwaredev/docs/standards` (0.66), `softwaredev/environment/makefile` (0.65), `softwaredev/code` (0.61), `softwaredev/delivery/implement` (0.55, 0.64).

Many other "noise" hits above 0.55 are actually family-adjacent (adr/migration when expected=adr, testing/mocking when expected=testing) — those are arguably fine or even desirable. But the genuine FPs are real.

This isn't blocking — 0.55 is a defensible starting point and the PR body correctly says tuning discrimination at the stub level is a separate follow-up. Worth calling out explicitly in the PR description that (a) the multi gate is tight and at least one real-world signal (ru "закоммитить") rides right at the edge, and (b) the `meta/project-health` stub appears to be leaking into German/English prompts across multiple intents.

### Worth discussing

**EN signal_min 0.01 data artifact.** The non-English prompts in the EN test produce near-zero scores that pull signal_min down. The PR body asterisks this correctly. Consider splitting the signal-report battery into "EN prompts only for EN model stats" and "multilingual prompts only for multi model stats" so the reported signal_min per model is meaningful without the footnote. Current setup is fine for eyeballing the plots, less good as an ongoing calibration tool.

**`apply_yaml_overrides` test doesn't cover new fields.** `config.rs:208-214` tests YAML parsing for `language` and `parent_threshold_multiplier` but not `parent_boost_floor` or `default_multi_embed_threshold`. Parsing is mechanical and the `default_values` test at 197 covers the defaults, but a one-line addition to the YAML test would close the loop.

### Nits

- `match_cmd.rs:104` the `#[allow(dead_code)]` on `default_corpus_path` — if it's truly unused after the rework, delete it rather than suppress.
- `scan/mod.rs:367-371` the `if en_fires { ... } else if multi_fires { ... }` ordering means EN label wins when both fire. Fine, but if you add a `both` label later for confidence, this is where it goes.
- `scoring.rs:19` `best_score` implementation is functional but dense; a plain loop would read as well and avoids the `Option<f64>` fold ceremony. Personal taste — leave it.
- PR body already flagged that `scan/mod.rs` is at 652 lines (approaching 800). Agreed, not now, but the scan functions (prompt, task, command, file) have enough shared shape that a helper extraction would be worthwhile next pass.

### Strengths worth naming

- `EmbedScores { en, multi }` is the right abstraction. Renders the "apples-to-oranges" invariant into the type system — no caller can accidentally cross-compare.
- Per-model threshold plumbing is uniform: `effective_thresholds` returns both numbers, `match_prompt` and `check_semantic_score` each gate independently, `ways match` shows both columns. Consistent mental model end-to-end.
- `parent_boost_floor` is a small addition with a clean story: cascades can't push the bar into the noise band. One line of defense against a known failure mode.
- Committing `docs/signal-analysis/` with plots + CSV makes the calibration reproducible and auditable. Good discipline for a threshold-tuning change.

PR #71, opened 2026-04-22
feat(ways): ADR-126 window-relative refire with named presets

aaronsb:
## Summary

Implements ADR-126: `refire:` frontmatter field that accepts either a numeric fraction of the session context window or a preset name. Resolution happens at fire-evaluation time — the engine multiplies by whatever `model_to_window()` returns — so way files stop encoding host-specific token counts. Way frontmatter becomes portable across any agent framework that can report its context capacity.

## Design pillars

- **Dual form, one field.** `refire: 0.2` (numeric, pinned to today's model) and `refire: rare` (preset, tracks project config) coexist. Numeric is the primitive; presets sit on top.
- **Engine boundary preserved.** `sensor-trait` is untouched. The resolver lives entirely in `ways-cli`; the engine sees only concrete `Curve::Exponential { half_life }`. ADR-123's unit-agnostic engine boundary remains intact.
- **Presets are external references, not opinions.** The `refire_presets:` section in `$XDG_CONFIG_HOME/ways/config.yaml` (with project overlay — ADR-115) defines the resolution table. Re-tuning for a new model is a one-file edit, not a 95-way sweep.
- **Framework portability** is the strong claim (fractions generalize over any finite capacity); model-generation preset portability is a weaker sub-claim documented alongside.
- **Binary lint checks.** After simplification discussion, the lint surface is two diagnostic rules: "fire-bearing way missing `refire:`" (WARNING) and "curve: block is foreign/unknown" (UNKNOWN). No preset-matching info suggestion, no `--fix-preset` action — complexity pruned in favor of clarity.

## Commits

1. **`d92bf2e` docs(adr)** — ADR-126 Draft with the full design, alternatives considered, four-phase migration plan, portability framing. Renames earlier `banded-refire-abstraction-for-ways.md` to `window-relative-refire.md` to match the refined thesis.

2. **`4245fda` feat(ways)** — Phase 1 engine:
   - `RefireSpec` enum in `frontmatter.rs` (untagged `Numeric(f64) | Preset(String)`).
   - `Frontmatter::resolved_curve(window)` resolver; `refire:` wins over `curve:` when both present.
   - `refire_presets: HashMap<String, f64>` in `config.rs` with built-in defaults (once/rare/normal/frequent), YAML section parsing, project-overlay compatible.
   - `way_refire_threshold_k` takes a window parameter; call sites (`show`, `list`, `rethink`) fetch the window once via `cmd::context::get_context()` and thread it through.
   - 9 new unit tests covering parse, resolution, window scaling, precedence, fallbacks, and defensive clamp.

3. **`72c86bb` chore(ways)** — Phase 2 mechanical migration. 92 way files (plus 6 untracked `workstation/` ways carried along) converted from `curve: Exponential + half_life` to `refire: <fraction>` using each file's window-at-tuning-time:
   - 14 ADR-127-hacked files (1M-tuned) → `refire: 0.2`
   - 78 files at default 30k (200k-era) → `refire: 0.15`
   - 3 files at 20k outlier → `refire: 0.1`

4. **`dd1a72d` feat(ways)** — lint enforces the ADR-126 schema:
   - `hooks/ways/frontmatter-schema.yaml`: `curve:` block removed, `refire:` added.
   - `per_file.rs`: WARNING when a fire-bearing way (description + vocabulary, not attend, not a check file) has no `refire:`. The existing unknown-field handler picks up `curve:` as foreign automatically.
   - ADR-126 §"Frontmatter and lint changes" updated to reflect the simpler binary-check shape.

5. **`ffbbbac` docs(ways)** — teach the authoring surface:
   - `meta/knowledge/authoring/authoring.md` (the way that fires when agents edit way files): adds a "Firing cadence (refire:)" section with numeric vs preset forms, common cadence choices mapped to intent, and explicit note that `curve:` now triggers lint UNKNOWN warnings.
   - `tools/ways-cli/src/cmd/template.rs`: `ways template` scaffolder emits `refire: 0.15` as default in new ways. First-run lint-clean.
   - `docs/hooks-and-ways/context-decay.md`: theoretical doc updated — table row renamed "per-way curve" → "per-way refire"; follow-up paragraph describes refire semantics instead of `Curve::Exponential` half-life.

6. **`c019749` chore(ways)** — template + test fixtures adopt refire:
   - `init.rs TEMPLATE_CONTENT` (written to new projects as `.claude/ways/_template.md` by `ways init`): rewritten top-to-bottom. Empty frontmatter so the template never matches at runtime; body shows canonical frontmatter (`description`/`vocabulary`/`refire`/`scope`) and a cadence reference table. Removed obsolete `threshold: 2.0` example and the pre-ADR-123 "loads once per session" language.
   - 9 test fixtures converted (`curve: Exponential half_life: 30000` → `refire: 0.15`); `with-check.check.md` had no curve block and stays untouched.

## Behavior change

Pre-PR: the 81 unhacked ways at `half_life: 30000` fire ~22 times per session on 1M Opus (the "5 fires" problem from ADR-126, generalized across the tree). Post-PR: those same files fire ~3–4 times — the resolver multiplies `refire: 0.15` by the actual 1M window, closer to original design intent.

The 14 ADR-127-hacked files behave identically (`refire: 0.2 × 1M = 200k` = the hacked value). 200k-window sessions (Sonnet/Haiku) continue to fire at the original 3×/session cadence (`refire: 0.15 × 200k = 30k` = the original value). Side-effect fix aligned with authorial intent everywhere.

## Verification

- [x] `cargo test` — 44 unit + 10 scenario tests (54 total) pass. 9 new tests in `frontmatter.rs`, 1 new in `config.rs`.
- [x] `ways lint --global` — 0 errors, 0 refire warnings, 1 pre-existing unrelated warning (`macro: without requires:` on the ADR migration helper).
- [x] `ways lint` on project-local `.claude/ways/_template.md` — clean.
- [x] Synthetic regression (reverting `quality.md` to a `curve:` block): lint surfaces both UNKNOWN-curve and missing-refire warnings as designed.
- [x] Project-scope smoke test — created `/tmp/refire-smoke` with numeric-form, preset-form, missing-refire, and legacy-curve project-local ways plus a `ways.yaml` overriding `rare: 0.35` and adding `project_custom: 0.25`. All four lint states fire correctly; `ways config show` confirms the project overlay is applied to `refire_presets`.
- [x] Fixture tests exercise project-scope ways end-to-end through the full firing engine post-migration (scenario_1 through scenario_10 all pass).

- **Phase 3** — per-way authorial judgment on which ways should use preset names vs pinned numbers. Incremental, not mechanical.
- **Phase 4** — lint promotes `half_life:` in frontmatter from a no-op to an error in a later minor release, once Phase 3 is settled.
- **`rethink.rs` is 922 lines** (over the 800-line quality threshold). My change here was a 3-line param plumbing addition, but the file deserves splitting — separate concern, separate PR.


aaronsb:
## Code Review — ADR-126 refire

Thorough change, design is clean, tests are mostly solid. Several gaps between the ADR and the implementation worth closing before merge. Items ordered by severity.

---

### Must-fix

**1. ADR says unknown presets error; implementation silently falls back.**
ADR-126 §"Frontmatter shape" (line 88-89): *"A string is looked up in the `refire_presets` config section at fire time. Unknown names are an error (likely typo; failing closed is safer than silent default)."*

`tools/ways-cli/src/frontmatter.rs:37-44` does the opposite — it emits a stderr warning and returns `0.15`. Because resolution happens at fire time (not parse time), lint can't catch the typo either. A way with `refire: rrare` or `refire: nromal` will load without incident, run with `normal` cadence, and the stderr warning surfaces only when the way fires — easy to miss.

Pick one:
- Update the ADR to document fail-soft semantics and the rationale (a misspelled preset is less catastrophic than a parse failure that breaks the entire frontmatter load).
- Change `fraction()` to return `Result<f64>` and have `to_curve()` / `resolved_curve()` propagate — so a misspelled preset surfaces as a fire-site error, matching the ADR's failing-closed intent.

Either is defensible; the inconsistency is the bug.

**2. Lint "fire-bearing" under-covers trigger-only ways.**
`tools/ways-cli/src/cmd/lint/per_file.rs:148`: `is_fire_bearing = !is_check && has_desc && has_vocab && !is_attend`.

But `show::way()` at `tools/ways-cli/src/cmd/show/mod.rs:53-58` hard-errors when neither refire nor curve is resolved. So a way with only `files:` / `commands:` / `pattern:` / `trigger: session-start` and no description+vocabulary can exist today without refire (lint won't flag it) yet will still dispatch through `show::way()` via `capture_show_way` in `cmd/scan/mod.rs:541`, hitting the resolved_curve error.

Migration happened to cover every such way in the current tree (confirmed — `find hooks/ways -name "*.md" -not -name "*.check.md" | xargs grep -L "^refire:" | xargs grep -l "^pattern:\|^files:\|^commands:\|^trigger:"` returns empty). But this is a foot-gun for future authors, especially for state-triggered or files-triggered ways that are common enough patterns. Broaden `is_fire_bearing` to also cover ways with any of `pattern:`/`files:`/`commands:`/`trigger:` and no refire.

---

### Should-fix

**3. Preset resolution has no test coverage.**
All 9 new tests in `frontmatter.rs` exercise `RefireSpec::Numeric(..)`. None test:
- `RefireSpec::Preset("rare").fraction()` returns the expected 0.4
- Unknown preset name returns the 0.15 fallback (and logs to stderr)
- Project-overlay preset (e.g. `rare: 0.35` in ways.yaml) flows through `fraction()` → `to_curve()`

The headline new capability — presets — is untested. `config.rs` has one apply_yaml test but it doesn't exercise the RefireSpec→Curve path. At minimum add:
```rust
#[test] fn preset_resolves_from_default_config() { … }
#[test] fn unknown_preset_falls_back_to_normal() { … }
```

Testing the project-overlay flow is trickier because `config::global()` is a `LazyLock`, but that itself suggests a testability seam worth addressing (take `&Config` as a parameter to `fraction` rather than reaching for the global).

**4. Documentation drift: `refire: 0.2` and `refire: rare` are not equivalent.**
`hooks/ways/meta/knowledge/authoring/authoring.md:94` says static-heavy payloads want `refire: 0.2` OR `refire: rare`. Same "either/or" pairing in `tools/ways-cli/src/cmd/init.rs:61`. But per the preset table in `config.rs:68`, `rare = 0.4` — twice the cadence of 0.2. On a 1M window that's ~2 fires/session vs ~5 fires/session. Authors reading "use rare or 0.2 interchangeably" will pick inconsistently.

Pick one:
- Change `rare` default to `0.2` (aligning with ADR-127 hacked files' intent and the PR's migration of those 14 files to `refire: 0.2`).
- Change the doc tables to map 0.2→(no preset), 0.4→rare.

The first is probably cleaner — `rare` then lines up with the existing ADR-127 hack as "1–2 fires per session on a large window," and `0.4` slips up toward `once` territory faster than authors likely want.

**5. `rethink.rs` uses a different window-detection path than `show.rs` / `list.rs`.**
- `cmd/show/mod.rs:50` and `cmd/list.rs:60`: `crate::cmd::context::get_context(…)` → `model_to_window()` which honors `CLAUDE_CONTEXT_WINDOW` env override for unknown models.
- `cmd/rethink.rs:158`: `session::detect_context_window_for(…)` → `context_window_from_transcript()` in `session.rs:231-260` which only handles `opus-4 → 1M` with a bare 200k fallback, no env override.

Pre-PR this divergence was cosmetic (rethink renders a bar). Post-PR it determines the half-life used to render per-way refire thresholds in replay mode, so a rethink session on an unknown model (future Sonnet 5, etc.) will render different thresholds than `ways list` shows on the same transcript. Consolidate on one window-detection code path, or document why rethink uses the narrower one.

**6. The "refire + curve duplication" warning from the ADR doesn't exist.**
ADR-126 §"Frontmatter and lint changes" bullet 1 says: *"If `curve:` is also present, `refire:` wins and lint emits a warning about the duplication."* Implementation just flags `curve:` as a generic `UNKNOWN: unknown field 'curve'` via the unknown-field handler. The message doesn't mention refire precedence or ADR-126 — it just says "unknown."

Either simplify the ADR to document that `curve:` is now treated as foreign/unknown like any other removed field (which is what the PR actually did, and the commit message confirms was deliberate after simplification), or add a specific diagnostic: `WARNING: {rel} — legacy 'curve:' block present (ADR-126). Replace with 'refire: <fraction|preset>'.`

---

### Consider

**7. Attend handlers got a 5× longer refire window on Opus.**
`meta/attend/build-complete`, `context-pressure`, `reflection-overdue` were at `half_life: 30000` → now `refire: 0.15` → 150k on 1M Opus. Attend handlers are signal-driven, not prompt-driven — the cadence semantics for "signal fires twice" are different from "user says the thing twice." Per the ADR's "window-at-tuning-time" rule this is formally correct, but signal handlers specifically might want to stay on absolute half-life. Worth a look per-handler: does context-pressure re-firing after 30k make more sense than after 150k? Build-complete is probably fine either way. Not a blocker.

**8. `to_curve(0)` edge case.**
`RefireSpec::Numeric(0.5).to_curve(0)` gives `(0.5 * 0.0).round() as u64 = 0`, then clamped to 1. Fine in the happy path. But a negative numeric (`refire: -0.5`) gives a negative product, which then hits `.round() as u64` — a saturating cast in Rust that produces 0, then clamped to 1. Silently swallowed. A `> 1.0` value (e.g. `refire: 2.0`) just extends the half-life beyond the window, which is fine for `once`-like intent but probably not the author's plan. Consider adding an `Invalid` guard at parse (bounds-check to `[0.0, 2.0]` or similar) with a lint warning, or at least a stderr warning at resolution time for negative values.

### Nice

**9. `config::global()` in `RefireSpec::fraction()` is a LazyLock reach.**
Makes the method hard to unit-test with a synthetic preset table. Future refactor (outside this PR): thread `&Config` through `resolved_curve(window, cfg)` so fraction resolution is injectable.

**10. `fraction()` does a HashMap lookup every fire.**
Negligible in the current hot path but if the resolver ever ends up in a tight loop, consider memoizing. Not important now.

---

### Question for the author

The ADR/implementation mismatch on unknown-preset behavior (item 1) needs a judgment call: **fail-closed (match the ADR, harder to introduce typos) or fail-soft (match the current code, more forgiving)?** The answer shapes items 3 and 4 slightly — if fail-closed wins, the "unknown preset" test becomes "unknown preset errors"; if fail-soft wins, the ADR gets a text update.

---

### Overall

**Merge with fixes.** The design is sound, the engine boundary discipline is impressive, the migration reasoning is correct, and the 9 tests that exist are well-chosen. But items 1, 2, and 4 are user-facing rough edges: the first is an ADR-code contradiction on a silent-fallback path, the second is a missing lint check that makes future authoring riskier, the third is a preset-vs-number inconsistency across three documentation surfaces. None is architecturally deep; all three are short edits. Item 3 (preset test coverage) is worth adding in the same pass so the headline capability has regression coverage.

*AI-assisted review via Claude (Opus 4.7)*

PR #73, opened 2026-04-22
refactor(ways): declare fire-bearing fields once for lint

aaronsb:
## Summary

- Moves the fire-eligibility field list (`pattern`, `files`, `commands`, `trigger`) into `frontmatter::FIRE_BEARING_FIELDS` and exposes `fires_on_something(fm_str)` alongside it.
- Lint's five-disjunct predicate in `per_file.rs` becomes a single call; adding a new trigger channel is now a one-line edit to the const.
- No runtime change — there's no runtime-side fire-eligibility predicate to unify with. Each scan entry point dispatches on only the fields it handles. The const's doc comment makes this explicit (advisory declaration, not a dispatcher).

## Context

From the post-ADR-126 simplification plan (`docs/design-notes/post-adr-126-simplification.md`, Opportunity 3). Initial scoping suggested unifying runtime dispatch with lint; investigation found the runtime has no such predicate, so the refactor is scoped to eliminating lint's drift risk when new channels get added. Opp 1 (delete dead `Curve` variants) will follow in a separate PR; Opp 2 is dropped.

## Test plan

- [x] `cargo test -p ways --bins` — 64 tests pass (6 new on `fires_on_something`, covering semantic channel, iteration over the const, and non-channel fields).
- [x] `ways lint --global` on the full tree — same output as before (0 errors, 1 pre-existing unrelated warning for `macro: without requires:`).
- [x] No runtime changes; existing dispatch paths unchanged.

### DIFF
--- a/tools/ways-cli/src/cmd/lint/per_file.rs
+++ b/tools/ways-cli/src/cmd/lint/per_file.rs
@@ -137,24 +137,16 @@ pub(super) fn lint_file(
         }
     }
 
-    // ADR-126: fire-bearing ways should carry a `refire:` field. A way is
-    // fire-bearing if any of its trigger channels are wired: semantic
-    // (description + vocabulary), regex (pattern/files/commands), or state
-    // (trigger:). Check files and attend signal handlers are exempt —
-    // checks ride on their parent way's firing, and attend handlers are
-    // triggered by signal name rather than the refire engine.
+    // ADR-126: fire-bearing ways should carry a `refire:` field. Fire
+    // eligibility (which frontmatter fields count as firing channels) is
+    // declared in `frontmatter::FIRE_BEARING_FIELDS` so adding a new channel
+    // flows through automatically. Check files and attend signal handlers
+    // are exempt — checks ride on their parent way's firing, and attend
+    // handlers are triggered by signal name rather than the refire engine.
     let has_refire = has_field(&fm_str, "refire");
     let has_curve = has_field(&fm_str, "curve");
-    let has_pattern = has_field(&fm_str, "pattern");
-    let has_files = has_field(&fm_str, "files");
-    let has_commands = has_field(&fm_str, "commands");
-    let has_trigger = has_field(&fm_str, "trigger");
-    let fires_on_something = (has_desc && has_vocab)
-        || has_pattern
-        || has_files
-        || has_commands
-        || has_trigger;
-    let is_fire_bearing = !is_check && !is_attend && fires_on_something;
+    let is_fire_bearing =
+        !is_check && !is_attend && crate::frontmatter::fires_on_something(&fm_str);
     if is_fire_bearing && !has_refire {
         eprintln!(
             "  WARNING: {rel} — no `refire:` field (ADR-126). \
--- a/tools/ways-cli/src/frontmatter.rs
+++ b/tools/ways-cli/src/frontmatter.rs
@@ -149,6 +149,34 @@ impl Frontmatter {
     }
 }
 
+/// Frontmatter fields that wire a way into a runtime dispatch channel.
+/// A way carrying any of these participates in firing and therefore needs
+/// a `refire:` field (ADR-126). The semantic channel (description +
+/// vocabulary) is checked separately by [`fires_on_something`] because it
+/// requires both fields together.
+///
+/// **Adding a channel:** extending this list picks the new field up in
+/// `ways lint`'s fire-eligibility check automatically. The corresponding
+/// dispatch entry point in `cmd/scan/` still needs wiring separately — this
+/// const is the advisory declaration, not the dispatcher.
+pub const FIRE_BEARING_FIELDS: &[&str] = &["pattern", "files", "commands", "trigger"];
+
+/// Does this frontmatter wire the way into any firing channel?
+/// Used by `ways lint` to flag fire-bearing ways that lack a `refire:` field.
+///
+/// Operates on the raw YAML string rather than the parsed [`Frontmatter`]
+/// struct because several channel fields (`pattern`, `files`, `commands`,
+/// `trigger`) aren't modeled on the serde type. Keeping the predicate
+/// string-based lets it run against partially-valid frontmatter during lint.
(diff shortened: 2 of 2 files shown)


aaronsb:
## Review

### What this delivers

Narrow, honest, and well-scoped. The initial framing in the design note (Opp 3) proposed "lint and runtime share one predicate," but you verified there's no runtime-side predicate to share — each scan entry point in `cmd/scan/scoring.rs` (lines 190, 287, 489, 503) dispatches on only the field it handles. Re-scoping to "eliminate drift in lint's hardcoded field list" is the correct call, and the `FIRE_BEARING_FIELDS` const doc (`frontmatter.rs:152-161`) says exactly that: "advisory declaration, not the dispatcher." Future readers won't mistake it for something it isn't.

The lint call site shrinks from 9 lines of disjuncts to one call (`per_file.rs:148-149`), and the comment block above now points to the const as source of truth. That's the whole win — no more, no less.

### On the stated claim

Yes: if someone adds a new trigger channel (e.g., `event:`), adding `"event"` to `FIRE_BEARING_FIELDS` is sufficient for lint to flag `event:`-bearing ways lacking a `refire:`. I grepped the codebase — no other hardcoded list of these four channel names exists to drift from. The `scoring.rs` dispatch sites read fields by name individually (one channel per dispatcher), which is the architecture you accurately described, not a duplication of the full list.

One subtlety worth noting in the doc comment: the semantic channel isn't *declared* in the const (it's hardcoded in `fires_on_something` because it requires both `description` AND `vocabulary`). Your doc at line 155-156 mentions this, but a reader scanning only `FIRE_BEARING_FIELDS` might miss that semantic is also fire-bearing. Not a blocker — the `fires_on_semantic_channel` test anchors the behavior.

### Scope judgment

Right-sized. You resisted three things that would have been busywork:
- Enum-modeling the channels (`enum Channel { Pattern, Files, ... }`) — would require threading through dispatchers that already work fine.
- Unifying `Frontmatter` (serde) with `WayCandidate` (raw-string `get_fm_field`) — a real seam, but one with deliberate reasons (partial-validity parsing for lint), and not what this PR is about.
- Extracting a shared `has_field` primitive across `frontmatter::fires_on_something`, `lint::helpers::has_field` (line 56), and `scan::candidates::get_fm_field` (line 164) — three near-identical string-parse helpers.

That last one is the most defensible future PR, but your self-containment trade-off holds for now: the three helpers diverge in return type (`bool` vs `Option<String>`) and in what counts as "value present" (empty-value handling). Collapsing them is a separate, larger refactor with its own design choices. Don't do it as a rider on this one.

### Code quality on `fires_on_something`

The nested `fn has` closure (lines 172-175) duplicates `lint::helpers::has_field` verbatim. Self-containment is defensible here — `frontmatter.rs` currently has no dependency on `cmd/lint/`, and inverting that would be wrong (lint depends on frontmatter, not vice versa). Extracting a third shared helper to a `util/` or `yaml/` module would be strictly correct but would also be the smallest imaginable change adding the largest-possible blast radius. Leave it.

### Tests

The six new tests are well-chosen. `fires_on_each_declared_channel` (lines 620-632) is the keystone — iterating the const catches the exact drift class the refactor is meant to prevent. `does_not_fire_on_empty_or_non_channel_fields` (lines 634-642) correctly exercises the negative cases including `curve:` and `refire:` themselves (which are cadence, not firing).

Edge cases not covered, ordered by whether they matter:

- **Indented top-level field** (`  pattern: foo`): `starts_with("pattern:")` returns false, so `fires_on_something` misses it. But the surrounding code (`lint::helpers::has_field`, `scan::candidates::get_fm_field`) has the same behavior — this is a consistent codebase convention, and production way files never indent top-level fields. `detect_legacy_redisclose` handles indentation specifically because of a known yaml-style variant, not because it's expected. Don't change this without changing the other two.
- **Comment before field** (`# pattern: regex-goes-here`): not currently an issue because no production way file comments out channels, but the predicate would false-positive if someone did. Low-priority follow-up.
- **`pattern:` as a sub-field** inside a larger YAML block (e.g., `when:\n  pattern: x`): would false-positive. Again, not observed in production and would also fool `has_field`. Same class as above.

None of these are new to this PR — they're inherited from the codebase's string-parse convention. A nice-to-have follow-up would be to tighten all three helpers together.

### Home of the const

`frontmatter.rs` is defensible. Arguments:

- **For `frontmatter.rs`**: the const describes a property of frontmatter (which fields wire firing). Lives next to the `Frontmatter` struct and `parse`. One module, one concept.
- **For `cmd/scan/`**: the actual dispatchers are there. But — and this is the real answer — the dispatchers don't consume the const. They each handle one field. Co-locating the advisory list with dispatchers it doesn't drive would be misleading.
- **For a new `channels.rs`**: premature. With one consumer (lint), a new module is ceremony.

Current placement is right.

### ADR traceability

Good. The const doc references ADR-126, the lint comment block (`per_file.rs:140-145`) references ADR-126 and points to `FIRE_BEARING_FIELDS`. PR body traces to the design note and explains the scope shift honestly. A future reader hitting this code can reconstruct the "why" in two hops (code → ADR-126 / design note).

### Summary

**Solid:**
- Scope rightly narrowed; doc makes the limit explicit ("advisory declaration, not dispatcher")
- `fires_on_each_declared_channel` test anchors the drift prevention claim
- Self-containment of `has` closure correct given module boundaries
- No other drift-prone copies of the channel list in the codebase
- ADR linkage clear

**Worth changing before merge:** nothing blocking.

**Nice-to-have follow-ups (separate PRs, explicitly NOT this one):**
- Unify the three string-parse helpers (`frontmatter::fires_on_something` inner `has`, `lint::helpers::has_field`, `scan::candidates::get_fm_field`) if/when a second consumer of `FIRE_BEARING_FIELDS` appears.
- Tighten edge cases (indented top-level fields, commented-out fields, sub-field collisions) across all three helpers together.
- If a future channel lands that's *not* a top-level field, the const-driven model will need an adjustment — worth a sentence in the doc acknowledging that bound.

Ship it.

---

PR #75, opened 2026-04-23
fix(attend): reject --focus to empty groups; drop misleading "(project)" label

aaronsb:
## Summary
- `attend send --focus <name>` used to accept any string, deposit a signal in `@<name>/`, and print "signal written" — even when no peer was scanning that directory. The sender saw a successful send; the message vanished. This change validates `--focus` like `--to` already does: refuses the send and points at the broadcast default if the named group has no peer members.
- `attend peers` rendered "(project)" in the Focus column for self and same-cwd peers. That label never named a real focus group, but it primed an agent into running `attend send --focus project "..."` — the exact failure that triggered this PR. The label is now `(self)` for the self row, empty for same-cwd peers, and the footer count no longer treats the implicit row as a focus group.

## Failure mode this kills
Two peer Claude sessions running `attend`. Sender intent: ask the other agent a question. Sender ran:
```
attend send --focus project "Jovan — Aaron asking: ..."
[attend] signal written (focus, 1 dirs): 2f5153e5-...-1776896713.signal
```
But neither agent was a member of any explicit focus group. The signal sat unread in `@project/`. After this change, the same command returns:
```
error: no focus group named 'project'

active focus groups:
  bosectl-demo — 2 members

drop --focus to broadcast (reaches every peer):
  attend send <message>
```
…and exits 1. The default `attend send "msg"` was already correct; the trap was an optional flag failing silently.

## Test plan
- [x] `cargo test --manifest-path tools/Cargo.toml -p attend -p attend-chat -p sensor-peers` passes (39 + 30 + ... unit tests, no regressions)
- [x] Smoke: `attend send --focus nonexistent "hi"` → exits 1 with helpful error
- [x] Smoke: `attend send --focus self-only "hi"` (after `attend focus on self-only`) → exits 1 with "you are the only member"
- [x] Smoke: `attend send "hi"` (default broadcast) → still writes to `_broadcast/`
- [x] Smoke: `attend peers` → "(self)" on self row, footer reads "0 focus group(s)" when not joined to any

### DIFF
--- a/tools/Cargo.lock
+++ b/tools/Cargo.lock
@@ -219,7 +219,7 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "attend"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "agent-fmt",
  "agent-identity",
--- a/tools/attend/Cargo.toml
+++ b/tools/attend/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "attend"
-version = "0.6.0"
+version = "0.6.1"
 edition = "2021"
 description = "Active awareness module — adaptive sensing and disclosure for Claude Code sessions"
 license = "MIT"
--- a/tools/attend/src/cmd/peers.rs
+++ b/tools/attend/src/cmd/peers.rs
@@ -49,12 +49,17 @@ pub(crate) fn cmd_peers() {
     if !peers.is_empty() {
         t.add(vec!["", "", "", ""]);
         for (peer_cwd, _project, status, ctx) in &peers {
-            // Same-cwd peers used to render "(project)" in the Focus
-            // column; that string is misleading (it is not a focus
-            // group anyone joined) and the cwd basename is already
-            // visible in the Agent column's dim parens. Leave the
-            // Focus cell empty for project-only peers.
-            let focus_label = String::new();
+            // Same-cwd peers previously rendered "(project)" in the
+            // Focus column — a string that primed agents into running
+            // `attend send --focus project`, a group nobody had
+            // joined. "(here)" pairs with the self row's "(self)"
+            // marker, preserves the same-cwd visual scan, and uses
+            // parentheses to make clear it is not a focus group name.
+            let focus_label = if *peer_cwd == cwd {
+                "(here)".to_string()
+            } else {
+                String::new()
+            };
             let peer_id = Identity::for_cwd(peer_cwd, caps);
             let label = render_agent_label(&peer_id, caps);
             t.add_owned(vec![
--- a/tools/attend/src/cmd/peers.rs
+++ b/tools/attend/src/cmd/peers.rs
@@ -27,8 +27,12 @@ pub(crate) fn cmd_peers() {
         .unwrap_or_default();
     let self_id = Identity::for_cwd(&cwd, caps);
     let self_label = render_agent_label(&self_id, caps);
+    // Self-row Focus cell: "(self)" identifies which row IS the runner of
+    // `attend peers`, without using the literal string "project" — that
+    // priming caused agents to reach for `attend send --focus project`
+    // and silently land signals in an empty `@project/` room.
     t.add_owned(vec![
-        "(project)".to_string(),
+        "(self)".to_string(),
         self_label,
         "working".to_string(),
         String::new(),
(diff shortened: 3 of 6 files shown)


aaronsb:
## Code Review — fix(attend): reject --focus to empty groups

Reviewed against the four priorities you flagged. Read-through only; did not re-run cargo.

### 🟡 1. Stale-member liveness gap in `--focus` validation

`send.rs:129` calls `r.all_groups()`, which returns raw `members.len()` straight from `_groups.yaml` with no liveness check. If a peer joined a group, then died without calling `attend focus off` (crash, kill -9, parent shell exit), its session id stays in `members`. The new validation will see `count >= 1`, conclude the room has readers, and write the signal — recreating the exact silent-failure mode this PR exists to kill.

Concretely: agent A runs `attend focus on project`, then `pkill -9 claude`. `_groups.yaml` still lists A in `project`. Agent B runs `attend send --focus project "..."`. Validation passes, signal lands in `@project/`, nobody reads it.

Two options to close this:
1. Call `r.cleanup_stale()` before reading `all_groups()` (heavier — rewrites the file).
2. Filter `members` through `session_alive()` inline when computing `peer_count` (cheap, read-only). Requires exposing the member list, not just the count.

Option 2 is closer to the spirit of validation (don't mutate during a send). Worth a follow-up if not this PR.

This is the meaningful asymmetry with `--to`: that path goes through `PeerSensor::list_peers()`, which is a live process scan. `--focus` reads file state. The PR description claims the new code "mirrors the --to check" — it mirrors the *shape*, but not the *liveness guarantee*.

### 🟢 2. Self-exclusion math is correct

`if self_in_group { count.saturating_sub(1) } else { *count }` correctly identifies the non-self peer count. `all_groups()` returns raw `members.len()` (verified at `groups.rs:152`); `joined_group_names()` filters by `members.contains(&self.session_id)` (verified at `groups.rs:163`). The `saturating_sub` defends against the impossible case where state mutates between the two `load_state()` calls — fine, but the race window is microseconds and only affects error message wording, not behavior.

Pinned-empty groups (`count=0, pinned=true`) correctly fall through to the rejection branch. Good.

### 🟢 3. Missing fuzzy suggestion (minor symmetry gap)

`--to` validation provides "did you mean: \<peer\>?" via `find_closest_peer`. `--focus` doesn't. With short group names (`project` / `projects` / `proj`), a typo suggestion would be equally useful. Not blocking — the active-groups list serves the same purpose at low cardinality.

### 🟡 4. Same-cwd peer distinction is lost in `peers.rs`

You changed both:
- Self row: `(project)` → `(self)` — clearly improves on the misleading label.
- Same-cwd peer row: `(project)` → `""` — drops the visual marker entirely.

The same-cwd distinction is now only visible by eyeballing cwd basenames in the dim Agent-column parens. Before, you could scan the Focus column and immediately see "this peer is in my project." That signal is gone.

The trade-off is correct in principle (the old label primed the silent-failure send), but you could preserve the distinction with a non-misleading marker like `·` or `(here)`. Optional — the basename comparison works.

### 🟢 5. Footer count is consistent

Removing `+ 1` from `focus_count` aligns with the new self-row semantics: `(self)` is not a focus group. Footer now reports the literal number of joined groups. Matches the table's meaning.

### 🟢 Other read-through notes

- Comment quality is solid — the "priming caused agents to reach for `attend send --focus project`" comment is exactly the kind of rationale that survives the next refactor.
- Error message phrasing (singular/plural handling on `member{s}`, broadcast remediation) is clean.
- Moving `let r = get_groups();` out of the dest-dirs block is harmless. Slightly redundant work in the early-exit path; fine.
- Tests: the project has zero tests for `cmd_send`/`cmd_peers`. No need to start here.

### Assessment

The validation logic does what it claims for the case where peers are *currently alive and joined*. The remaining gap (stale members) is worth at least a TODO comment near the `let groups = r.all_groups();` line, since the silent-failure mode you fixed can recur through that path. Otherwise the change is clean: small, well-commented, attacks a real bug, and the label cleanup removes a documented priming trap.

Recommend merging after either (a) closing the stale-member gap or (b) adding a comment acknowledging it as known and bounded.

---

aaronsb:
Addressed both 🟡 findings in commit a108a7b:

**Stale-member liveness gap** — Added `PeerSensor::live_session_ids()` (sensor-peers/lib.rs) and `Groups::members(name)` (groups.rs). The `--focus` validation now intersects yaml membership with the live discover_peers set, mirroring `--to`'s liveness discipline. A peer that joined-and-died can no longer pass the validation through a phantom yaml entry.

**Same-cwd peer marker** — Same-cwd peer rows now render `(here)` in the Focus column. Pairs with the self row's `(self)`, restores the visual scan, and the parentheses make clear it is not a focus group name.

Smoke-tested:
- `attend send --focus nonexistent "..."` → exits 1, no such group
- `attend send --focus self-only "..."` → exits 1, "no live peers (you are the only listener)"
- `attend send "..."` → still broadcasts to `_broadcast/`
- `attend peers` → `(self)` on self row, `(here)` on same-cwd peer rows (visually verified single-agent path; multi-agent path is read-through-clean)

Version bumped to attend v0.6.1 in 8dc7711.

PR #77, opened 2026-04-28
Instance suffix + heartbeat liveness for attend (ADR-129)

aaronsb:
Implements ADR-129 — disambiguates same-cwd Claude Code sessions in attend's identity layer and replaces the stub `session_alive` check with a real heartbeat-based liveness gate.

## Summary

- **ADR-129** drafted, status `Proposed`. Docs: `docs/architecture/system/ADR-129-instance-suffix-and-heartbeat-liveness.md`.
- **`attend-heartbeat`** (new crate): per-session sidecar at `~/.cache/attend/heartbeat/<session-id>` — mtime is the timestamp, no body. Default grace 90s. `flock`-based `SessionLock` enables a duplicate-attend guard at startup.
- **`attend-instances`** (new crate): per-cwd YAML registry at `~/.cache/attend/instances/<encoded-cwd>.yaml`. Session-bound permanent slots, Greek-letter allocator (alpha..omega), numeric fallback past 24, 7-day age-based GC, `flock`-serialized read-modify-write.
- **Render layer** (`attend identity_view`, `attend cmd/peers`, `attend-chat chip`): always emit `<nickname>-<instance>` for claude senders. Same-cwd siblings now show as `Tamsin-alpha` / `Tamsin-beta`.
- **Liveness gate** (`groups::session_alive`, `attend-chat::known_identities`): heartbeat-fresh check replaces the `return true` stub at `groups.rs:347`. Stale agents disappear from the chat legend; message history is retained.
- **Time-based refresh** in attend-chat: a 5s tick state drives re-discovery so peers appear/disappear without a manual reload.
- **`#group` send hardening**: empty-group sends rejected at the chat layer (mirrors PR #75's `attend send --focus` discipline).
- **`@nickname` routing**: forgiving Levenshtein-≤-2 fallback when no exact match exists; ambiguity at min distance returns None.
- **Tab cycle**: successive Tabs cycle through prefix-matching candidates for `@` and `#`. Implicit reset when the buffer's tail no longer matches.
- **`make purge-attend-state`**: recovery target wiping `~/.cache/attend/`. Never auto-invoked. Refuses to run while attend is live. Post-build advisory printed after `make attend(-rebuild)` and `make attend-chat(-rebuild)`.

## Verified live

Two parallel claude sessions in `/home/aaron/.claude` registered as `Tamsin-alpha` (this session) and `Tamsin-beta` (a peer). Confirmed end-to-end:
- `attend peers` from each side renders both with distinct suffixes.
- Heartbeat staleness: killing the peer's attend → its heartbeat ages past 90s → attend-chat reload filters it from the legend (history retained).
- Cross-session messaging: broadcasts and `@Tamsin-α` directed routes both delivered.
- Fuzzy match: live `@Tasmin-alpha` (transposed m/s) routed correctly to alpha.
- Tab cycle: `@T<Tab>` → `@Tamsin-alpha`, `<Tab>` → `@Tamsin-beta`, `<Tab>` → wraps.
- Duplicate guard: a second `attend run` for the same session exits cleanly with the held-lock message.

## Test plan

- [x] Unit tests across affected crates: 39 attend, 138 attend-chat, 14 attend-instances, 8 attend-heartbeat, 30 sensor-peers. All green.
- [x] Live verification of suffix render in `attend peers` (both sides).
- [x] Live verification of heartbeat liveness gate (peer exit → legend filters).
- [x] Live verification of cross-session messaging (broadcast + directed).
- [x] Live verification of fuzzy `@nickname` resolution.
- [x] Live verification of Tab cycle.
- [x] Live verification of duplicate-attend guard.
- [x] `make purge-attend-state` smoke test (refuses while attend is live).
- [ ] Reviewer to verify ADR-129 reads cleanly and the rejected alternatives capture the design conversation.
- [ ] Reviewer to confirm chip.rs split into smaller modules is a reasonable follow-up (file is now 588 lines, over the 500-line review threshold).

## Follow-ups noted in commits

- `chip.rs` (588 lines), `attend-instances/src/lib.rs` (640 lines incl. tests), `attend/src/cmd/run.rs` (636 lines) all over the 500-line review threshold. Natural seams identified for a follow-up split — not bundled here to keep this PR scoped.
- `sensor-peers` peer-message disclosure text still uses `claude/<cwd>` rather than the suffixed nickname. Cosmetic gap; closes when the disclosure path takes a registry lookup.


aaronsb:
## Code review — ADR-129 instance suffix + heartbeat liveness

**Scope read**: ADR + every file in the diff (heartbeat crate, instances crate, run.rs wire-in, identity_view, peers, groups, chip, app, legend, chat groups, sensor-peers cwd disclosure, Makefile). Tests run green locally (39 attend, 138 attend-chat, 14 attend-instances, 8 attend-heartbeat, 30 sensor-peers).

The two-mechanism separation (registry for naming, heartbeat for liveness) is clean, the ADR's independence invariant is preserved at the seams, and the bootstrap-migration / self-reload story for the duplicate-attend lock is correct. One blocker, one perf concern, one cross-cwd ambiguity worth flagging, plus the seams you already noted.

---

### Blocker

**1. `attend-instances::register_with_age` lock is not effective against concurrent registers.**

`tools/attend-instances/src/lib.rs:154-194`. The pattern is `OpenOptions::new().create(true)…open(path)` → `flock(LOCK_EX)` on that FD → read-modify-write → `rename(tmp, path)` → drop FD. flock is keyed on the open-file-description (i.e. the inode), not the path. After a writer renames tmp into place, `path` points to a brand-new inode. The old inode survives unlinked while any FD holds it, but a *new* opener of `path` gets the new inode — and flocks on a different inode do not contend with flocks on the old one.

Race scenario for two concurrent fresh registers in the same cwd (the exact case the ADR cites — "two simultaneous fresh sessions racing for `alpha`"):

1. A opens `path` → INODE_1, flocks INODE_1, reads (empty), writes A.tmp, renames A.tmp → `path` (path now → INODE_2), drops FD. INODE_1 unlinked.
2. B opened `path` → INODE_1 *before* A renamed, blocked on flock; eventually acquires INODE_1's lock. Reads `path` and now sees INODE_2 (A's content with sess-A=alpha) — fine so far.
3. **C arrives** *while B is still in its critical section but after A's rename*. C opens `path` → INODE_2 (different inode from B's INODE_1). C's flock on INODE_2 succeeds immediately because no one holds INODE_2's lock.
4. B and C now both hold valid-looking locks on different inodes, both read the same content (sess-A=alpha), both compute "next free = beta", both write their own .tmp and rename. Whichever rename lands second wins; the other update is silently dropped, and the allocator can hand out the same letter (`beta`) twice.

The ADR text claims "atomic-write plus compare-and-swap; two simultaneous fresh sessions racing for `alpha`: first writer wins; the loser reads-after-conflict and takes `beta`." The implementation has the atomic-write half but not the CAS half. There is no read-after-conflict retry, and the lock does not serialize the way the comment ("concurrent registers on the same cwd serialize cleanly") asserts.

This is the exact same-cwd duplicate-name failure the ADR exists to fix. It will not surface in the current test suite because every test calls `register` sequentially in the same thread.

**Fix options, in order of cleanliness:**

- **Separate sentinel lockfile.** Lock `<encoded-cwd>.yaml.lock` (a file you create once and never rename). flock that sentinel for the entire RMW + rename. Rename the data file freely under the sentinel lock. This is the standard Unix pattern and matches the heartbeat crate's lock-on-a-file-that-isn't-renamed approach (which is correct).
- **Real CAS retry loop.** After write, re-stat + re-read; if the disk content no longer reflects what we just wrote, the rename was clobbered → retry the whole RMW. Bounds + backoff get fiddly.
- **`O_NOFOLLOW` + post-acquire inode check.** Open path, lock it, `fstat` the FD vs `stat(path)`; if inode differs, the file was renamed under us → close, reopen, retry. Fewer files but harder to reason about.

I'd take the sentinel-lockfile path. It's six extra lines and the property becomes obvious.

A regression test in the same vein as `first_lock_acquires_second_returns_none` for heartbeat would help — spawn N threads each calling `register(cwd, format!("sess-{i}"))` and assert each returns a distinct instance. Today's tests would not catch a regression of this kind.

(Note: the heartbeat crate's `try_acquire_session_lock` does NOT have this problem. The heartbeat file is touched, never renamed. The lock there is sound, including on the self-reload exec() path.)

---

### Concerns / open questions

**2. Per-render registry I/O in attend-chat.**

`tools/attend-chat/src/chip.rs::with_instance` (lines 258-263) and `known_identities_with_liveness` both instantiate `Registry::new()` and call `lookup`, which does `fs::read_to_string` + `parse_registry` on the registry file *per chip / per signal*. iocraft re-renders on every key event and on the 5s tick (the new `tick` future). At chat scale today this is negligible, but the cap is 5000 signals — a buffer near the cap with cross-cwd peers means thousands of registry reads + YAML parses per render.

A snapshot per *unique cwd* per render is the natural fix. `Registry::snapshot(&cwd)` already returns the right shape; pre-build a `HashMap<&str, BTreeMap<String, InstanceEntry>>` once at the top of `known_identities_with_liveness` and the `app.rs` chip loop, and let `with_instance` consume that map instead of re-reading each call. Not a blocker — just worth doing before chat traffic gets dense, and it's a contained change in `chip.rs` if you go in for the split below.

**3. Cross-cwd nickname-pool collision (pre-existing, surfaced by ADR-129's promise).**

`Identity::for_cwd(cwd)` is an FNV-1a hash into a finite nickname pool (`tools/agent-identity/src/names.rs`). Two genuinely different cwds can hash to the same nickname stem AND each independently hold `alpha` in their own per-cwd instance registry. The chat then has two `Tamsin-alpha` entries from two different cwds. `chip::resolve_nickname` exact-match returns the first hit (`return Some(hit.cwd.clone())` in the pass-1 `find`), silently misrouting.

ADR-129 fixes the same-cwd case (correctly), but the suffix promise reads as "every render produces a single addressable name" — which is not true across cwd-collision pairs. Worth either a follow-up note in the ADR's "Negative" / "Neutral" section, or a small change to `resolve_nickname` to detect the case and refuse like it does for fuzzy ties (collect all exact matches, return None on ambiguity). The fuzzy path already does the right thing on ties; the exact path doesn't.

This is a pre-existing nickname-pool issue, not regressed by this PR. Flagging because the PR's surface is the right time to surface it.

**4. ADR vs. implementation drift on concurrency semantics.**

The ADR says "atomic-write plus compare-and-swap." The implementation uses flock (with the issue in (1)) and atomic rename — no CAS, no retry. Whatever the fix to (1), align the ADR text with the chosen mechanism so a future reader doesn't have to reverse-engineer the gap.

---

### Things I checked and found sound

- **Self-reload bootstrap migration** for the duplicate-attend lock (`run.rs:144-170`). Old binary held no lock → new code's `try_acquire_session_lock` returns `Ok(Some(_))` on a fresh FD. Reload from a binary that already takes the lock → post-exec child inherits the FD via descriptor inheritance, kernel flock state rides the inherited open-file-description, the `Ok(None) if reloaded` arm correctly treats EWOULDBLOCK as "we already own it." Both bootstrap and steady-state work.
- **Read-time vs write-time GC separation** matches the ADR. `register_with_age` does GC; `touch_at` does not. The tick-loop's periodic `instance_registry.touch(...)` correctly takes the GC-free path. No thundering herd, no oscillation.
- **Liveness gate placement** in `chip::known_identities_with_liveness` covers both the signal-replay branch (line 176) and the seed branch (line 228). Tests pass `|_| true`; production wraps `attend_heartbeat::is_fresh`.
- **Tab-cycle** edge cases are covered: cursor-not-at-end early-returns and resets the cycle (line 229-232); the `expected_tail` check (`<sigil><name> `) only matches when `apply_completion`'s exact form is still the buffer's tail, so an intervening keystroke implicitly breaks the cycle even without an explicit `clear_cycle` call. Switching sigils between Tabs naturally falls out of the tail-mismatch path.
- **Fuzzy nickname tie detection** in `resolve_nickname` is correct — `(dist, cwd, tie_count)` tracking, `tie_count > 1` → None. Two-pass exact-then-fuzzy is right. (The exact path's first-match-wins is the (3) issue, separate from fuzzy.)
- **Empty-group rejection** mirrors `attend send --focus` discipline; base channel correctly bypasses the check.
- **Heartbeat sidecar** API is small and right (mtime is the timestamp, no body, future-dated mtime treated as fresh — the right defensive choice). Tests cover the lock contention and idempotency cases I would have asked for.

---

### Recommended follow-ups (non-blocking, post-merge)

- **`chip.rs` split (currently 686 lines, threshold-busted).** The seams I'd recommend, distinct from the ones you noted: (a) `chip_for` + `color_for` + `truncate` stay (rendering); (b) `KnownIdentity` + `known_identities*` move to a `legend_index` module (registry / liveness gating); (c) `resolve_nickname` + `levenshtein` move to a `routing` module (nickname → cwd dispatch). That's three ~200-line files with one clear responsibility each. The legend-index split also gives the per-render snapshot caching (concern 2) a natural home.
- **`run.rs` tick-loop body extraction (636 lines).** `cmd_run_with_catchup` is mostly setup followed by a `loop { ... }` of mixed responsibilities (heartbeat, self-reload check, sensor poll, governor, checkpoint, instance touch, cleanup). Extract `tick_iteration(...)` so the function decomposes into ~150 lines of setup + a 5-line loop. The seams are visible right now while the diff is fresh.
- **`attend-instances/lib.rs` (640 lines, ~410 implementation + ~230 tests).** Implementation alone is comfortably under threshold. Leave as-is; splitting tests into a separate file doesn't help reviewers.
- **YAML parser duplication.** Three hand-rolled minimal-YAML parsers now (`attend::groups`, `attend-chat::groups`, `attend-instances`). The golden-mirror tests are good drift insurance, but if a fourth shape appears, an ADR for a shared `attend-yaml` micro-crate is overdue. Not now.
- **Test gap noted under (1):** add a multi-thread `register` race regression once the lock fix lands. Without it, the next regression of the same shape will be invisible to CI.
- **Cosmetic:** the `sensor-peers::list_peers` disclosure text `format!("claude/{}", source_cwd)` is the known gap you noted. Out of scope for this PR; cheap to address in a follow-up alongside any other render-site sweep.

aaronsb:
Review feedback addressed in three commits on top of `59fe07d`:

- **`4547498` blocker fix.** `attend-instances::register_with_age` and `touch_at` now lock a sentinel `<cwd>.yaml.lock` file rather than the data file. The data file's atomic rename can no longer split flock state across inodes. Adds `concurrent_registers_assign_distinct_instances` — 12 threads register simultaneously and assert distinct instances + all entries on disk. Catches the race the previous tests missed.

- **`a274afd` per-render caching.** New `attend_instances::SnapshotCache` wraps a `Registry` and a per-cwd memoization map. Built once at the top of every render (and once per Enter/Tab handler), threaded through `chip_for`, `known_identities`, and `known_identities_with_liveness` as `&SnapshotCache`. The yaml is read at most once per distinct cwd per render. Two new tests cover correctness and per-cwd isolation.

- **`29231b7` cross-cwd ambiguity.** `resolve_nickname` exact-match path now counts hits. A single match routes, zero falls through to fuzzy, multiple returns `None`. Mirrors the fuzzy-tie discipline. Two tests: ambiguity refusal and unique-match still routes.

Test counts: 39 attend, 140 attend-chat, 17 attend-instances, 8 attend-heartbeat, 30 sensor-peers — all green.

Follow-ups left for separate PRs (per review):
- `chip.rs` (now larger after the cache plumbing) split into rendering / legend-index / routing modules.
- `run.rs` extract `tick_iteration` for testability.
- ADR text alignment with the sentinel-lockfile mechanism (the text describes "flock on the registry file" — should be edited to match what the code now does).

PR #80, opened 2026-04-30
fix(ways): UserPromptSubmit hook output now reaches agent context

aaronsb:
## Summary

The UserPromptSubmit hook pipeline has been silently dropping all output for the entire history of the system. `ways list` showed prompt-matched ways as "fired," but their bodies never reached the agent's context — markers stamped, content discarded. Only SessionStart and PreToolUse paths actually injected anything.

Two distinct bugs in the same flow:

1. **`scan::prompt` discarded its output entirely.** The match loop called `let _ = crate::cmd::show::way(...)` and never `println!`-ed anything. The marker side effects in `show::way` (engagement state, `way-markers/`, `way-epochs/`) ran successfully, which is why `ways list` reported firings — but stdout was empty.
2. **`scan::state` used the wrong envelope on UserPromptSubmit.** It emitted `{"additionalContext": "..."}` which works on SessionStart and PreToolUse (surfaces as `attachment` entries in the JSONL) but is silently rejected on UserPromptSubmit. The Claude Code hook contract requires `{"hookSpecificOutput": {"hookEventName": "UserPromptSubmit", "additionalContext": "..."}}` for UPS — that shape is wrapped in a discrete system reminder, no separate attachment.

Caught by fingerprint testing: distinctive phrases from `itops/policy.md` (`WRITE_HIGH`, `Multi-party approval`, `Blast Radius`) appeared in the JSONL only at lines from explicit `Read` tool calls, never from any UserPromptSubmit firing.

Reference: https://code.claude.com/docs/en/hooks.md

## Changes

- `scan::prompt` (`tools/ways-cli/src/cmd/scan/mod.rs`) accumulates `show::way` output into a context buffer and emits via `emit_hook_context` using the `hookSpecificOutput` envelope.
- `scan::state` takes a `--hook-event` arg and dispatches between the two envelope shapes — SessionStart firings keep the legacy attachment delivery, UserPromptSubmit firings get the discrete system-reminder delivery.
- `emit_hook_context` helper centralizes the per-event envelope decision so future scan modes only have one place to reach.
- `check-state.sh` extracts `hook_event_name` from hook input JSON and passes `--hook-event` through.
- `ways reset` also clears `/tmp/claude-response-topics-${session_id}` alongside the per-session dir under `sessions_root()`. Previously this side file (written by the Stop hook, read by `check-prompt.sh`) survived reset and biased post-reset matching with stale topics. New `session_side_files()` helper makes it easy to add other off-root state paths if needed.

## Test plan

- [x] `make test-sim` — 10/10 pass (existing scenarios assert markers, not stdout shape)
- [x] Smoke test: `ways scan prompt --query "operation classification destructive..."` emits `{"hookSpecificOutput":{"hookEventName":"UserPromptSubmit","additionalContext":"# Policy Way\n## Operation Classification\n..."}}`
- [x] Smoke test: `ways scan state --hook-event UserPromptSubmit ...` emits the new envelope; `--hook-event SessionStart` (default) keeps the legacy envelope
- [x] Live test in this session after `ways reset --confirm`: way bodies (STRIDE, supply chain tier names, Policy class scheme) demonstrably reach agent context — answers cite config-specific fingerprints that were unreachable before the fix
- [x] `ways reset --session ... --confirm` cleared 210 files (was 209 — the side file is now included)

## Known follow-ups (out of scope for this PR)

- `tools/ways-cli/src/cmd/scan/mod.rs` is at 677 lines, in the Code Quality Way's Review tier (>500). Natural split: extract `state` and its private helpers (`evaluate_context_threshold`, `detect_window_chars`, `transcript_size_since_summary`, `evaluate_file_exists`, `strip_frontmatter`, `capture_show_core`) into `scan/state.rs` (~250 lines moved → mod.rs to ~430).
- Existing session simulator scenarios assert `marker_exists` after `scan_prompt` but never verify non-empty stdout. That's the testing gap that allowed this bug to ship — worth a regression test in a follow-up.
- `make lint` fails in `attend-instances` on a pre-existing doc-comment indentation error (reproduces on `main`). Unrelated; separate cleanup.

### DIFF
--- a/hooks/ways/check-state.sh
+++ b/hooks/ways/check-state.sh
@@ -12,10 +12,11 @@ AGENT_ID=$(echo "$INPUT" | jq -r '.agent_id // empty')
 [[ -n "$AGENT_ID" ]] && export CLAUDE_AGENT_ID="$AGENT_ID"
 TRANSCRIPT=$(echo "$INPUT" | jq -r '.transcript_path // empty')
 PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(echo "$INPUT" | jq -r '.cwd // empty')}"
+HOOK_EVENT=$(echo "$INPUT" | jq -r '.hook_event_name // "SessionStart"')
 
 export CLAUDE_PROJECT_DIR="${PROJECT_DIR}"
 
-ARGS=(--session "$SESSION_ID" --project "$PROJECT_DIR")
+ARGS=(--session "$SESSION_ID" --project "$PROJECT_DIR" --hook-event "$HOOK_EVENT")
 [[ -n "$TRANSCRIPT" ]] && ARGS+=(--transcript "$TRANSCRIPT")
 
 "${HOME}/.claude/bin/ways" scan state "${ARGS[@]}"
--- a/tools/ways-cli/src/cmd/reset.rs
+++ b/tools/ways-cli/src/cmd/reset.rs
@@ -1,10 +1,24 @@
 //! Reset session state — clear markers, epochs, and check fire counts.
 //!
 //! Unjams stale session state without restarting Claude Code.
+//!
+//! Clears both the per-session directory under `sessions_root()` and any
+//! known side files written by hooks to fixed `/tmp` paths (e.g. the
+//! response-topics file written by the Stop hook and consumed by
+//! `check-prompt.sh` to enrich prompt matching).
 
 use anyhow::Result;
+use std::path::PathBuf;
 
 use crate::session;
+
+/// Return paths to off-root, per-session state files that hooks write
+/// outside `sessions_root()`. Reset must clear these alongside the main
+/// session directory or matching can be biased by stale topic context
+/// after a reset.
+fn session_side_files(sid: &str) -> Vec<PathBuf> {
+    vec![PathBuf::from(format!("/tmp/claude-response-topics-{sid}"))]
+}
 
 
 pub fn run(session: Option<&str>, all: bool, confirm: bool) -> Result<()> {
@@ -44,26 +58,42 @@ pub fn run(session: Option<&str>, all: bool, confirm: bool) -> Result<()> {
     for sid in &sessions {
         let dir = format!("{}/{sid}", session::sessions_root());
         let path = std::path::Path::new(&dir);
-        if !path.is_dir() {
+        let side_files: Vec<PathBuf> = session_side_files(sid)
+            .into_iter()
+            .filter(|p| p.exists())
+            .collect();
+
+        let has_main = path.is_dir();
+        if !has_main && side_files.is_empty() {
             continue;
         }
 
-        let count = count_files(path);
+        let main_count = if has_main { count_files(path) } else { 0 };
+        let count = main_count + side_files.len();
         let short_id = &sid[..sid.len().min(12)];
 
         if dry_run {
             println!("Session {short_id}... ({count} state files)");
-            // Show summary of what's in the directory
-            let ways = session::list_fired_ways(sid);
-            if !ways.is_empty() {
-                println!("  ways: {}", ways.len());
+            if has_main {
+                let ways = session::list_fired_ways(sid);
+                if !ways.is_empty() {
+                    println!("  ways: {}", ways.len());
+                }
+                let epoch = session::get_epoch(sid);
+                if epoch > 0 {
+                    println!("  epoch: {epoch}");
+                }
             }
-            let epoch = session::get_epoch(sid);
-            if epoch > 0 {
-                println!("  epoch: {epoch}");
+            if !side_files.is_empty() {
+                println!("  side files: {}", side_files.len());
             }
         } else {
-            let _ = std::fs::remove_dir_all(path);
+            if has_main {
+                let _ = std::fs::remove_dir_all(path);
+            }
+            for f in &side_files {
+                let _ = std::fs::remove_file(f);
+            }
             println!("Session {short_id}...: cleared ({count} state files)");
             total += count;
         }
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -52,6 +52,8 @@ pub fn prompt(query: &str, session_id: &str, project: Option<&str>) -> Result<()
 
     let embed_matches = batch_embed_score(query);
 
+    let mut context = String::new();
+
     for way in &candidates {
         if !session::scope_matches(&way.scope, &scope) {
             continue;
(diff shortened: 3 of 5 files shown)


aaronsb:
## Code Review — PR #80 (self-review)

Self-PR by repo owner; posting as a comment, not a formal review.

**Size**: 81/-20 across 4 files. Focused fix.

### Assessment

The bug analysis is correct and the fix is minimal. The smoke tests in the PR description plus the live-session evidence (way bodies actually reach context after fix; cited fingerprints unreachable before) is solid empirical confirmation. The two bugs share a single hop in the pipeline and are appropriate to bundle.

### Findings

**1. `emit_hook_context` discriminator — fine for current wiring, but the `if hook_event == "UserPromptSubmit"` test is the wrong shape long-term.**

Per the Claude Code hook contract, `hookSpecificOutput` is the canonical envelope and is required by `UserPromptSubmit`, `SessionStart`, `PreToolUse`, `PostToolUse`, and `SubagentStart` (your own `check-post.sh:79-84` and `inject-subagent.sh:145-146` already emit it for PostToolUse and SubagentStart). The simpler top-level `additionalContext` works on `SessionStart` / `PreToolUse` only because those events tolerate it as a legacy attachment shape — it is not the documented contract. So the dispatch is currently inverted: the special case (`hookSpecificOutput`) is the rule; the simple shape is the exception.

Pragmatically the current code is correct — `check-state.sh` is wired to exactly `SessionStart` and `UserPromptSubmit` (verified: `jq -r '.hooks | to_entries[] ...'` against `settings.json`), and the other emit sites in `scan/mod.rs:262-271` and `scan/mod.rs:338-346` carry a `decision: approve` field this helper does not model, so they correctly weren't migrated. But the next person who adds a `--hook-event PreCompact` or wires `check-state.sh` into `Stop` will get the wrong shape and re-encounter exactly this bug.

**Recommended**: invert to allowlist the simple shape (or, better, default to `hookSpecificOutput` and gate the legacy shape behind a known-tolerant set):

```rust
fn emit_hook_context(hook_event: &str, context: &str) {
    // hookSpecificOutput is the documented contract for all events that
    // can inject context. SessionStart / PreToolUse historically also
    // accept the simpler top-level shape; we keep that for backward
    // compatibility with the existing PreToolUse emit sites that also
    // carry `decision`.
    let payload = match hook_event {
        "SessionStart" => serde_json::json!({ "additionalContext": context }),
        _ => serde_json::json!({
            "hookSpecificOutput": {
                "hookEventName": hook_event,
                "additionalContext": context,
            }
        }),
    };
    println!("{payload}");
}
```

This is also more honest to the contract: any new event added to `check-state.sh` works without code change.

**2. `--hook-event` default of `SessionStart` — preserves legacy, but masks misconfiguration.**

If `check-state.sh` is ever invoked without piped JSON (manual debugging, a misconfigured shell), `jq -r '.hook_event_name // "SessionStart"'` falls back to `SessionStart` and so does the CLI default. Two layers of the same fallback means a misrouted hook silently emits the wrong shape and is dropped — exactly the failure mode this PR fixes. Recommend the CLI either require the flag or default to a sentinel that errors on the simple-shape branch. Not blocking, but worth a defensive `eprintln!` warning if `hook_event` ends up falling back twice.

**3. Helper placement — fine where it is.**

`emit_hook_context` is a scan-output concern, not a scoring concern. `scoring.rs` is about embedding score combination. Leaving it in `scan/mod.rs` next to the other scan emit sites is correct. If/when a third caller appears outside `scan/`, hoist to a new `scan/output.rs`.

**4. `scan::prompt` hardcoding `"UserPromptSubmit"` — pragmatic, document it.**

`scan::prompt` is only ever invoked from `check-prompt.sh`, which is wired exclusively to `UserPromptSubmit`. Hardcoding the string is fine but the function is now silently coupled to its caller's event type. Add a doc-comment on `pub fn prompt`: "Caller is assumed to be a `UserPromptSubmit` hook; the output envelope is hardcoded to that shape." If it ever gets reused elsewhere the assumption is loud.

**5. `session_side_files()` hardcoded path — DRY concern is real.**

`/tmp/claude-response-topics-{sid}` now appears in three places: `reset.rs:20`, `check-prompt.sh:18`, `check-response.sh:23`. Recommended follow-up (not blocking this PR): a `RESPONSE_TOPICS_PATH` constant exposed via `ways print-paths --kind response-topics ${SESSION_ID}` or similar, so the shell hooks `source` from a single source of truth. The current PR is internally consistent; the DRY violation predates it.

**6. Regression test gap — should land with this PR.**

The PR description explicitly identifies the test gap that allowed this bug to ship: simulator scenarios assert `marker_exists` after `scan_prompt` but never assert non-empty stdout in the expected envelope shape. This is precisely the gap that let the bug exist for the entire history of the system. Adding a one-line assertion to one existing scenario (parse stdout as JSON, assert `.hookSpecificOutput.additionalContext` is non-empty after a known-firing prompt) is small and prevents regression. **Strong recommend** before merge — the cost is low and the fix is exactly the kind of thing that silently breaks again.

**7. `scan/mod.rs` at 677 lines — follow-up split is the right call.**

This PR is a bug fix; expanding scope to refactor a 677-line file into modules would obscure the fix and make it harder to revert. The PR description names the natural seam (extract `state` + helpers to `scan/state.rs`, drops `mod.rs` to ~430). Land the fix, then split. Do **not** add to `mod.rs` in subsequent PRs without doing the split first — the file is already in the Code Quality Way's Review tier.

### Triage

**Recommended before merge**:
- Add a regression test asserting non-empty `hookSpecificOutput.additionalContext` on a known-firing prompt scenario (point 6).

**Recommended follow-ups (separate PRs)**:
- Invert `emit_hook_context` discriminator to allowlist the legacy shape (point 1) — prevents recurrence when adding new event wirings.
- Hoist `/tmp/claude-response-topics-${sid}` to a single source of truth (point 5).
- Execute the planned `scan/state.rs` split (point 7).

**Observations**:
- `--hook-event` default fallback layering (point 2) — defensive `eprintln!` would help future debugging.
- Doc-comment on `scan::prompt` clarifying caller assumption (point 4).

Nothing blocks merge except point 6, and even that is "strong recommend" rather than "must."

aaronsb:
Soft-block addressed in commit 7685ef7:
- Added regression assertion to scenario_1: `scan_prompt` output must contain `hookSpecificOutput` when a way matches, and a repeated prompt must emit empty (idempotency).
- Doc comment on `scan::prompt` recording the hardcoded UserPromptSubmit assumption.

Follow-ups from the review tracked as a stacked bundle in #81 (discriminator inversion, DRY `/tmp/claude-response-topics-{sid}`, `scan/state.rs` split, defensive stderr trace).

Ready for re-review or merge.

Issue #81, opened 2026-04-30
Follow-ups from PR #80: hook envelope canonicalization, DRY paths, scan/mod.rs split


aaronsb:
Stacked follow-ups from the PR #80 code review (https://github.com/aaronsb/agent-ways/pull/80). Each can be a separate PR; ordered roughly by priority. Filed as a single issue per the request to keep them in one bundle.

## 1. Invert `emit_hook_context` discriminator — make `hookSpecificOutput` canonical

**Why:** Per the Claude Code hook contract (https://code.claude.com/docs/en/hooks.md), `hookSpecificOutput` is the canonical envelope; the simple `{"additionalContext": "..."}` shape works on `SessionStart` and `PreToolUse` only as a legacy tolerance. Other places in this repo (`hooks/ways/check-post.sh`, `hooks/ways/inject-subagent.sh`) already use `hookSpecificOutput`. The current `if hook_event == "UserPromptSubmit"` reverses rule and exception — the next person wiring `Stop` or `PreCompact` will silently re-encounter the same bug PR #80 fixed.

**Shape:** flip `emit_hook_context` so `hookSpecificOutput` is the default branch, and `SessionStart` / `PreToolUse` are the explicit legacy cases. Update `scan::prompt`'s hardcoded call to drop the explicit `"UserPromptSubmit"` argument since it would be the default.

## 2. DRY the `/tmp/claude-response-topics-{sid}` path

**Why:** the path is now hardcoded in three places:
- `tools/ways-cli/src/cmd/reset.rs:20` (the side-files helper)
- `hooks/ways/check-prompt.sh:18` (the consumer)
- `hooks/ways/check-response.sh:23` (the writer)

A divergence between any two of these silently breaks the reset/inject loop.

**Shape:** extract a shared constant or a small helper script that emits the path given a session ID, and have all three call sites resolve through it. Could also live as a `ways response-topics-path <sid>` subcommand if shell access is preferred.

## 3. Split `tools/ways-cli/src/cmd/scan/mod.rs` to address the file-length flag

**Why:** mod.rs is 677 lines, in the Code Quality Way's Review tier (>500). PR #80 noted this as deferred.

[...]

PR #82, opened 2026-04-30
refactor(scan): make hookSpecificOutput the canonical envelope default

aaronsb:
## Summary

Item 1 of #81 — invert `emit_hook_context`'s discriminator so `hookSpecificOutput` is the default envelope and `SessionStart` / `PreToolUse` are the explicit legacy cases.

Per the [Claude Code hook contract](https://code.claude.com/docs/en/hooks.md), `hookSpecificOutput` is canonical for all hook events; the simpler top-level `additionalContext` is a legacy tolerance accepted only on `SessionStart` and `PreToolUse`. The previous `if hook_event == "UserPromptSubmit"` reversed rule and exception — when `Stop` or `PreCompact` get wired later, they would silently re-hit the same bug #80 fixed (output discarded by the consumer because the envelope shape is rejected on those events).

After this change, the canonical envelope is the default branch — `Stop`, `PreCompact`, `SubagentStop`, etc. all get the right shape automatically. Brings `scan::*` in line with `check-post.sh` and `inject-subagent.sh`, which already use `hookSpecificOutput`.

## Changes

- `tools/ways-cli/src/cmd/scan/mod.rs`: `emit_hook_context` now `match`es on the event name with `SessionStart | PreToolUse => legacy`, `_ => canonical`. Docstrings on `emit_hook_context` and `scan::prompt` rewritten to reflect inverted semantics.
- `tools/ways-cli/src/main.rs`: `--hook-event` clap doc-comment rewritten to match. Default value of `"SessionStart"` is intentionally retained — the silent-misroute concern is tracked separately as #81 item 4.
- `tools/ways-cli/tests/session_sim.rs`: scenario_10 gains positive + negative envelope assertions on the SessionStart path (`additionalContext` present at top level, `hookSpecificOutput` absent) to lock in the legacy tolerance after the flip. scenario_1's existing canonical-envelope assertion on `UserPromptSubmit` still passes via the new default branch.

## Out of scope

The `scan/mod.rs` file is at 685 lines (Code Quality Way Review tier). The split is item 3 of #81 and ships separately.

## Test plan

- [x] `make test-sim` — 10/10 pass (scenario_1 covers canonical default; scenario_10 now covers explicit legacy branch)
- [x] Smoke test: `ways scan state --hook-event UserPromptSubmit ...` → `{"hookSpecificOutput":...}`
- [x] Smoke test: `ways scan state --hook-event SessionStart ...` → `{"additionalContext":...}`
- [x] Smoke test: `ways scan state --hook-event PreToolUse ...` → `{"additionalContext":...}` (other legacy event)
- [x] Smoke test: `ways scan state --hook-event Stop ...` would route to canonical (verified envelope branch reached; output empty in this env because no state-trigger ways match `Stop`)

Refs: #81 (item 1)

### DIFF
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -42,10 +42,11 @@ pub(crate) struct WayCandidate {
 /// Match user prompt against ways and emit matched bodies for the agent.
 ///
 /// Wired only from the `UserPromptSubmit` hook (`check-prompt.sh`), so the
-/// envelope is hardcoded. If this is ever reused from another hook event,
-/// take a `hook_event` parameter and route through `emit_hook_context` —
-/// `SessionStart` and `PreToolUse` accept the simpler `additionalContext`
-/// shape, while `UserPromptSubmit` requires `hookSpecificOutput`.
+/// envelope event name is hardcoded. The call routes through the canonical
+/// `hookSpecificOutput` default branch of `emit_hook_context`. If this is
+/// ever reused from another hook event, just pass that event's name —
+/// `SessionStart` and `PreToolUse` are the only events that take the
+/// legacy top-level `additionalContext` envelope.
 pub fn prompt(query: &str, session_id: &str, project: Option<&str>) -> Result<()> {
     let project_dir = project
         .map(|s| s.to_string())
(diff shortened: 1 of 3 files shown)


aaronsb:
## Code Review: approve

Tight, surgical follow-up to PR #80. The discriminator inversion lands cleanly and the test guard genuinely pins the contract.

### What this change accomplishes

The pre-PR-#80 `emit_hook_context` had its match arms backwards: `UserPromptSubmit` was the sole canonical case, every other event fell to the legacy `additionalContext` shape. PR #80 fixed UPS by adding a special case, but left the discriminator philosophy inverted — the *exception* was treated as canonical. This PR aligns the code with the Claude Code hook contract: `hookSpecificOutput` is canonical for all events; `SessionStart` and `PreToolUse` are the only events that accept the legacy top-level `additionalContext` envelope.

### Verification

**Match logic** (`scan/mod.rs:471-481`): `SessionStart | PreToolUse` → legacy, `_` → canonical. Inversion is exact and matches the contract.

**Runtime hookEventName** (`scan/mod.rs:477`): `"hookEventName": hook_event` — no longer hardcoded to `"UserPromptSubmit"`. When `Stop` / `PreCompact` get wired later they self-label correctly.

**Docstrings** — three sites flipped consistently with no rotted UPS-only claims:
- `scan/mod.rs:42-49` (scan::prompt)
- `scan/mod.rs:462-469` (emit_hook_context)
- `main.rs:361-364` (--hook-event clap arg)

**scenario_10 assertions** (`tests/session_sim.rs:498-511`) — positive `additionalContext` + negative `!hookSpecificOutput` together do real discrimination work. Because the canonical envelope nests `additionalContext` *inside* `hookSpecificOutput`, a wrong-branch emission would contain both strings; the negative assertion is what actually pins the legacy branch. The inverse guard for the canonical default lives in scenario_1 (line 258), forming a complete two-sided regression net.

**Scope discipline** — diff is 36/-16 across 3 files. No backwards-compat shims, no dead code, no speculative validation. The eprintln misroute trace correctly deferred to issue #81 item 4.

### Considerations for follow-up

- The two `additionalContext` literals at `scan/mod.rs:276` (PreToolUse `decision: approve`) and `scan/mod.rs:351` (file scan) are intentionally outside `emit_hook_context` because they carry the `decision` field — flagging only so a future "unify all envelope emitters" refactor doesn't lose that distinction.
- No new caller passes a non-UPS, non-SessionStart, non-PreToolUse event yet, so the canonical default is currently exercised only via UPS. When `Stop` / `PreCompact` land, a scenario like scenario_10 but for those events would close the loop — until then the test is a unit-of-discrimination guard rather than an end-to-end one.

### Test status

scenario_1 + scenario_10 pass locally. CI: clippy green, test job (Linux x86_64) green at review time; cross-arch builds in flight but not gating logic.

---

PR #83, opened 2026-04-30
refactor(session): canonicalize response-topics path through one helper

aaronsb:
## Summary

Item 2 of #81 — DRY the `/tmp/claude-response-topics-{sid}` path that was hardcoded in three places.

The path was inlined in:
- `tools/ways-cli/src/cmd/reset.rs:20` — the side-files cleanup helper
- `hooks/ways/check-response.sh:23` — the Stop-hook writer
- `hooks/ways/check-prompt.sh:18` — the UserPromptSubmit consumer

A divergence between any two would silently break the reset/inject loop — reset would leave stale topics behind, or the writer and reader would disagree on where to look.

## Changes

- `tools/ways-cli/src/session.rs`: new `pub fn response_topics_path(session_id: &str) -> PathBuf` — single source of truth, lives next to `sessions_root()` since that's the same module that owns session-state path conventions.
- `tools/ways-cli/src/main.rs`: new `ways response-topics-path <sid>` subcommand exposes the helper for shell hooks. One-line dispatch — `println!` the path and exit.
- `tools/ways-cli/src/cmd/reset.rs`: `session_side_files` now calls `session::response_topics_path` directly.
- `hooks/ways/check-prompt.sh`, `hooks/ways/check-response.sh`: resolve the path via `$("${HOME}/.claude/bin/ways" response-topics-path "$SESSION_ID")`.

The shell scripts already invoke `ways` as their core operation; one extra binary call for the path lookup is negligible compared to the matching/scanning that follows.

## Test plan

- [x] `make test-sim` — 10/10 pass
- [x] Smoke test: `ways response-topics-path test-sid-123` → `/tmp/claude-response-topics-test-sid-123` (matches old hardcoded shape exactly, so existing topics files written before this change remain readable)
- [x] Shell command substitution form `STATE=$(... response-topics-path "$SESSION_ID")` returns the same string that was previously inlined

Refs: #81 (item 2)

### DIFF
--- a/hooks/ways/check-prompt.sh
+++ b/hooks/ways/check-prompt.sh
@@ -14,8 +14,10 @@ PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(echo "$INPUT" | jq -r '.cwd // empty')}"
 AGENT_ID=$(echo "$INPUT" | jq -r '.agent_id // empty')
 [[ -n "$AGENT_ID" ]] && export CLAUDE_AGENT_ID="$AGENT_ID"
 
-# Read response topics from Stop hook (if available)
-RESPONSE_STATE="/tmp/claude-response-topics-${SESSION_ID}"
+# Read response topics from Stop hook (if available).
+# Path resolves through the binary so the writer (check-response.sh),
+# the consumer (here), and `ways reset` cannot drift.
+RESPONSE_STATE=$("${HOME}/.claude/bin/ways" response-topics-path "$SESSION_ID")
 RESPONSE_TOPICS=""
 if [[ -f "$RESPONSE_STATE" ]]; then
   RESPONSE_TOPICS=$(jq -r '.topics // empty' "$RESPONSE_STATE" 2>/dev/null)
--- a/hooks/ways/check-response.sh
+++ b/hooks/ways/check-response.sh
@@ -20,7 +20,9 @@ STOP_ACTIVE=$(echo "$INPUT" | jq -r '.stop_hook_active // false')
 # Need transcript
 [[ ! -f "$TRANSCRIPT" ]] && exit 0
 
-STATE_FILE="/tmp/claude-response-topics-${SESSION_ID}"
+# Path resolves through the binary so this writer, the consumer
+# (check-prompt.sh), and `ways reset` cannot drift.
+STATE_FILE=$("${HOME}/.claude/bin/ways" response-topics-path "$SESSION_ID")
 
 # Extract last assistant message from transcript (JSONL format)
 # Use tail instead of tac to avoid reading entire file
--- a/tools/ways-cli/src/cmd/reset.rs
+++ b/tools/ways-cli/src/cmd/reset.rs
@@ -15,9 +15,10 @@ use crate::session;
 /// Return paths to off-root, per-session state files that hooks write
 /// outside `sessions_root()`. Reset must clear these alongside the main
 /// session directory or matching can be biased by stale topic context
-/// after a reset.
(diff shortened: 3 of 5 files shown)


aaronsb:
## Review — approve

**Verdict**: approve. Tight, single-purpose refactor that does exactly what #81 item 2 asked for. No scope creep, no speculative API surface, all 70 unit + 10 sim tests pass on the branch.

### Byte-identical claim — verified

The PR description claims existing topics files written before this change remain readable. Confirmed by direct comparison against `main`:

| Site | Old literal | New |
|------|------------|-----|
| `tools/ways-cli/src/cmd/reset.rs:20` | `format!("/tmp/claude-response-topics-{sid}")` | `session::response_topics_path(sid)` |
| `hooks/ways/check-prompt.sh:18` | `"/tmp/claude-response-topics-${SESSION_ID}"` | subcommand |
| `hooks/ways/check-response.sh:23` | `"/tmp/claude-response-topics-${SESSION_ID}"` | subcommand |

Helper at `tools/ways-cli/src/session.rs:56-58` produces `/tmp/claude-response-topics-{session_id}` — matches all three former literals exactly. Smoke test confirms: `ways response-topics-path test-sid-123` -> `/tmp/claude-response-topics-test-sid-123`. Even with empty session id (the failure-mode case), old and new both produce `/tmp/claude-response-topics-`.

A `grep -rn claude-response-topics` across the tree shows only the three new call sites + the helper definition. No stragglers.

### Subcommand failure modes

- Empty `SESSION_ID` (jq returned empty): subcommand still exits 0, prints `/tmp/claude-response-topics-`. Identical to the old literal's behavior with empty `${SESSION_ID}`. `[[ -f \"\" ]]` is false in the consumer, so check-prompt safely no-ops; check-response's heredoc redirect to `\"\"` would have failed before and still does. No regression.
- Missing arg: clap prints usage and exits 2. Not reachable from the hooks (they always pass `\"$SESSION_ID\"` even when empty), so harmless.
- Success: exit 0, single-line stdout — clean for `$(...)` capture.

### Shell command-substitution pattern

Neither hook sets `set -e`/`set -u`, so a missing/broken `ways` binary degrades gracefully:
- `check-prompt.sh` is already gated by `require-ways.sh` (line 8), so the binary is guaranteed present before the substitution runs.
- `check-response.sh` is **not** gated by `require-ways.sh`. If the binary is missing, `STATE_FILE=\"\"`, and the heredoc `cat > \"\"` fails silently. This is a minor behavior shift from `main` (where the path was a pure string assignment, which couldn't fail) — but check-response is best-effort instrumentation, not critical path, and the hook would do nothing useful without the binary anyway since check-prompt couldn't read what it wrote. Not worth fixing in this PR; mention only for awareness.

### SOLID / quality

- Single Responsibility: `response_topics_path` does one thing, lives next to `session_dir` / `sessions_root` where path conventions belong.
- DRY: was the entire point — three literals collapsed to one helper.
- File length: `session.rs` 630 -> 640, `main.rs` 611 -> 618. Both already over the 500-line review threshold from the way scan, but this PR is +10 lines for a real DRY win, not the kind of growth that warrants a split-now pushback.
- Doc comment on the helper explains *why* it lives outside `sessions_root()` (predates the hierarchy, hardcoded in shell hooks). Useful for future readers.

### Nit (optional, do not block)

The doc comment on `response_topics_path` says \"the path is hardcoded in shell hooks\" — true at time of writing but no longer literally accurate after this PR (the shell hooks now resolve through this helper). Could be reworded to something like \"... and the path crossed the FFI boundary into shell hooks; canonicalized here so reset, the writer, and the consumer cannot drift.\" Pure docs polish.

PR #84, opened 2026-04-30
chore(scan): warn on stderr when ways scan state defaults --hook-event

aaronsb:
## Summary

Item 4 of #81 — defensive observability for the `--hook-event` fallback.

`ways scan state` had `default_value = "SessionStart"` on its `--hook-event` arg, layered on top of the shell's `jq -r '.hook_event_name // "SessionStart"'` fallback in `check-state.sh`. Two layers of the same default mean a misrouted hook (one whose JSON input has no `hook_event_name`) silently emits the wrong envelope shape — the exact failure PR #80 fixed for the UserPromptSubmit path.

This PR doesn't change behavior — the SessionStart fallback still runs. It just surfaces the misroute on stderr where hook-execution logs will catch it before silent damage accumulates.

## Changes

- `tools/ways-cli/src/main.rs`: `--hook-event` becomes `Option<String>` (no clap default). At the dispatch site, `unwrap_or_else` emits a `[ways]`-tagged stderr trace when `None` and substitutes `"SessionStart"`. Removing the clap default means the fallback is applied in exactly one place — easier to reason about than two layers.
- `tools/ways-cli/tests/session_sim.rs`: new `scan_state_full(hook_event: Option<&str>)` helper returns `(stdout, stderr)`. New scenario_11 asserts (a) the stderr trace surfaces when `--hook-event` is omitted and (b) content delivery is unchanged (SessionStart fallback still fires).

## Test plan

- [x] `make test-sim` — 11/11 pass (scenario_11 added)
- [x] Smoke: `ways scan state --session smoke --project /tmp/nonexistent` emits the `[ways]` notice on stderr
- [x] Smoke: `ways scan state --hook-event SessionStart ...` emits nothing on stderr (no false positives)

Refs: #81 (item 4)

### DIFF
--- a/tools/ways-cli/src/main.rs
+++ b/tools/ways-cli/src/main.rs
@@ -368,9 +368,13 @@ enum ScanCommand {
         /// Hook event that invoked this scan (drives output envelope shape).
         /// `hookSpecificOutput` is canonical for all events; `SessionStart`
         /// and `PreToolUse` are the only events that take the legacy
-        /// top-level `additionalContext` shape. Defaults to SessionStart.
-        #[arg(long, default_value = "SessionStart")]
-        hook_event: String,
+        /// top-level `additionalContext` shape. When omitted, falls back to
+        /// `SessionStart` and emits a stderr trace — `check-state.sh` already
+        /// applies the same fallback via jq, so a missing value here means
+        /// neither layer received `hook_event_name` and the envelope shape
+        /// may be wrong for the actual invoking event.
+        #[arg(long)]
+        hook_event: Option<String>,
     },
 }
 
@@ -529,7 +533,15 @@ fn main() -> Result<()> {
                 cmd::scan::task(&query, &session, project.as_deref(), team.as_deref())
             }
             ScanCommand::State { session, project, transcript, hook_event } => {
-                cmd::scan::state(&session, project.as_deref(), transcript.as_deref(), &hook_event)
+                let event = hook_event.unwrap_or_else(|| {
+                    eprintln!(
+                        "[ways] scan state invoked without --hook-event; defaulting to SessionStart. \
+                         If the invoking hook is not SessionStart/PreToolUse, the output envelope \
+                         shape will be wrong and the agent will silently receive nothing."
+                    );
+                    "SessionStart".to_string()
+                });
+                cmd::scan::state(&session, project.as_deref(), transcript.as_deref(), &event)
             }
         },
         Commands::Show { what } => match what {
--- a/tools/ways-cli/tests/session_sim.rs
+++ b/tools/ways-cli/tests/session_sim.rs
@@ -158,18 +158,33 @@ impl Session {
     }
 
     fn scan_state(&self) -> String {
+        self.scan_state_full(None).0
+    }
+
+    /// Run `ways scan state`, optionally passing `--hook-event`, and return
+    /// `(stdout, stderr)`. Lets tests assert both content delivery and the
+    /// misroute warning channel.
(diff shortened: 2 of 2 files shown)


aaronsb:
## Review Summary — Approve

**Scope check**: 2 files, +62/-10. Exactly the observation-only change item #4 of #81 specified. No scope creep.

### What this changes
Replaces the clap `default_value = \"SessionStart\"` on `--hook-event` with `Option<String>` + `unwrap_or_else` at the dispatch site. When `None`, emits a `[ways]`-tagged stderr trace and substitutes `\"SessionStart\"` (preserving prior behavior). Adds `scan_state_full(Option<&str>) -> (stdout, stderr)` test helper and scenario_11 to lock the trace + content-delivery contract.

### Verification

**Behavior preservation** — `main.rs:535-544`. `unwrap_or_else` returns `\"SessionStart\".to_string()`; `&event` is passed to `cmd::scan::state(.., hook_event: &str)`. Downstream value is byte-identical to the previous clap-default path. Only difference is the stderr line. No double-firing, no drift.

**Single fallback site** — Removing `default_value` and moving the fallback into the dispatch arm is the right call. The clap default + jq fallback pair was the original problem (two layers, neither observable). One fallback in one place is easier to reason about, and the trace makes it audible.

**Single `ScanCommand::State` call site** — `grep -rn \"ScanCommand::State\"` confirms `main.rs:535` is the only destructure. The `Option<String>` migration is safe.

**Stderr-emission landscape** — Only other `eprintln` in `cmd/scan/` is at `scoring.rs:83,89` (embedding engine setup warning, unrelated). The new trace at `main.rs:537-541` is the only stderr-emitting site in the scan dispatch path. The `[ways]` prefix matches the hook-execution log convention.

**Trace message shape** — Tagged (`[ways]`), names the missing flag (`--hook-event`), explains the consequence (`envelope shape will be wrong`). Long but appropriate for a defensive trace that should rarely fire.

**Test helper soundness** — `session_sim.rs:167-188`. Uses the same `ways_bin()`, `fixture_home()`, and `XDG_CACHE_HOME` derivation as the existing `scan_state`. `Command::output()` captures stderr to a pipe (does not inherit), so scenario_11's `stderr.contains` assertion is genuine — it would not silently pass on noise from another layer. Arg construction correctly skips the flag when `None`. `scan_state()` cleanly delegates via `.scan_state_full(None).0`, preserving all 9 existing scenario call sites without modification. stdin not piped — matches the prior helper exactly.

**scenario_11 assertions** — Checks both:
1. stderr trace surfaces (`[ways]` and `--hook-event` substrings present)
2. SessionStart fallback still delivers content (`State Trigger Test Way` in stdout)

That's the right pair — proves observability *and* preserved behavior in one test.

### Minor observations (non-blocking)

- The trace message references `SessionStart/PreToolUse` as the events that take the legacy envelope. Consistent with the doc comment on the field (lines 366-370) and `emit_hook_context` at `cmd/scan/mod.rs:470-477`. Coherent.
- scenario_11 doesn't assert the *absence* of the trace when `--hook-event` is supplied. The PR description's smoke test #5 covers that manually; adding a positive-case assertion in the test would be marginal value vs. maintenance. Fine as-is.
- Doc comment on `hook_event` (lines 370-376) is a useful audit trail explaining *why* the fallback is layered — good capture of the institutional knowledge from #80.

### Assessment

Textbook defensive observability change. The misroute condition is preserved (still falls back to SessionStart, agent still receives envelope) but is now audible on stderr where hook-execution logs will catch it. The `unwrap_or_else` pattern is idiomatic, the test captures the right channel, and the scope is disciplined.

No issues found. Recommend merge.

PR #85, opened 2026-04-30
refactor(scan): extract state scan into its own module

aaronsb:
## Summary

Item 3 of #81 — pure structural split. `tools/ways-cli/src/cmd/scan/mod.rs` was 689 lines, in the Code Quality Way's Review tier (>500). PR #80 noted this as deferred follow-up.

## Changes

- New file `tools/ways-cli/src/cmd/scan/state.rs` (218 lines): `pub fn state` and its private helpers (`evaluate_context_threshold`, `detect_window_chars`, `transcript_size_since_summary`, `evaluate_file_exists`, `strip_frontmatter`, `capture_show_core`) moved verbatim. The state path was the natural seam — 200+ lines of state-trigger evaluation sharing no helpers with the prompt / command / file / task paths apart from `emit_hook_context`.
- `tools/ways-cli/src/cmd/scan/mod.rs` (485 lines, was 689): adds `mod state;` + `pub use state::state` so the call site `cmd::scan::state` is unchanged. `emit_hook_context` (still used by `prompt` for the canonical UPS envelope) is re-scoped private → `pub(super)` so `state.rs` can route through it.

## Out of scope

The other scan modes (`prompt`, `command`, `file`, `task`, plus the matching helpers `match_prompt`, `effective_thresholds`, `check_semantic_score`, `regex_matches`) remain in `mod.rs`. Each is under 110 lines and they share matching infrastructure — splitting them further would just trade file count for indirection.

## Test plan

- [x] `make test-sim` — 11/11 pass (scenario_10 + scenario_11 both exercise the relocated `state` function end-to-end)
- [x] `cargo clippy --workspace --all-targets` clean for `ways-cli` (only unrelated `attend-chat` warnings on main)

## File length

| File | Before | After |
|---|---|---|
| `scan/mod.rs` | 689 | 485 |
| `scan/state.rs` | — | 218 |

`mod.rs` falls below the 500-line Review threshold; nothing else in `tools/ways-cli/src/cmd/scan/` is over the threshold.

Refs: #81 (item 3)

### DIFF
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -5,7 +5,9 @@
 
 mod candidates;
 mod scoring;
+mod state;
 pub(crate) use scoring::batch_embed_score;
+pub use state::state;
 
 use anyhow::Result;
 use regex::Regex;
@@ -467,7 +469,7 @@ fn regex_matches(pattern: &str, text: &str) -> bool {
 /// attachment). Defaulting to canonical means new event wirings
 /// (`Stop`, `PreCompact`, ...) get the right shape automatically rather
 /// than silently re-hitting the bug PR #80 fixed.
-fn emit_hook_context(hook_event: &str, context: &str) {
+pub(super) fn emit_hook_context(hook_event: &str, context: &str) {
     let payload = match hook_event {
         "SessionStart" | "PreToolUse" => {
             serde_json::json!({ "additionalContext": context })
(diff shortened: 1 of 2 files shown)


aaronsb:
## Code Review — approve

**Verdict**: approve. Pure structural split, behavior preserved, all integration points verified.

### What I checked

**Byte-identical move (the headline guarantee):**
- Extracted lines 487-689 from pre-split `mod.rs` (`HEAD~1`) and diffed against `state.rs` lines 16-218 (post-imports). `diff` returns empty — zero drift in function bodies, comments, match arms, `unwrap_or_default` calls, or whitespace.

**Cross-module references resolve:**
- `super::candidates::collect_candidates` → `candidates.rs:12 pub(crate)` ✓
- `super::scoring::{capture_show_way, default_project}` → `scoring.rs:159, 169 pub(crate)` ✓
- `super::emit_hook_context` → `mod.rs:472 pub(super)` ✓
- `crate::cmd::show::core` → `show/mod.rs:261 pub fn` ✓
- `crate::session::*` (`detect_scope`, `core_is_shown`, `core_marker_ts`, `clear_core`, `scope_matches`, `sessions_root`, `log_event`) — unchanged from pre-split

**Re-export shape:**
- `main.rs:544` calls `cmd::scan::state(...)` — `pub use state::state` in `mod.rs:10` preserves this exactly. No call-site change needed.

**Import hygiene:**
- `mod.rs` retains `Result`, `Regex`, `PathBuf`, `regex`, and the `candidates`/`scoring` aliases — all still used by the prompt/task/command/file paths and the `WayCandidate` struct.
- `glob` was never an import (it was always called as `glob::glob(...)` path-prefixed); it moved with `evaluate_file_exists` cleanly.
- `state.rs` imports are all used (verified by `cargo clippy --all-targets` on `ways-cli` — zero warnings).

**Scope:** Diff touches exactly two files. Nothing else snuck in.

### One minor suggestion (non-blocking)

**Location**: `tools/ways-cli/src/cmd/scan/mod.rs:472`

The `pub(super)` on `emit_hook_context` works, but it's slightly broader than the minimum. `super` of `scan/mod.rs` is `cmd`, so this exposes the helper to the entire `cmd::*` tree. The strictly-minimum visibility would be `pub(in crate::cmd::scan)` — visible only within the `scan` subtree where it's actually consumed (`mod.rs` itself and the new `state.rs` sibling).

In practice no other `cmd::*` module calls it (verified via grep), so the surface is identical today. `pub(super)` is also the more idiomatic Rust choice and easier to read. Mentioning purely because the PR description called out "minimum required" — if you want to be literal about that, the `pub(in path)` form is tighter. Not worth changing on its own.

### File length outcome

| File | Before | After | Threshold |
|---|---|---|---|
| `scan/mod.rs` | 689 | 485 | 500 (Review) |
| `scan/state.rs` | — | 218 | — |

`mod.rs` clears the Review tier. `state.rs` lands well under. The seam (state-trigger evaluation shares only `emit_hook_context` with the prompt/command/file/task paths) was the right place to cut — no helpers had to follow.

### Considerations for follow-up

- The remaining `mod.rs` still mixes five distinct scan modes (`prompt`, `task`, `command`, `file`) with their shared matching infrastructure (`match_prompt`, `effective_thresholds`, `check_semantic_score`, `regex_matches`, `emit_hook_context`). The PR description's argument for stopping here — "each is under 110 lines and they share matching infrastructure" — is reasonable. Worth revisiting only if `mod.rs` drifts back over 500.
- Item #3 of #81 closes cleanly with this merge.

PR #89, opened 2026-05-11
fix(attend): resolve session UUID in background sessions

aaronsb:
## Summary

Background sessions launched via `claude --bg` or `/bg` (Claude Code v2.1.139's new agent view) couldn't be peers in attend: their attend Monitor would run, heartbeat would tick, signal files would write — but discovery and liveness checks treated them as dead. Tested empirically by spawning a backgrounded session and observing that:

- group membership and heartbeat keyed off different pid-fallback identities
- `attend send --focus` reported "no live peers" even with a fresh heartbeat
- broadcast wrote to a single peer dir (the sender's), never reaching the bg session

Two places in `sensor-peers` relied on the process `comm` field equalling `claude`. Foreground sessions invoke the user-launched wrapper at `~/.local/bin/claude`, so their `comm` is `claude`. Background sessions are forked directly from the versioned binary at `~/.local/share/claude/versions/<ver>`, so `comm` is the version string (e.g. `2.1.139`). Both code paths walked past bg pids and lost the trail.

### Fixes

- **`find_own_session_id`** — drop the comm filter. Read `~/.claude/sessions/*.json` (the pid → session UUID index, written for both kinds of session) into a HashMap, then walk own process ancestry probing each pid against the map. First match wins.
- **`pid_is_claude`** — switch from `ps -o comm` to `ps -o args`. The full command line keeps the binary path, which contains `claude` for both foreground and bg processes regardless of which entry-point they exec'd.

## Verification

Reproduction script: spawn a fresh bg session running attend in a named focus group, observe identity in `_groups.yaml` and `heartbeat/`.

Pre-fix:
- `_groups.yaml` member: `pid-150303` (the agent's bash subprocess pid)
- `heartbeat/`: `pid-151279` (the Monitor's own pid)
- `attend send --focus`: `error: no live peers in focus group`

Post-fix:
- `_groups.yaml` member: `44d42b6f-576b-4a44-9c05-2918778b2522` (real session UUID)
- `heartbeat/`: same UUID — both subcommands agree
- `attend send --focus`: `signal written (focus, 1 dirs)`
- `attend peers` from a foreground session lists the bg session as a live peer

Several long-running peer sessions on the box that had been invisible (Severin, Horatio, Primrose — all background) showed up in `attend peers` immediately after the rebuild, confirming the fix lifts a class-wide blindness, not just the test case.

## Test plan

- [x] `cargo test -p sensor-peers --release` — 30 tests pass, no regressions
- [x] Foreground session regression check — `attend status` from this PR's session still resolves to its real UUID, not a pid fallback
- [x] Bg session round-trip — message sent from foreground, delivered to bg session's focus group dir
- [ ] Manual smoke test by a reviewer with a bg session of their own (optional)

### DIFF
--- a/tools/sensor-peers/src/lib.rs
+++ b/tools/sensor-peers/src/lib.rs
@@ -743,17 +743,23 @@ fn parse_session_file(content: &str) -> Option<SessionFile> {
 }
 
 /// Check if a PID is alive and running a claude process.
+///
+/// Inspects the full command line rather than `comm`. The `comm` field is
+/// the executable basename, which for background sessions is the version
+/// number (e.g. `2.1.139`) because they exec the versioned binary at
+/// `~/.local/share/claude/versions/<ver>` directly. The full argv keeps
+/// the binary's path, which contains `claude` for both foreground and
+/// background sessions.
 fn pid_is_claude(pid: u32) -> bool {
     let output = Command::new("ps")
-        .args(["--no-headers", "-p", &pid.to_string(), "-o", "comm"])
+        .args(["--no-headers", "-p", &pid.to_string(), "-o", "args"])
         .output()
         .ok();
 
     match output {
         Some(out) if out.status.success() => {
-            let comm = String::from_utf8_lossy(&out.stdout);
-            let comm = comm.trim();
-            comm == "claude" || comm.contains("claude")
+            let args = String::from_utf8_lossy(&out.stdout);
+            args.contains("claude")
         }
         _ => false,
     }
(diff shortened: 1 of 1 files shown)


aaronsb:
## Code Review

**Verdict:** Ship it. The empirical diagnosis is strong and the fix is proportional. Nothing blocking.

### Correctness — walk termination

`find_own_session_id` (lib.rs:823-862) terminates cleanly in every degenerate case I traced:
- `ppid==0` or self-parent: `Ok(ppid) if ppid > 0 && ppid != pid` guard → `break`
- missing /proc / ps returns empty: parse fails → `break`
- ps Command fails entirely: `.ok()?` returns `None` early — note this is the one path that differs from `is_own_session` (lib.rs:783-793), which uses `.ok()` + match and breaks. Minor inconsistency, not a bug.
- depth cap 15: ample for the ~7-level bg tree you described. Won't loop.

The probe-then-walk ordering is right: the immediate pid is checked before walking up, so a direct-launch case where own_pid is in the index resolves on iteration 0.

### `pid_is_claude` substring on argv (lib.rs:753-766)

False-positive risk is low *in context*: the only caller is `discover_peers` (lib.rs:439), and `sf.pid` is sourced from a `~/.claude/sessions/*.json` entry — i.e. a pid that was claude when the session file was written. The substring match is a liveness check, not an identity check.

Pid recycling is the realistic failure mode: claude exits, kernel recycles the pid to `vim ~/notes/claude-thoughts.md`, attend treats it as a live peer. The old `comm == \"claude\"` would also be fooled by something named `claude` exactly; the new check widens the window slightly (any argv containing the literal `claude`). In practice the session file gets stale fast and the worst outcome is a transient ghost peer — not worth a stat-by-mtime hardening in this PR. Worth a one-line comment at the call site noting the assumption, maybe.

### Performance

`find_own_session_id` now reads every `sessions/*.json` unconditionally instead of after pinning a claude_pid. On a heavy box with hundreds of archived session files this is more I/O per `attend send/focus/inbox` invocation. Per your note these are short-lived subcommands, not the tick loop, so it's fine — but if `~/.claude/sessions/` ever balloons, this becomes the first thing to revisit. Cheap mitigation if needed later: stop reading once a map hit is found during the ancestry walk (lazy fill), or bound by mtime.

### Defensive style

The `.ok()?` on the ps call (lib.rs:854) silently returns `None` if `ps` itself can't be spawned (binary missing, sandbox blocks exec). That matches the function's contract — `Option<String>` — and callers handle `None` via pid fallback. Not swallowing anything meaningful. The `break` on parse failure is the right call for transient ps weirdness.

Asymmetry with `is_own_session` (which uses `.ok()` + break throughout) is cosmetic. Pick one if you ever touch this file again.

### SOLID / code quality

- Doc comments are excellent — the rationale block at lib.rs:813-822 is exactly what a future-you needs.
- Pid-map construction (lib.rs:824-841) could be extracted to `load_session_pid_index() -> HashMap<u32, String>` for reuse and testability. Not urgent for a 55-line PR.
- `pid_to_session.is_empty()` early return at lib.rs:839 is a nice cheap exit but means a brand-new install with zero session files always returns `None`. Acceptable — attend correctly falls back to `pid-<self>` in that case.

### Tests

Identity resolution being untested is a real gap, but not for *this* PR — the bug existed before and 30 green tests didn't catch it. Worthwhile follow-up sketch:
- `pid_is_claude` against argv strings: exact `claude`, versioned path, foreign argv containing `claude` literal (false-positive boundary), empty output
- `find_own_session_id` with a temp HOME, a fixture session dir, and a fake ancestry — would require factoring the process-walk behind a trait, which is more change than the bug warranted

### Verdict

Solid fix. Diagnosis quality (empirical repro, before/after evidence including unrelated peers reappearing) is well above the bar. Merge and forget.

---

PR #94, opened 2026-05-21
fix(scan): skip embed for custom-agent dispatches

aaronsb:
## Summary

The `PreToolUse:Task` hook (matcher `Task`) still fires for the renamed `Agent` tool in current Claude Code. When the dispatch is to a custom agent (project `.claude/agents/`, user `~/.claude/agents/`, or plugin-provided), Claude emits a long structured delegation prompt that overruns the embedding model's position-embedding table, causing way-embed to SIGABRT in `ggml_compute_forward_get_rows`. The Rust caller failed open so the crash was silent — KDE's crash reporter is what surfaced it.

Custom agents already carry their own constitution in their `.md` file, so injecting ways into them is redundant. Skip the scan entirely when `subagent_type` resolves to a known custom-agent definition; generic dispatches (`general-purpose`, `Explore`, `Plan`, etc.) keep getting the embed-based way matching they were designed for.

## Diagnosis

- Crash signature: `SIGABRT` in `ggml_compute_forward_get_rows` → `ggml_abort`. Position index overruns the model's position-embedding rows.
- Recent crashes (PIDs 88563, 128657, 156817) all had `--query` payloads that were full agent delegation prompts (vue-expert, code-reviewer) — multi-thousand chars.
- 600+ such crashes since 2026-05-06, peaking at 136 in a single day. Silent until KDE's crash reporter started popping.

## Why this is the right discriminator (not a kill-switch)

A kill-switch disables a feature. This fix preserves the feature for the agents that need it (no `.md` constitution) and skips it for the agents that don't (already have a constitution). The mismatch was conceptual: ways-injection was designed for context-less generic subagents, then custom agents arrived and the matcher caught them too.

## Test plan

- [x] Custom agent (`vue-expert`) with 4500-char prompt → 0 way-embed spawns, 0 crashes
- [x] Global custom agent (`code-reviewer`) → 0 way-embed spawns
- [x] Plugin agent (`code-simplifier`) → 0 way-embed spawns
- [x] `general-purpose` → way-embed still spawns as before (existing behavior preserved)
- [x] `bash -n` clean

### DIFF
--- a/hooks/ways/check-task-pre.sh
+++ b/hooks/ways/check-task-pre.sh
@@ -4,6 +4,13 @@
 # Phase 1 of two-phase subagent injection:
 # 1. This script: ways scan task (matches ways, writes stash)
 # 2. SubagentStart: inject-subagent.sh (reads stash, emits content)
+#
+# Skips dispatches to custom-defined agents (project, global, or plugin):
+# their .md file IS the agent's constitution, so ways injection would be
+# redundant — and their delegation prompts are large enough that embedding
+# them against the corpus crashes the embedder on its position-embedding
+# limit. The scan/embed exists for generic subagents that lack such a
+# constitution (general-purpose, Explore, Plan, etc.).
 
 source "$(dirname "$0")/require-ways.sh"
 
@@ -14,8 +21,25 @@ AGENT_ID=$(echo "$INPUT" | jq -r '.agent_id // empty')
 [[ -n "$AGENT_ID" ]] && export CLAUDE_AGENT_ID="$AGENT_ID"
 PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(echo "$INPUT" | jq -r '.cwd // empty')}"
 TEAM_NAME=$(echo "$INPUT" | jq -r '.tool_input.team_name // empty')
+SUBAGENT_TYPE=$(echo "$INPUT" | jq -r '.tool_input.subagent_type // empty')
 
 [[ -z "$TASK_PROMPT" || -z "$SESSION_ID" ]] && exit 0
+
+# Skip custom agents — they have their own .md constitution.
+if [[ -n "$SUBAGENT_TYPE" ]]; then
+  if [[ -f "${PROJECT_DIR}/.claude/agents/${SUBAGENT_TYPE}.md" ]] \
+     || [[ -f "${HOME}/.claude/agents/${SUBAGENT_TYPE}.md" ]]; then
+    exit 0
+  fi
+  # Plugin-provided agents live under .../plugins/<plugin>/agents/.
+  # Glob via bash so we don't pay for a find() walk on every dispatch.
+  shopt -s nullglob
+  plugin_hits=("${HOME}/.claude/plugins/marketplaces/"*"/plugins/"*"/agents/${SUBAGENT_TYPE}.md")
+  shopt -u nullglob
+  if (( ${#plugin_hits[@]} > 0 )); then
+    exit 0
+  fi
+fi
 
 ARGS=(--query "$TASK_PROMPT" --session "$SESSION_ID" --project "$PROJECT_DIR")
 [[ -n "$TEAM_NAME" ]] && ARGS+=(--team "$TEAM_NAME")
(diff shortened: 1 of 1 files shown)


aaronsb:
## Code review — `fix/skip-embed-for-custom-agents`

**Assessment**: Approve with two non-blocking notes. The discriminator is conceptually right, the comment block carries its own weight (explains *why*, not just *what*), and the existing-behavior preservation for built-ins is clean. The two issues below are about glob hygiene with model-emitted input — neither is exploitable in a meaningful way, but both are worth a follow-up.

### Finding 1 — `subagent_type` participates in pathname expansion (suggestion)

**Location**: `hooks/ways/check-task-pre.sh:36`

```bash
plugin_hits=("${HOME}/.claude/plugins/marketplaces/"*"/plugins/"*"/agents/${SUBAGENT_TYPE}.md")
```

`SUBAGENT_TYPE` is interpolated into a word that already contains unquoted `*` segments, so the entire word is subject to pathname expansion — *including* any glob metacharacters that arrived inside `subagent_type`. The variable is also used unguarded in the two `-f` tests above, which are safer (no glob expansion under `[[`), but still allow `..` traversal.

Concrete cases:

| `subagent_type` value | Effect |
|---|---|
| `"*"` | Plugin pattern becomes `.../agents/*.md` → matches every plugin agent.md → unintended skip |
| `"../foo"` | First two `-f` checks resolve to `.claude/foo.md`; if such a file exists, scan is skipped |
| `"[abc]"` | Bracket expression in glob → may match unrelated files |

**Why it matters**: The blast radius is bounded — the only consequence is "ways aren't injected for this dispatch". No code execution, no data leak. But `subagent_type` is model-emitted, which means the *adversary model* (a confused or prompt-injected Claude) could trigger it without needing filesystem write access. Worth hardening because the cost is one line.

**Suggestion**: Reject any value containing `/`, `..`, or glob metacharacters before using it as a path component.

```bash
# After the -n check:
if [[ "$SUBAGENT_TYPE" == *[/\\\*\?\[]* || "$SUBAGENT_TYPE" == .* ]]; then
  # Not a valid agent name — fall through to scan (or exit 0; either is defensible)
  SUBAGENT_TYPE=""
fi
```

Alternatively, use `compgen -G` for the plugin glob which doesn't require toggling `shopt`:

```bash
if compgen -G "${HOME}/.claude/plugins/marketplaces/*/plugins/*/agents/${SUBAGENT_TYPE}.md" > /dev/null; then
  exit 0
fi
```

(Still want the metachar guard either way — `compgen -G` would expand `*` inside `SUBAGENT_TYPE` the same way.)

### Finding 2 — `shopt` save/restore is unconditional (nit)

**Location**: `hooks/ways/check-task-pre.sh:34-36`

```bash
shopt -s nullglob
plugin_hits=(...)
shopt -u nullglob
```

If `require-ways.sh` (sourced at line 8) ever turns `nullglob` on, the unconditional `shopt -u` after the glob clobbers the inherited setting. The script also doesn't `set -e`, so a mid-block error wouldn't leak the option today — but it's the kind of pattern that ages badly if either condition changes.

**Suggestion**: Save and restore.

```bash
prev_nullglob=$(shopt -p nullglob)
shopt -s nullglob
plugin_hits=(...)
eval "$prev_nullglob"
```

Or sidestep the toggle entirely with `compgen -G` as above.

### What looks good

- The block comment at lines 7-13 explains the *conceptual mismatch* (ways-injection was designed for context-less generic subagents; custom agents arrived with their own constitution and got caught by the matcher). This is exactly the rationale a future reader needs and would not be able to reconstruct from `git log` alone.
- Discriminator is the right shape: three lookup paths matching the three locations Claude Code itself searches. False positives require operator-induced filename collisions; false negatives would require Claude Code to introduce a fourth custom-agent location.
- `[[ -n "$SUBAGENT_TYPE" ]]` guard is correct — built-in dispatches (no `subagent_type` in `tool_input`, or empty) fall straight through to the scan, preserving existing behavior for `general-purpose`, `Explore`, `Plan`.
- Comment explains *why* glob-via-bash instead of `find` (per-dispatch cost) — saves a reviewer the "why not find" question.
- Test plan in the PR description covers all three custom-agent surfaces plus the negative case (`general-purpose` still spawns). Solid.

### Considerations for future work

- The two findings above are the same root cause: `subagent_type` is treated as a trusted path component. A small `validate_agent_name()` helper at the top of the script (or in `require-ways.sh`) would centralize this and make the next person who adds an agent-path lookup safe by default.
- If Claude Code ever stabilizes the `Agent` tool name in the matcher (so this hook stops firing for `Agent`), the whole skip block becomes dead code worth deleting. Worth a TODO comment referencing that, or an ADR note if there's broader architectural drift here.

**Verdict**: Ship it. Findings are non-blocking; the fix correctly addresses the SIGABRT and the rationale is durable.

---
*AI-assisted review via Claude (Opus 4.7).*

PR #95, opened 2026-05-21
fix(scan): cap bash embed query at 256 chars

aaronsb:
## Summary

Bash tool inputs can carry kilobyte-scale heredoc bodies, JSON payloads, and multi-line PR descriptions — `gh pr create --body "$(cat <<EOF…)"`, `curl -d '{…}'`, `git commit -m "$(cat <<EOF…)"`. All of those overrun the MiniLM embedding model's position-embedding table and abort way-embed in `ggml_compute_forward_get_rows`. The Rust caller fails open so the crash was silent until KDE's reporter surfaced it.

The signal that distinguishes one bash command from another lives in the program name and first few args — `gh pr create`, `git commit`, `npm install`. Heredoc bodies carry no signal for "what kind of command is this." This caps the command at 256 chars (~60 tokens, safely under the 128-token model limit) before passing to `ways scan command`. Every existing `commands:` regex in `hooks/ways/` matches within 106 chars, so this changes no current matching behavior.

## Sister PR

This is the bash-hook companion to #94 (which discriminates by `subagent_type` for the Task/Agent hook). Same conceptual fix: the embed pipeline needs narrower input than the raw tool field. Custom agents → skip entirely. Bash commands → cap to the semantic prefix.

## Diagnosis

Crashes since 2026-05-06 (~600 cumulative, peaks at 136/day). Failure signature is consistent: `SIGABRT` in `ggml_compute_forward_get_rows`, query is always a long bash command body. The crashes were silent because `run_embed_match` catches non-zero exit and returns `None`. The April 30 fix (4756412 — UserPromptSubmit output routing) made the scan-path's *side effects* visible, but the underlying cost and crashes were already there.

## Test plan

- [x] `ls -la` (short cmd) → way-embed spawns, scan path runs as before
- [x] 4845-char heredoc command (mimics the just-now PID 177965 crash) → 0 crashes, scan path completes
- [x] `bash -n` clean
- [x] All existing `commands:` regex patterns ≤106 chars — no behavior change

### DIFF
--- a/hooks/ways/check-bash-pre.sh
+++ b/hooks/ways/check-bash-pre.sh
@@ -3,8 +3,26 @@
 #
 # The ways binary handles: command pattern matching, semantic scoring,
 # check curve scoring, session state, and content output.
+#
+# The command is truncated to its semantic prefix before being passed to
+# `ways scan command`. Heredoc bodies (`gh pr create --body "$(cat <<EOF…)"`),
+# JSON payloads (`curl -d '{…}'`), and other large argument bodies carry
+# no signal for "what kind of command is this" — the program name and
+# first few args do. The MiniLM embedding models cap at ~128 tokens of
+# position embeddings; queries past that abort the embedder (ggml
+# get_rows out-of-range). 256 chars ≈ 60 tokens, safely under the limit
+# with headroom for the description that gets appended downstream.
+#
+# This truncation feeds both the embed query *and* the regex matcher in
+# the ways binary. Every existing `commands:` pattern under hooks/ways/
+# matches on the program name + first arg (≤106 chars), so cropping at
+# 256 changes no current behavior. Future patterns that need to look
+# past char 256 of a bash command would be misusing this trigger
+# anyway — that signal belongs in `pattern:` against the description.
 
 source "$(dirname "$0")/require-ways.sh"
+
+readonly CMD_QUERY_MAX=256
 
 INPUT=$(cat)
 CMD=$(echo "$INPUT" | jq -r '.tool_input.command // empty')
@@ -14,9 +32,11 @@ AGENT_ID=$(echo "$INPUT" | jq -r '.agent_id // empty')
 [[ -n "$AGENT_ID" ]] && export CLAUDE_AGENT_ID="$AGENT_ID"
 PROJECT_DIR="${CLAUDE_PROJECT_DIR:-$(echo "$INPUT" | jq -r '.cwd // empty')}"
 
+CMD_QUERY="${CMD:0:$CMD_QUERY_MAX}"
+
 export CLAUDE_PROJECT_DIR="${PROJECT_DIR}"
 "${HOME}/.claude/bin/ways" scan command \
-  --command "$CMD" \
+  --command "$CMD_QUERY" \
   --description "$DESC" \
   --session "$SESSION_ID" \
   --project "$PROJECT_DIR"


aaronsb:
## Code Review — PR #95

**Scope**: 21 additions / 1 deletion in `hooks/ways/check-bash-pre.sh`. Focused review.

### Verdict: Approve

Surgical, well-scoped fix. The diagnosis (silent SIGABRT in `ggml_compute_forward_get_rows`, 600+ crashes since May 6, Rust caller fails open) maps cleanly to the chosen mitigation, and the cap is justified by an empirical survey of every existing trigger. Notes below are suggestions/nits, not blockers.

---

### Verified the claims

**1. Both code paths consume the same `--command` arg.** Confirmed in `tools/ways-cli/src/cmd/scan/mod.rs`:
- Line 211: `regex_matches(cmds_pattern, cmd)` — way regex matcher
- Line 236-242: `query_for_checks = format!(\"{} {}\", cmd, description...)` then `batch_embed_score(&query_for_checks)` — embed matcher

So the docstring's claim that the truncation feeds *both* regex and embed is accurate. The single-knob design is right; splitting the truncation per consumer would invite skew.

**2. The 106-char survey holds.** Reproduced locally:
```
grep -rh '^commands:' hooks/ways/ | awk '{print length, \$0}' | sort -rn | head
106 commands: ^(npm|npx|pytest|cargo|go |pip |docker |eslint|prettier|ruff |tsc|webpack|vite|jest|mocha|twine)
 69 commands: npm\ test|yarn\ test|jest|pytest|cargo\ test|go\ test|rspec
 64 commands: npm\ install|yarn\ add|pip\ install|cargo\ add|go\ get
```
Every pattern anchors on program/first-arg; none has any business inspecting heredoc bodies. **256 chars gives ~2.4× headroom over the longest existing pattern** — generous enough that I don't see a near-future case wanting more.

**3. Embed-side budget is sound.** 256 chars ≈ 60 tokens is well below the 128-position-embedding limit, and the Rust caller concatenates `\"{cmd} {description}\"` before sending to way-embed (`scan/mod.rs:236-242`). Descriptions in this repo are short (a sentence). 60 + maybe 20 still fits comfortably under 128 — the headroom claim in the docstring is real.

---

### Suggestions (non-blocking)

**S1 — UTF-8 slicing is fine, but the original concern deserves a one-liner of empirical reassurance.** Tested locally on a UTF-8 locale (default on this machine, and on essentially every modern shell): `${CMD:0:N}` slices by **characters**, not bytes. A 9-byte string `aaa…bbb` (7 codepoints) is reported as length 7 and `${CMD:0:5}` returns `aaa…b` with the `…` preserved as a single 3-byte codepoint. So jq downstream gets well-formed UTF-8.

The only failure mode would be a `LANG=C`/POSIX-locale shell — but the hook is invoked by Claude Code, not by a user shell, and bash commands in this codebase are ASCII-dominant anyway (program names, flags, file paths). No action needed; mentioning here so reviewers know it's been checked.

**S2 — `readonly CMD_QUERY_MAX=256` (line 25) is good practice and has no re-source hazard.** The script doesn't get sourced by other scripts (it's invoked directly by the hook runner per `settings.json`), and even if it were re-sourced, `readonly` failing on the second assignment would print a warning to stderr but not abort the hook (no `set -e`). Worst case is a benign \"readonly variable\" stderr line. Fine to keep.

A minor alternative would be inlining `256` since the constant is used exactly once, but the named constant carries the intent (\"this is the query cap\") and pairs nicely with the docstring's reference to it. Keep as-is.

**S3 — Docstring quality.** The block at lines 7-21 is the strongest part of this PR. It captures:
- *What* is truncated (the command, before both consumers)
- *Why* the signal lives at the prefix (program + first args)
- *Why this number* (60 tokens vs 128-position limit, headroom for description)
- *Why this doesn't change behavior today* (106-char survey)
- *What future-pattern-authors should do instead* (pattern: against description)

That last point — pointing future authors to `pattern:` instead of stretching `commands:` past 256 chars — is the most valuable line. It pre-empts the next person's instinct to bump the cap when a regex doesn't fire on a long command. Good design hygiene.

**S4 — Sister-PR alignment.** PR #94 (still open) handles the Task/Agent path by skipping the scan entirely for custom agents. This PR handles the Bash path by narrowing the input. Two different discriminators, same root cause (model position-embedding overrun, silent because Rust fails open). The two together are a coherent story:
- Custom agents → no scan (they carry their own constitution)
- Bash → scan with truncated prefix (the signal is at the prefix)
- Generic subagents and short commands → unchanged

Suggestion: when both land, consider adding a one-line note in an ADR or design note pointing to this pair as the canonical reference for \"embed input must be bounded.\" That would save the next person from rediscovering it the hard way. (Not blocking — this is an artifact-discipline observation, not a code issue.)

**S5 — No test for the truncation.** `bash -n` passing is fine for a syntax check, but a regression test would be cheap: assemble a > 256-char command, invoke the hook with that input, confirm the `ways scan command` call receives exactly 256 chars. The existing fixtures in `tools/ways-cli/tests/fixtures/` could host a hook integration test, but the hook layer doesn't appear to have a test harness yet, so this would be net-new infra. Reasonable to defer — but worth tracking as \"hook-test-harness\" if crashes like this recur. The test plan in the PR description (\"4845-char heredoc command → 0 crashes\") is solid manual coverage for now.

---

### Things I checked and don't have a concern about

- **Error paths**: no behavior change for the unhappy path. `${CMD:0:256}` on empty `CMD` yields empty string, same as before truncation. `jq -r ... // empty` already handles missing `tool_input.command`.
- **Side effects**: hook still exits silently if `ways` binary is absent (via `require-ways.sh`), still exports `CLAUDE_AGENT_ID` and `CLAUDE_PROJECT_DIR` unchanged.
- **Variable hygiene**: `CMD_QUERY` is a new local; doesn't shadow anything in `require-ways.sh` or the hook environment.
- **Quoting**: `\"$CMD_QUERY\"` is correctly quoted on the `--command` arg (line 39), preventing word splitting even though it's pre-truncated.
- **Precision loss on 256-char one-liners**: trimming a 300-char single-line `find ... -exec ...` to 256 chars affects only the embed similarity score (regex patterns anchor on prefix). The signal degradation is bounded and uniform — acceptable cost for crash elimination. The 106-char survey justifies this.

---

### Summary

This is a textbook example of a fix that earns its line count. The docstring carries more value than the code change itself: it tells the next person *why* this exists, *why* this number, *what* changes (nothing today), and *where* to add future patterns that would otherwise be tempted to stretch the cap. The mitigation is bounded, the cost is bounded, and the cap has a survey-backed justification with comfortable headroom.

Approve. Ship it.

---

PR #98, opened 2026-05-21
feat(scan): sentence-salience reducer for embed inputs (ADR-130)

aaronsb:
## Summary

Implements ADR-130. Replaces the interim character-truncation caps (PRs #94, #95, #96) with a structural fix: a sentence-salience reducer that runs ahead of every `batch_embed_score` call, preserving high-signal prose instead of clamping at an arbitrary char index.

Inputs that fit the embedder's working window pass through unchanged. Long inputs collapse to top-salience sentences within a per-hook token budget. Non-prose inputs (bash commands, JSON blobs, heredoc bodies — anything `split_sentences` returns ≤1 sentence on) degrade to bag-of-words top-K token selection rather than emitting empty.

## Algorithm

```
tokenize input → frequency count → softmax (numerically stable)
                                       ↓
                             per-sentence salience =
                             avg(softmax weights of its tokens)
                                       ↓
                       sort sentences by salience desc
                       take top-N within token budget
                       resort by document position
                                       ↓
                            join → embedder input
```

Bag-of-words fallback for non-prose, hard char-budget fallback when even tokenization yields nothing usable. Panic-free on input-derived values.

## Per-hook budgets

| Hook | Budget (approx tokens) |
|---|---|
| `scan prompt` | 110 |
| `scan task` | 110 |
| `scan command` | 75 |
| `scan file` | 30 |

Sized for ~85% of MiniLM's 128-token position-embedding window. The approximate tokenizer (`max(word_count, char_count/4)`) over-counts vs WordPiece, so real tokens land safely under 128 at all four budgets.

## What stays from the interim PRs

- PR #94's custom-agent discriminator (`subagent_type` check) — stays. Custom agents have their own constitution; they shouldn't reach the embed path at all.
- PRs #95 / #96 char truncation caps in the hook scripts — stay as belt-and-suspenders fallback. The reducer should never emit anything they'd clamp. Removable in a follow-up after observation window.

## Tests

**18 unit tests in `reduce.rs`** — pass/fail behavior:
- passthrough under budget, empty input, prose reduction with topical term preservation
- document order preservation, single-long-sentence fallback, budget enforcement
- stopword filtering, punctuation handling, softmax correctness, sentence splitter
- non-prose fallback (heredoc bodies, repeated-term blobs)
- behavioral tests against the three actual crash payloads from today (vue-expert dispatch, gh pr create heredoc, task-notification blob)

**Existing 84 unit + 11 simulator tests** still pass.

## Recall validation (the merge gate from the briefing)

Production-faithful gate: queries embedded both ways, threshold gating applied (EN ≥ 0.40, multi ≥ 0.55, per ADR-125), top-1 firing way compared. None == None counts as agreement.

| Case | Full fires | Reduced fires | Agree |
|---|---|---|---|
| vue agent dispatch | (none) | (none) | ✓ |
| gh pr create heredoc | (none) | (none) | ✓ |
| task notification | meta/subagents | meta/subagents | ✓ |
| debugging question | (none) | (none) | ✓ |
| schema migration question | softwaredev/delivery/migrations | softwaredev/delivery/migrations | ✓ |
| test structure question | (none) | (none) | ✓ |
| kubectl long args | (none) | (none) | ✓ |
| code review request | (none) | (none) | ✓ |

**Firing-set agreement: 8/8 (100%)** — Gate: ≥ 90%.

Reproducible via `cargo test -p ways --release -- --ignored validate_recall_against_live_corpus --nocapture` (requires the live corpus + way-embed binary).

Of note: 6 of 8 queries correctly produce "no match" with both inputs — the corpus doesn't have a strong-enough match for those topics under production thresholds. The reducer preserves this null result. The two queries that *do* fire produce identical top-1 ways before and after reduction.

## Manual crash repro

Three payloads identical in shape to today's actual SIGABRTs:
- 11,040-char vue-expert agent dispatch → 0 crashes
- 3,387-char gh pr create heredoc → 0 crashes, scan path emits hook context as designed
- 2,864-char task-notification blob → 0 crashes

## Commits

1. `feat(scan): sentence-salience reducer module` — reduce.rs + 14 unit tests
2. `feat(scan): wire reducer into prompt/task/command/file paths` — integration
3. `test(scan): behavioral tests against actual crash repros` — 4 behavioral tests + slice-bounds bug fix in split_sentences
4. `test(scan): production-faithful recall gate + budget tuning` — threshold-aware recall validation, extended stopwords, bumped budgets to ~85% window

## Refs

- Closes the implementation side of ADR-130
- Builds on the interim truncation safety net from PRs #94, #95, #96
- Respects ADR-125 (embedding-only matcher, no lexical tier reintroduced)
- Respects ADR-127 (input *reduction* not *enrichment* — text-density not the axis)

### DIFF
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -9,13 +9,16 @@ mod scoring;
 mod state;
 pub(crate) use scoring::batch_embed_score;
 
-// Per-hook embed-query budgets (approximate tokens, sized for ~75% of
-// MiniLM's 128-token position-embedding window). The reducer passes
-// inputs through unchanged when they already fit; long inputs collapse
-// to top-salience sentences within budget. See ADR-130.
-const BUDGET_PROMPT: usize = 95;
-const BUDGET_TASK: usize = 95;
-const BUDGET_COMMAND: usize = 60;
+// Per-hook embed-query budgets (approximate tokens). MiniLM's window
+// is 128 position embeddings; we budget ~85% of that. The reducer
+// passes inputs through unchanged when they already fit; long inputs
+// collapse to top-salience sentences within budget. The approximate
+// tokenizer here (whitespace + char-budget max) over-counts vs
+// MiniLM's WordPiece, so real tokens land safely under 128 even at
+// the higher budgets. See ADR-130.
+const BUDGET_PROMPT: usize = 110;
+const BUDGET_TASK: usize = 110;
+const BUDGET_COMMAND: usize = 75;
 const BUDGET_FILE: usize = 30;
 pub use state::state;
 
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -8,6 +8,15 @@ mod reduce;
 mod scoring;
 mod state;
 pub(crate) use scoring::batch_embed_score;
+
+// Per-hook embed-query budgets (approximate tokens, sized for ~75% of
+// MiniLM's 128-token position-embedding window). The reducer passes
+// inputs through unchanged when they already fit; long inputs collapse
+// to top-salience sentences within budget. See ADR-130.
+const BUDGET_PROMPT: usize = 95;
+const BUDGET_TASK: usize = 95;
+const BUDGET_COMMAND: usize = 60;
+const BUDGET_FILE: usize = 30;
 pub use state::state;
 
 use anyhow::Result;
@@ -61,7 +70,12 @@ pub fn prompt(query: &str, session_id: &str, project: Option<&str>) -> Result<()
     let scope = session::detect_scope(session_id);
     let candidates = collect_candidates(&project_dir);
 
-    let embed_matches = batch_embed_score(query);
+    // ADR-130: cap embed input to the model's working window via the
+    // sentence-salience reducer. Pattern/keyword matching downstream
+    // still operates on `query` (the full prompt) — only the embed
+    // signal sees the reduced form.
+    let reduced = reduce::reduce_for_embed(query, BUDGET_PROMPT);
+    let embed_matches = batch_embed_score(&reduced);
 
     let mut context = String::new();
 
@@ -113,7 +127,10 @@ pub fn task(
     let is_teammate = team.is_some();
     let candidates = collect_candidates(&project_dir);
 
-    let embed_matches = batch_embed_score(query);
+    // ADR-130: agent delegation prompts are the largest input class in
+    // practice. Reduce to the model's window before embedding.
+    let reduced = reduce::reduce_for_embed(query, BUDGET_TASK);
+    let embed_matches = batch_embed_score(&reduced);
 
     let mut matched: Vec<(String, String)> = Vec::new(); // (way_id, channel)
 
@@ -232,15 +249,19 @@ pub fn command(
         }
     }
 
-    // Check matching: commands regex + semantic scoring
+    // Check matching: commands regex + semantic scoring.
+    // ADR-130: cap embed input. Heredoc bodies (gh pr create --body
+    // "$(cat <<EOF…)"), curl -d JSON payloads, and similar argument-
+    // body bash commands can run kilobytes long. The regex matcher
+    // above already saw the full cmd; only the embed query is reduced.
     let checks = collect_checks(&project_dir);
     let query_for_checks = format!(
         "{} {}",
         cmd,
         description.unwrap_or("")
     );
-
-    let embed_check_matches = batch_embed_score(&query_for_checks);
+    let reduced_for_checks = reduce::reduce_for_embed(&query_for_checks, BUDGET_COMMAND);
+    let embed_check_matches = batch_embed_score(&reduced_for_checks);
 
     for check in &checks {
         if !session::scope_matches(&check.scope, &scope) {
@@ -316,7 +337,10 @@ pub fn file(filepath: &str, session_id: &str, project: Option<&str>) -> Result<(
     }
 
     let checks = collect_checks(&project_dir);
-    let embed_matches = batch_embed_score(filepath);
+    // ADR-130: filepaths are short by nature, but enforce the budget
+    // uniformly across all hook surfaces for consistency.
+    let reduced = reduce::reduce_for_embed(filepath, BUDGET_FILE);
+    let embed_matches = batch_embed_score(&reduced);
 
     for check in &checks {
         if !session::scope_matches(&check.scope, &scope) {
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -4,6 +4,7 @@
 //! scope/precondition gating, parent-threshold lowering, and show (display).
 
 mod candidates;
+mod reduce;
 mod scoring;
 mod state;
 pub(crate) use scoring::batch_embed_score;
--- a/tools/ways-cli/src/cmd/scan/reduce.rs
+++ b/tools/ways-cli/src/cmd/scan/reduce.rs
@@ -77,13 +77,39 @@ pub fn reduce_for_embed(input: &str, budget_tokens: usize) -> String {
 }
 
 fn approx_tokens(s: &str) -> usize {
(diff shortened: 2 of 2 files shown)


aaronsb:
## Review

Took the four commits in order; the story holds together — module first, integration second, behavioral tests with the slice-bounds fix third, threshold-aware recall gate plus budget tuning fourth. The merge gate logic in commit 4 is the right calibration to mirror production semantics.

### Algorithm correctness

Salience scoring matches ADR-130 as described. Walked the math:

- **Softmax stability** (`reduce.rs:152-167`): max-shift correct, the `if sum == 0.0` guard handles the (effectively unreachable) all-`-inf` case without producing NaN — good defensive code.
- **Empty-token denominator** (`reduce.rs:187`): `tokens.len().max(1) as f64` — empty-token sentences score `0/1 = 0.0` not NaN. No divide-by-zero.
- **Tie-breaking under equal salience**: Rust's `sort_by` is stable; ties resolve by enumerate-index (= document order). Then the explicit `sort_by_key(|(i, _)| *i)` after greedy selection guarantees document order in output. Verified intent.
- **UTF-8 safety in `split_sentences`** (`reduce.rs:101-150`): The splitter inspects bytes for ASCII `.!?` and `\n` (all single-byte, can never share a byte with a multi-byte UTF-8 codepoint), then slices `s[start..i+1]`, `s[start..i]`, `s[start..]` — every boundary lands on a single-byte ASCII char or is 0/len. Safe across UTF-8 inputs.
- **Trailing-punctuation-at-EOF**: `is_sentence_end` requires `i + 1 < bytes.len()` AND whitespace at `i+1`. Trailing \`.\` without whitespace falls through to the tail block which trims and pushes whatever's left. No infinite loop, no panic.
- **Slice-bounds fix in commit 3** (`reduce.rs:115-140`): The fix is correct. Pre-fix code did \`start = end + 1; i += 1\` for sentence-end, leaving the trailing whitespace at \`i\` to potentially get re-processed as part of a paragraph break — which set \`start > end\` on the next iteration's slice. Post-fix advances \`start\` and \`i\` together past punctuation+whitespace and gates the slice with \`start <= i+1\`/\`start <= i\`. Tight.

### Hot-path safety

Panic-free contract holds end-to-end. Audited:

- No \`unwrap()\` on input-derived data. The two \`unwrap_or(...)\` calls (lines 194, 242) are on \`partial_cmp\` Option (NaN handling) — Equal is correct fallback.
- No raw-index slicing without bounds (verified above).
- The fallback chain \`sentences → tokens → char-budget\` (\`reduce.rs:209-212, 222-230, 228-229\`) guarantees non-empty output from non-empty input. The single edge to flag is below.

**Suggestion (non-blocking)** — `reduce.rs:222-229`: \`reduce_by_tokens\` with \`budget_tokens == 0\` would produce empty output (\`take(0)\` and \`max_chars = 0\`). In current code this is unreachable (smallest budget constant is \`BUDGET_FILE = 30\`), but if a future caller wired in \`0\` for any reason the non-empty contract would break silently. Defensive option: \`let max_chars = (budget_tokens * CHARS_PER_TOKEN).max(1);\` and \`take(budget_tokens.max(1))\`. Cheap insurance.

### Budget calibration

\`approx_tokens\` heuristic (\`reduce.rs:79-87\`) using \`max(word_count, char_count/4)\`:

- **English prose / code-heavy bash**: overcounts vs WordPiece (claim holds — WordPiece averages ~4 chars/token on English, and the \`max\` with word count rounds up). Headroom comment in mod.rs is accurate.
- **Multilingual concern (CJK)**: CJK text has no spaces → word_count = 1 for an entire paragraph. \`char_count/4\` then drives the estimate, but Japanese/Chinese WordPiece often tokenizes ~1 token per character (or per 2 chars). So for CJK, the heuristic *under*-estimates real tokens. With \`BUDGET_PROMPT = 110\` (= ~440 char-budget for CJK), real WordPiece tokens could land at 220-440 — well past the 128 window. This is a genuine gap, but not a blocker for this PR because (a) the observed crashes were English, (b) the multi model takes over for non-English and the system has the belt-and-suspenders char caps from PRs #95/#96 still in place. **Recommend** filing a follow-up issue to add a CJK-aware char-budget multiplier (e.g., divide char_count by 2 instead of 4 when non-ASCII ratio crosses a threshold) before removing the interim caps.

### Stopword list

The 80-ish word list is small and reasoned. The PR/git scaffold additions (\`pr branch diff merge review body title summary scope...\`) are well-motivated by the dispatch-prompt structure but they DO bleed into legitimate PR/git topic queries. Quick check:

- \"should I rebase or merge this branch?\" → after stopwords: \`[rebase, should]\`. Salience scoring degrades but the embed input is still full sentences (only sentence *selection* uses salience), so the embed sees \`should I rebase or merge this branch?\` verbatim and can match a \"git workflow\" way semantically.
- A way like \`softwaredev/git/pr-review\` won't get *stronger* signal from a PR query because the topical terms are stopworded out of salience — but won't get *weaker* signal either, because PR queries are short and pass through unchanged under budget. The risk surfaces only when a long input has PR/git as its actual topic and we need to *choose* which sentences to keep. Acceptable trade-off given the observed crash payloads.

**Minor**: \`description\` (line 47) and \`summary\` (line 50) are both common topical English words. Worth monitoring for regressions if a future recall query targets ways about \"writing descriptions\" or \"writing summaries\" — these would score near 0 salience.

### Integration

The contract \"regex sees full input, embed sees reduced\" is honored at all four call sites and the inline comments make it explicit:

- \`mod.rs:80-81\` (prompt): reduced fed to \`batch_embed_score\`, \`query\` (unreduced) fed to \`match_prompt\` at line 95. Clean.
- \`mod.rs:135-136\` (task): same shape. Clean.
- \`mod.rs:231-244\` (command): regex against unreduced \`cmd\`. Then \`mod.rs:266-267\` reduces \`query_for_checks\` for embed. Clean.
- \`mod.rs:333\` (file): regex against unreduced \`filepath\`. Then \`mod.rs:345-346\` reduces for embed. Clean.

### Tests

18 unit + 4 behavioral + 1 ignored integration. Coverage is good and the behavioral tests built from real SIGABRT payloads are the right anchor — they're the artifact of \"the bug that started this work.\" \`split_sentences\` has explicit tests for the paragraph-break case that triggered the panic fix.

**Nit (non-blocking)** — \`reduce.rs:543\`: \`use std::process::Command;\` inside \`top1_fires\` is unused (only \`best_match\` actually uses it at line 568). Under \`#[cfg(test)]\` this won't break a build but it'll produce a dead-import warning under \`cargo clippy --tests --all-targets\`. Trivial cleanup.

**Observation (non-blocking)** — \`reduce.rs:198-205\` (greedy fitting): the loop uses \`continue\` not \`break\` when a sentence exceeds remaining budget. That means a high-salience long sentence can be skipped while a lower-salience short sentence later still gets included. This is a divergence from strict top-N-by-salience but it maximizes budget utilization. Not a bug — worth a one-line comment noting the deliberate behavior so the next reader doesn't \"fix\" it to \`break\`.

### ADR alignment

- **ADR-125** (no lexical tier reintroduced): the reducer never touches the corpus side — it only conditions the query before embedding. The matcher (\`match_prompt\` in mod.rs) is unchanged; per-way \`embed_threshold\` gating, parent-boost, dual EN/multi paths all preserved. Contract honored.
- **ADR-127** (input reduction not embedded-representation enrichment): the reducer outputs *fewer* sentences from the source. No vocabulary augmentation, no semantic expansion, no synonym injection — it strictly removes. Contract honored.
- **ADR-130** (this PR): the merge gate (≥ 90% firing-set agreement on the 8-query validation set, run under production threshold semantics) is the gate the ADR specifies and it lands at 8/8 (100%). The ignored test is reproducible and the assertion (\`pct >= 90\`) is the right gate.

### Verdict

**Approve.** No blockers. Two non-blocking items worth a follow-up issue:

1. **CJK token estimation gap** in \`approx_tokens\` — file before removing the interim char-caps from PRs #95/#96.
2. **Dead import cleanup** at \`reduce.rs:543\`.

Optional cleanups (your call): the \`budget_tokens == 0\` defensive guard in \`reduce_by_tokens\`, and a one-line comment at \`reduce.rs:200\` documenting the \`continue\` (not \`break\`) choice.

Nice work taking the time to write the production-faithful recall validation. The threshold-aware mirroring of \`match_prompt\` semantics in \`top1_fires\` is the difference between \"recall test\" and \"recall *gate*.\"

---

PR #101, opened 2026-05-24
chore(settings): roll up accumulated session permission grants

aaronsb:
## Summary
- ~800 lines of accumulated permission grants (Bash, MCP, WebFetch) from session use.
- `settings.json` only — no code or hook changes.
- 10 entries scrubbed: auto-captured `curl -u user:token` commands had API tokens embedded in the permission string. Removed before commit; never pushed to remote.

## Test plan
- [ ] Diff is permission-list-only; no behavior change.
- [ ] JSON validates (`python3 -m json.tool settings.json`).


aaronsb:
Closing in favor of a smaller pattern-based PR. The 842-entry raw accumulation would have undone main's curated 59-entry pattern list. Replacing with: minimal additions of broad wildcards (Bash(curl:*), mcp__server__*, etc.) on top of the existing curation.

PR #106, opened 2026-06-05
fix(ways): make project-local ways fire (Windows-safe embedding + CRLF)

jonmuse9:
## Problem

Project-local ways never fired semantically, and on Windows often did not fire at all. Reported symptom: `ways corpus` embedded **0 project ways** on Windows even with `CLAUDE_PROJECT_DIR` set and `ways lint` clean, so `ways scan --project` matched only global ways — and even a dead-simple `pattern:` keyword trigger didn't fire.

Investigation of the v0.7.0 source (ADR-109's project-scope embedding *is* implemented) found three independent defects.

### Bug A — Windows path resolution
`ways corpus` discovered projects only by lossy-decoding `~/.claude/projects/*` dir names into POSIX paths (`/C/Development/...`) that never exist on Windows → 0 project ways embedded; it never used `CLAUDE_PROJECT_DIR`'s real path.

**Fix:** embed the current project straight from `CLAUDE_PROJECT_DIR` (canonicalized, no decode); the `~/.claude/projects/*` loop now keys off each project's resolved real path.

### Bug B — ID namespace mismatch (all platforms, the showstopper)
The corpus namespaced project ways as `{encoded}/{id}` but `ways scan` built **bare** ids and matched embeddings by exact `id ==`. The namespaces never met, so project-way embeddings were never found — semantic never fired for project ways on any OS. (Keyword still worked *if* the file was collected — see CRLF.)

**Fix:** both sides derive one shared key from the **real** path via `util::encode_project_key` (canonicalize → strip `\?\` → lowercase on Windows → flatten separators/`:` to `-`). Scan looks up embeddings under a new `corpus_id`; the bare `id` still drives session markers, show, and parent-boost. IDs join with `/` on all platforms (`util::path_to_id`), so they never leak backslashes on Windows.

### CRLF intolerance (Windows)
The scan collector (`candidates.rs`) and `resolve_way_file` (`session.rs`) gated on `content.starts_with("---\n")`, silently skipping any way authored with CRLF line endings — so **neither keyword nor semantic** fired. This is likely why the reporter's keyword pattern didn't fire. The corpus side was already line-based/tolerant.

**Fix:** all frontmatter gates route through `util::has_frontmatter` (line-based, CRLF-tolerant). Six call sites unified.

### Bug C — `--ways-dir` clobber footgun
`ways corpus --ways-dir X` re-embedded and overwrote the canonical user corpus (observed 1495 → 1 entries).

**Fix:** added `--output <dir>` for isolated builds, plus a warning when `--ways-dir` would overwrite the canonical corpus.

## Verification

- `cargo clippy --all-targets -D warnings` — clean.
- New `tests/project_ways.rs` (cross-platform, model-free by default): namespaced-id embedding, keyword firing, **CRLF firing**, `--output` non-clobber, footgun warning, and a model-gated semantic leg.
- Semantic leg run with the real model on Windows: a keyword-free query fired the project way via embedding — conclusive proof of the Bug B fix (even through a path containing a space).
- New unit tests for `encode_project_key`, `path_to_id`, `has_frontmatter`, and CRLF frontmatter parsing.
- Pre-existing failures confirmed unrelated (identical on `main`): `memory_seed::parse_matches_freshly_written_seed` (test CRLF/LF) and the Unix-only `session_sim` scenarios.

## Docs

ADR-109 moved **Draft → Accepted** with an implementation note correcting the namespacing decision (real-path key, not the lossy encoded dir name) and documenting the `--output` footgun guard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jonmuse9:
Closing — retargeting to my own fork's main (jonmuse9/agent-ways).

PR #110, opened 2026-06-09
Cross-platform support: Windows + macOS first-class, with CI + regression guard

jonmuse9:
## Summary

Makes `agent-ways` run first-class on **Windows and macOS** alongside Linux, and adds a **regression guard** (portability lint + a Windows/macOS/Linux CI matrix) so OS assumptions can't silently creep back in. All changes preserve existing Linux behavior — they're gated behind OS conditionals (`#[cfg(windows)]` / `uname` cases) or no-op-on-Linux config.

This bundles the accumulated Windows-compat work from a downstream fork. Commits:

- **`fix(windows): full Windows compatibility for make install and ways match`** — process-exec model (Unix `exec` vs Windows spawn+exit), Makefile `SHELL`/symlink-vs-copy/`.exe` handling, atomic rename, subprocess stdio.
- **`fix(attend): restore Windows compat`** — file locking (`flock` vs `share_mode` lockfile), `home_dir()` `USERPROFILE` preference, path-separator and process-inspection (`ps` vs PowerShell) in sensors.
- **`fix(way-embed): quote MODEL_DIR`** — `$HOME` with spaces.
- **`fix(ways): project-local ways fire (Windows-safe embedding + CRLF)`** — CRLF-tolerant frontmatter, path normalization, project-key encoding.
- **`fix: harden agent-ways for cross-platform`** (this session) — see below.

## Hardening + prevention (final commit)

| Area | Change |
|------|--------|
| **Line endings** | `.gitattributes` pins `eol=lf` for scripts/source so Git Bash never gets CRLF scripts (root cause of prior breakage); overrides `autocrlf` on all platforms |
| **Session paths** | Windows `%LOCALAPPDATA%` branch in `sessions_root()` **mirrored** in `sessions-root.sh` (the two compute independently and must agree). Same for `response_topics_path` and the config-update cache (`metrics.rs` + `check-config-updates.sh`), which were silently divergent on Windows |
| **Shell** | Portable temp cascade for `/tmp` markers; standardized shebangs to `#!/usr/bin/env bash`; portable link extraction in `doc-graph.sh` (BSD grep has no `-P`) |
| **settings.json** | Removed leaked `/home/aaron/.claude/**` permission entries |
| **Tests** | `session_sim` now passes on Windows: copy fixtures instead of the `#[cfg(unix)]`-only symlink, and set `USERPROFILE` (which `home_dir()` prefers on Windows) so the binary reads the fixtures rather than the real `~/.claude` |
| **Prevention** | `scripts/check-portability.sh` (CRLF, shebangs, hardcoded-home, **sessions_root binary-vs-shell agreement**) wired into `pre-commit`; new `.github/workflows/portability.yml` runs build + full test suite + lint on **Windows + macOS + Linux** |
| **Hygiene** | Stop tracking generated `tools/way-embed/build/` |

## Verification

- Full CI green on the fork: Clippy, Build ways, Build attend, Build way-embed, and the new **Cross-platform CI (Windows + macOS + Linux)** all pass.
- `cargo test -p ways` on Windows: **118 unit + 6 project_ways + 11 session_sim** pass (session_sim green single-threaded and parallel).
- `cargo clippy -p ways --all-targets -D warnings` clean.
- Merges cleanly into `main` (no conflicts).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aaronsb:
Merged with thanks — this is the kind of contribution that makes agent-ways durable beyond one machine.

**Verified locally on Linux before accepting** (the new `portability.yml` matrix can't run in-repo until it's on `main`, so this closes that gap for now):
- `cargo build` (debug + release): clean
- `cargo test -p ways`: **118 unit + 6 project_ways + 11 session_sim**, all green
- `cargo clippy --all-targets -D warnings`: clean
- `scripts/check-portability.sh`: all four checks pass, including the `sessions_root` binary-vs-shell agreement (`ways sessions-root` == `sessions-root.sh` == `/run/user/1000/claude-sessions`)

A few things that stood out as genuinely well-judged:

- **`.gitattributes eol=lf`** kills the CRLF breakage class at its root using git's own mechanism, instead of scattering normalization through the scripts. The right lever for the actual problem.
- **The `sessions_root` design** is the part worth calling out: a new `ways sessions-root` subcommand as the canonical source of truth, a fast shell mirror in `sessions-root.sh` so the hot PostToolUse path never spawns a subprocess, and a build-time equality check enforcing that the two agree. That's the correct trade-off — speed where it's hot, canonical-binary-as-truth where it's checked. (I reviewed it closely enough to nearly suggest collapsing the mirror into the binary call, before seeing why it's deliberately a mirror. It earns its place.)
- Everything OS-gated with Linux behavior preserved — purely additive.

The lasting value is the **regression guard**: portability lint in pre-commit plus the Windows/macOS/Linux CI matrix mean OS assumptions can't silently creep back in. Thanks for doing the unglamorous cross-platform work *and* the prevention.

PR #117, opened 2026-06-12
feat(scan): way_nearmiss telemetry (ADR-134 task A)

aaronsb:
## What

Implements **ADR-134 Decision 1** — near-miss telemetry. When a way scores close to its firing threshold but does **not** fire, persist that already-computed score as a `way_nearmiss` event in `~/.claude/stats/events.jsonl`. This is the evidence substrate the later tuning passes (`ways tune --cadence` #9, `--precision` #10) will consume. No new embedding is performed — purely persistence of scores the matcher already computed and was throwing away.

This is a **separate initiative from the over-build loop (#7)** — corpus-wide tuning, not the over-build gate. Per the decoupling decision, the over-build gate is a scoreless postcheck predicate and has nothing for the near-miss machinery to attach to; ADR-134 builds on its own merit (the measured 17/47 irrelevant-fire problem).

## How

- **`match_prompt`** now returns `Fired(channel) | NearMiss { scores, thresholds, margin } | NoMatch` instead of `Option<String>`. Firing is **byte-identical** — same channel strings, same order (EN before multi). The only new behavior is a silent telemetry line on the *non*-firing path.
- **Near-miss predicate:** at least one model landed in `thr - margin <= score < thr`. Reported `margin` is the smallest shortfall across qualifying models — how close the way came on its best path. Measured against the **same effective thresholds** the fire path uses, so parent-boost (ADR-125) is honored automatically.
- Pattern/keyword matches are deterministic and scoreless → never near-misses.
- **New config knob** `near_miss_margin` (default `0.05`), parsed alongside the other embed thresholds in `apply_yaml`. A pure logging knob; it cannot change firing.
- **Event shape** mirrors `way_fired`'s field order for reader symmetry: `way`, `corpus_id`, `domain`, `score_en`, `score_multi`, `thr_en`, `thr_multi`, `margin`, `trigger` (`prompt`|`task`), `scope`, `project`, `session`, `query_tokens`.
- Logged from **both** scan paths — `prompt()` and `task()` — distinguished by the `trigger` field.

## Scope boundary

Logging only. No `tune` consumer (#9/#10), no rotation/cap (#12). The events.jsonl growth this stream adds is bounded by task E (#12); a breadcrumb is left at the log site. ADR-134 stays **Draft** until task F (#13) flips it on the final merge.

## Tests

- New `near_miss_tests` module unit-covers the margin decision (fires EN / fires multi / in-band near-miss / smallest-shortfall-wins / beyond-margin no-match / pattern preempts / absent scores) — no embedding subprocess needed.
- Full suite green: `cargo test -p ways` (125 + 6 + 11 pass), `cargo clippy -p ways --all-targets` clean.
- Live release scan runs exit 0; embedding confirmed functional (a way fired semantically). A *deterministic* live near-miss depends on exact cosine values and isn't reliably reproducible from the CLI, so the decision logic is covered by unit tests instead.

## Note

`tools/ways-cli/src/cmd/scan/mod.rs` grew to ~705 lines (was 518; most of the growth is the `cfg(test)` module). It's in the >500 review band — splitting it is a separate refactor, deliberately out of scope here.

Refs ADR-134 Decision 1. Branch `adr-134-near-miss`.

### DIFF
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -77,6 +77,7 @@ pub fn prompt(query: &str, session_id: &str, project: Option<&str>) -> Result<()
 
     let scope = session::detect_scope(session_id);
     let candidates = collect_candidates(&project_dir);
+    let near_miss_margin = crate::config::global().near_miss_margin;
 
     // ADR-130: cap embed input to the model's working window via the
     // sentence-salience reducer. Pattern/keyword matching downstream
@@ -96,20 +97,25 @@ pub fn prompt(query: &str, session_id: &str, project: Option<&str>) -> Result<()
         }
 
         // Additive matching: pattern OR semantic
-        let channel = match_prompt(
+        match match_prompt(
             query,
             &way.pattern,
             &way.corpus_id,
             effective_thresholds(way, session_id),
             &embed_matches,
-        );
-
-        if let Some(trigger) = channel {
-            let out = capture_show_way(&way.id, session_id, &trigger);
-            if !out.is_empty() {
-                context.push_str(&out);
-                context.push_str("\n\n");
+            near_miss_margin,
+        ) {
+            PromptMatch::Fired(trigger) => {
+                let out = capture_show_way(&way.id, session_id, &trigger);
+                if !out.is_empty() {
+                    context.push_str(&out);
+                    context.push_str("\n\n");
+                }
             }
+            PromptMatch::NearMiss(nm) => {
+                log_near_miss(way, &nm, "prompt", &scope, &project_dir, session_id, query);
+            }
+            PromptMatch::NoMatch => {}
         }
     }
 
@@ -134,6 +140,10 @@ pub fn task(
 
     let is_teammate = team.is_some();
     let candidates = collect_candidates(&project_dir);
+    let near_miss_margin = crate::config::global().near_miss_margin;
+    // Session scope for telemetry: the task channel is subagent unless a team
+    // name marks it as a teammate dispatch.
+    let task_scope = if is_teammate { "teammate" } else { "subagent" };
 
     // ADR-130: agent delegation prompts are the largest input class in
     // practice. Reduce to the model's window before embedding.
@@ -162,16 +172,19 @@ pub fn task(
             continue;
         }
 
-        let channel = match_prompt(
+        match match_prompt(
             query,
             &way.pattern,
             &way.corpus_id,
             effective_thresholds(way, session_id),
             &embed_matches,
-        );
-
-        if let Some(trigger) = channel {
-            matched.push((way.id.clone(), trigger));
+            near_miss_margin,
+        ) {
+            PromptMatch::Fired(trigger) => matched.push((way.id.clone(), trigger)),
+            PromptMatch::NearMiss(nm) => {
+                log_near_miss(way, &nm, "task", task_scope, &project_dir, session_id, query);
+            }
+            PromptMatch::NoMatch => {}
         }
     }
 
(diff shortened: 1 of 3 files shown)


aaronsb:
## Review: way_nearmiss telemetry (ADR-134 task A)

Reviewed against the hard requirement (no firing-behavior change, ADR-135 / ADR-134 precision-first) and ADR-134 Decision 1 traceability. Standard-tier review (~235 lines, scan/mod.rs). **Assessment: sound, ships with two minor notes and one worth-confirming item.** No blocking issues.

### No-behavior-change guarantee — VERIFIED

The core requirement holds. Comparing `match_prompt` before/after:

- **Fire path is byte-identical.** Old: `en_fires = best_en >= thr.en`, `multi_fires = best_multi >= thr.multi`, then `if en_fires {en} else if multi_fires {multi} else None`. New: `if best_en.is_some_and(>= thr.en) return Fired(en); if best_multi.is_some_and(>= thr.multi) return Fired(multi);`. Same `>=` comparisons, same EN-before-multi ordering, same channel strings (`semantic:embedding:en` / `semantic:embedding:multi`), pattern checked first and still returns `keyword`. (scan/mod.rs:441-462)
- **Near-miss is unreachable until both fire conditions fail** — it sits after both early returns, so it is structurally incapable of causing or suppressing a fire. (scan/mod.rs:464-490)
- **Callers preserved.** `prompt()` still calls `capture_show_way` + pushes context identically on `Fired`; `task()` still does `matched.push((id, trigger))` on `Fired`. `NearMiss`/`NoMatch` only add a side-channel `log_event` or nothing. (scan/mod.rs:97-119, 175-187)

The only firing-adjacent risk would be a panic on the non-firing path aborting the scan. `log_near_miss` is panic-free: `unwrap_or(&way.id)` on the domain split (always Some anyway), `unwrap_or_default()` on float format, and `log_event` swallows all I/O errors. Safe.

### Correctness of the near-miss predicate — VERIFIED

`thr - margin <= score < thr` is implemented as `gap = thr - s; gap > 0.0 && gap <= margin`. Edge cases trace correctly:
- **score == thr** → `gap == 0.0`, `gap > 0.0` is false → not a near-miss. Correct: it already fired (caught by the `>=` fire check above), so it can never reach here anyway. Belt-and-suspenders.
- **negative gap** (score above thr) → unreachable (fired), and `gap > 0.0` would exclude it regardless.
- **both models in band** → `fold` with `.min()` takes the smallest shortfall. Verified by `smallest_shortfall_wins_across_models` (0.03 vs 0.02 → 0.02).
- **absent scores** → `score.and_then(...)` yields None, `flatten()` drops them, fold over empty → None → `NoMatch`. Covered by `absent_scores_are_no_match`.

The smallest-shortfall semantics ("how close it came on its best path") is the right choice for a recall signal and matches the ADR's framing.

### ADR-134 Decision 1 traceability — VERIFIED

Decision 1 requires: persist already-computed below-threshold scores within a margin as `way_nearmiss`, no new computation. Confirmed — `score_en`/`score_multi` come straight from the `EmbedScores` already computed for the fire check; no new `batch_embed_score` call. Event name, margin default (0.05, matching the ADR's \"e.g. 0.05\"), and the \"persistence not computation\" intent all align. Thresholds used are `effective_thresholds` (parent-boost honored per ADR-125), so the band is measured against the same gate the fire path uses — correct.

---

### Notes (non-blocking)

**1. Field-convention divergence from `way_fired` — confirm intent.** (scan/mod.rs:511-526)
The new event adds `corpus_id`, `score_en`, `score_multi`, `thr_en`, `thr_multi`, `margin`, `query_tokens` — all justified for tuning. But two asymmetries with the canonical fire events are worth a deliberate decision:
- `way_fired` from `show/mod.rs:156-159` attaches `team` (via `detect_team`) when present; `way_nearmiss` does not. For `tune --precision` to correlate near-misses with teammate-scoped sessions symmetrically, you may want the same `team` field. Minor — `scope`/`trigger` already carry teammate-vs-subagent info here.
- The doc comment claims field order \"mirrors `way_fired` (scan/state.rs) for reader symmetry,\" but it doesn't strictly — `way_nearmiss` interleaves the score fields between `domain` and `trigger`, where `way_fired` goes `domain → trigger` directly. The ordering is fine and readable; just consider softening the comment to \"follows the `way_fired` field conventions\" so a future reader doesn't expect literal positional parity.

**2. `scope` vs `trigger` semantics differ between the two callers — confirm this is intended.** (scan/mod.rs:108, 142-143)
- `prompt()` passes `scope = session::detect_scope(session_id)` (yields `\"agent\"` or `\"teammate\"`) and `trigger = \"prompt\"`.
- `task()` passes `scope = task_scope` (`\"subagent\"` or `\"teammate\"`) and `trigger = \"task\"`.

So the `scope` field carries a different value-space depending on which channel logged it (`agent`/`teammate` from prompt, `subagent`/`teammate` from task). That's defensible — it records the actual dispatch context — but `way_fired` uses `detect_scope` uniformly for its `scope` field, so a consumer joining nearmiss and fired streams on `scope` will see `subagent` only on the nearmiss side. Recommend confirming `tune --precision` (task C) keys off `trigger` (the reliable channel discriminator) rather than `scope` for this reason. Not a bug, but a schema seam worth a one-line note in the ADR or a code comment.

### Float formatting — correct
`score_*` use `fmt` → `{s:.4}` with empty-string for absent (so JSON gets `\"\"`, distinguishable from `\"0.0000\"`). `thr_*`/`margin` are always-present `{:.4}`. 4 decimals matches what tuning needs and is consistent within the event. Good.

### Idiom / hot-path — clean
- No allocations on the `NoMatch` path (the common case) beyond the two `Option<f64>` reads already done for firing; the closure-based `shortfall` is zero-cost. No regression to the per-prompt hot path.
- `into_iter().flatten().fold(...)` over a 2-element array is idiomatic and clippy-clean (you report clippy passes).
- `reduce::approx_tokens` visibility bump to `pub(crate)` (reduce.rs:79) is the minimal surface change. Fine.

### Test coverage — adequate for a logging-only change
`near_miss_tests` covers the pure decision arithmetic: en-fires, multi-fires-when-en-below, within-margin shortfall value, smallest-shortfall-across-models, beyond-margin, pattern-preempts-near-miss, absent-scores. That's the right set for the `match_prompt` decision boundary and directly exercises the no-behavior-change seam (the two Fired cases assert exact channel strings).

**Coverage gap (acceptable, but noting):** there is no test asserting the **exact threshold boundary** `score == thr` resolves to `Fired` rather than `NearMiss` — the one input where the `>=` fire check and the `gap > 0.0` near-miss guard must agree. A `run(Some(0.40), None, None) → Fired(en)` case would pin that boundary explicitly and guard against a future refactor flipping `>=` to `>`. Also untested: the end-to-end `log_near_miss` field emission (string formatting, empty-string-for-None), though that's I/O-adjacent and lower-value. Neither blocks.

---

**Considerations for downstream tasks:** the `scope`/`trigger` value-space seam (note 2) and the missing `team` field (note 1) both land in `tune --precision`'s (task C) lap — worth resolving the join key before that consumer is written so the schema doesn't need a migration. Task E (rotation/cap) is correctly referenced in the code comment as the growth bound.

---
*AI-assisted review via Claude Code (code-reviewer). Analysis only — no code modified.*

PR #138, opened 2026-06-20
fix(attend): split the message lane from the event-decay bus (ADR-136 Bug 2)

aaronsb:
Implements **ADR-136** (now Accepted) — the durability half of the attend messaging redesign. Bug 1 (multi-`@` addressing) shipped separately in #137.

## The problem

Authored messages and environmental events shared one pipeline and ran through the **same event-lane noise stack** built to throttle observation noise. The reported bug: a `#open` message reached one peer but not another. Live peer testing revealed it was gated by **three stacked mechanisms** — each fix exposed the next:

1. **Salience gate** (per-signal, mtime-seeded) — suppressed aged-but-unread mail → "peers don't hear."
2. **Action-potential refractory** (per-sensor) — after traffic the peers sensor hit `ABSOLUTE REFRACTORY` and held messages.
3. **Disclosure governor** (shared, rate-ballooning) — starved a coalesced digest for *minutes*. And the in-read 5-min shred let one peer delete a file another hadn't read.

## The fix — a durable, dedup-only message lane

- **Bypass the noise stack** (Decision 1): salience gate *removed* from the message path; refractory *bypassed* (and never built); governor replaced with a **separate permissive governor** (flat 3s, no ballooning) — permissive, not a full bypass.
- **Nothing reaped by age** (Decision 3): the destructive shred is gone; lifetime is bound to **project liveness** like Claude Code (a tray dies when its `~/.claude/projects/` entry does; shared signals reaped by sender project). Age-based cleanup (`--older-than`, retention) removed.
- **Don't overfill on rejoin** (Decision 5): cold-start backlog **baseline** (a fresh join isn't flooded) + a **count-led digest** (a poll surfacing >8 coalesces into one line) + **inbox paging** (`--limit`/`--page`/`--before`) over the never-reaped ledger.
- The neuron-decay model (`sensor_trait`) stays fully intact for the **event lane** (git/process, and a future Slack-as-event sensor).

## Net-simplifying

Per the de-hoard directive, this removes more than it adds: the dead per-signal `SignalSalience` wrapper, the in-read shred, the age-based cleanup machinery (`parse_duration_arg`, `--older-than`, retention), and the re-engagement reset. `salience.rs → signal_id.rs`.

## Validation (live)

Receiver (this repo) ↔ peer (`~/temp`), 2026-06-20: single lane clean ~36s round-trip; permissive governor at 3.0s cooldown (no starvation); refractory bypass confirmed; cold-start no flood; emergent safety — a peer can't be made to flood `#open` by another peer (classifier keys on *who* asks). See ADR-136 §Validation.

## Tests

`cargo build` / `clippy` clean workspace-wide; **508 tests** pass (24 sensor-peers incl. new digest/paging unit tests, 45 attend).

- Suppress the disclosure reheat on a same-hash self-reload (noise; flagged live).
- Config de-hoard residual: `cleanup.retention` + `SignalsConfig` are now parsed-but-unread (harmless; touch the 907-line `config.rs` + lint).

Refs ADR-136 (Decisions 1, 3, 5).


aaronsb:
Follow-ups from review, filed as issues (not blocking this PR):
- #139 — split message scanning into its own sensor (per-observation lane; fixes the peer-presence lane leak)
- #140 — suppress disclosure reheat on same-hash self-reload (noise)
- #141 — de-hoard dead config (cleanup.retention, SignalsConfig)

Issue #160, opened 2026-06-22
Shelve maintainer-maintained i18n ways; replace with an autonomous re-hydrate-and-tune workflow


aaronsb:
## Decision
Shelve the upstream-maintained multilingual ways layer. Stop pre-shipping translations; replace it with a self-service, **autonomous workflow** that lets any adopter re-hydrate and tune all ways into their own language on demand.

## Why
The layer was built so non-English Claude Code users could get ways' benefits — a sound intent. But:

- **Zero validated users.** 222 sessions / 10,019 way fires over ~4.5 months — one English speaker. No non-English usage signal exists.
- **Heavy ongoing cost.** ~1,547 hand-maintained translation lines across 91/119 ways; a second (multilingual) embedding model in every `make setup`; the whole `ways tune` / `language` audit subsystem; a ~17×-per-way authoring tax on every new way.
- **Architecturally leaky, and getting leakier.** Skills are English-only, and ADR-138 (skills=how / ways=5W) plus recent work move value *into* skills. A localized way now hands off to an English skill — we localized the *shrinking* half of the value surface.
- **Unverified substrate.** Claude Code's own non-English skill surfacing/matching has never been measured. We localized ways on top of an unmeasured foundation.

A genuinely complete non-English experience needs localized skills **and** verification of CC's native non-English behavior — a much bigger job that should be triggered by a real adopter, not pre-built against a hypothetical one.

## What "shelve" means (tasks)
- [ ] Make the multilingual embedding model **opt-in** — drop it from the default `make setup` (English installs get lighter).
- [ ] **Drop the per-way localization requirement** — new/edited ways are English-only by default; no 17× tax.
- [ ] Leave the existing ~1,547 stubs / 91 `.locales.jsonl` **dormant in git** (no deletion — fully reversible).
- [ ] Quiet the `ways tune` / locale audit so dormant stubs don't nag (keep the tooling, don't run it by default).
- [ ] Fix the `embed` display so English-only ways aren't buried by the Multi-column sort (it confused our own tuning pass).

## The replacement: re-hydrate-and-tune workflow + guide
Localization becomes adopter-run and autonomous, not maintainer-maintained:
- [ ] A short **guide**: "to run ways in <your language>, run this workflow."
[...]

PR #165, opened 2026-06-22
feat(i18n): shelve maintainer i18n — Phase A (ADR-139)

aaronsb:
Phase A of ADR-139 / #160 — the **subtractive shelve**. Removes the maintainer-maintained localization *data* and *obligations*; the Rust intl engine + multilingual model machinery stay **dormant** (the engine `ways-localize` will drive in Phase B).

## Changes
- **Delete 91 `.locales.jsonl`** (~3,094 entries). git history is the backstop; regenerable per-adopter.
- **`template.rs`** — stop generating locale stubs for new ways (drops the 17× authoring tax); remove now-dead locale helpers + `agents` import.
- **`match_cmd.rs`** — rank `embed`/`match` by the **EN anchor score** so English-only ways aren't buried by a higher multi score. Multi column kept as a dormant diagnostic (shows `—` until localized).
- **`way-embed/Makefile`** — multilingual model (127MB) is **on-demand** via `make model-multilingual`, not default `make setup`. English installs stay light.
- **`Makefile`** — drop `test-lang`/`test-locales`/`test-multilingual` from `make test` (targets kept for manual use).
- **authoring way** — ways are authored English-only; no per-way translation step.
- **`corpus.rs` unchanged** — already degrades to en-only gracefully (the dormant engine).
- **ADR-125 → Accepted** (its retrieval architecture is live; ADR-139 supersedes only its maintainer obligations).

## Verification
- `cargo build --release` clean — **no dead-code warnings** (confirms the helper/import/var removal was complete).
- `ways lint` clean — "0 locale stub files" handled gracefully.
- Full `make test` passes (11 sim scenarios, 0 failures).
- `ways embed`/`match` rank by EN, Multi column dormant `—`; corpus rebuild produces 0 multilingual entries, no nag.

## Scope / follow-ups (Phase B)
- `ways tune` semantics for a single active language (fidelity needs ≥2 variants → discrimination-only) — captured in #160, candidate design-note.
- Rewrite of the tuning way + `docs/hooks-and-ways/` localization pages to describe the adopter-run flow.
- The `ways-localize` skill, re-hydrate workflow, and detection macro.

Part of #160.

https://claude.ai/code/session_01DUyCtuGK5uewAdgp2aE3Dt


aaronsb:
## Code Review — Phase A (ADR-139) subtractive shelve

Reviewed against the ADR's explicit intent: delete localization **data + maintainer obligations**, keep the intl engine **dormant**. Build verified clean (forced rebuild of both edited files: `Finished release profile`, zero dead-code/unused warnings). Findings below, blocking vs. nit separated. Net: solid, surgical work — the two flagged items are dormant-target hygiene, not correctness defects in the live path.

### What's correct (verified, not assumed)

- **`template.rs`** — Removal is complete and surgical. The locale-stub block, `get_covered_languages`, `get_language_name`, the `use crate::agents;` import, and `locales_file` are all gone with no orphans. The symbols those helpers *called* — `agents::is_language_active`, `agents::LANGUAGES_JSON`, `frontmatter::parse_locales_jsonl` — still have live callers in the dormant engine (`tune.rs`, `corpus.rs`, `language.rs`, `util.rs`, `agents/mod.rs`), which is exactly why the build stays warning-free. `vocab` and `Path` are still used by the `.md` template, so no further import pruning is needed. Control flow is intact; the "Next steps" output was correctly renumbered.

- **`match_cmd.rs:55`** — The EN-primary sort key `pair.0.or(pair.1).unwrap_or(f64::NEG_INFINITY)` is correct. It ranks by EN when present (the anchor every way has → English-only ways no longer buried by a higher multi score), falls back to the multi score for a way that *only* has a multi entry (so a locale-only alias still sorts on real signal, not to the bottom), and sinks to `NEG_INFINITY` only when genuinely scoreless. The doc-comment update matches the behavior. No regression for the multi-only case.

- **Dormancy preserved (the ADR's hard line):** `corpus.rs` is untouched and still does the Pass-2 `.locales.jsonl` ingest (`:254`, `:326`, the EN/multi corpus split at `:386`), so reappearing locale data is processed with zero code change. `tools/way-embed/Makefile` keeps `model-multilingual` (`:70`) and `MODEL_MULTI` (`:64`) — only the *unconditional* fetch was dropped from `setup`, exactly per the ADR table. `test-lang`/`test-locales`/`test-multilingual` are kept as targets and only removed from the `test:` aggregate. Engine capability is dormant, not deleted. 

### Blocking — stale instruction that now misdirects a user (live path)

**`tools/ways-cli/src/cmd/corpus.rs:457`**
```rust
log("  Run: make setup  (downloads multilingual model, 127MB)");
```
This is reachable in the **dormant-engine live path**: it fires whenever an adopter regenerates a corpus that contains multilingual entries but lacks the model (`multi_count > 0 && !multi_model.is_file()` at `:455`). After this PR, `make setup` *no longer* downloads the multilingual model — the message now sends the very adopter who is reactivating localization to a command that won't fetch it. The correct target is `make model-multilingual`. This is the exact reactivation path Phase B depends on, so the wrong breadcrumb here is the one that matters. One-line fix:
```rust
log("  Run: make model-multilingual  (downloads multilingual model, 127MB)");
```

### Nit (warn, non-blocking) — dormant test targets now fail when run manually

The PR keeps `test-lang`/`test-multilingual` "for manual use," but with the data deleted they no longer degrade gracefully:

- **`Makefile:253` `test-lang`** asserts `len(active) > 0` on `ways language --json` → hard-fails now (0 active locales).
- **`Makefile:266` `test-multilingual`** route-checks prompts against deleted locale stubs → fails.
- `test-locales` is fine — it vacuously PASSes (verified: `Checked 0 files ... PASS`), though it still prints "17 active languages," which reads oddly against 0 coverage.

CI is unaffected (all three are out of `make test`). But `tests/run-all.sh:60` still unconditionally invokes `test-multilingual.sh`, so anyone running the documented `tests/run-all.sh` gets a red suite. Not blocking for Phase A, but worth a one-line guard (skip when no `.locales.jsonl` exist) or an explicit note that these are reactivation-gated. If you'd rather fold the run-all.sh guard into Phase B with the rest of the reactivation tooling, that's a defensible call — just flagging it so it's a decision, not an oversight.

### Out of scope (correctly deferred, not flagged as missing)

`docs/hooks-and-ways/matching.md:172` still says `make setup` fetches both models — stale, but the PR explicitly defers `docs/hooks-and-ways/` localization pages to Phase B, so this is in-scope-for-B, not a Phase-A miss. Same for `ways tune` single-language semantics and the tuning-way rewrite. ADR-125→Accepted and the INDEX.md bump are consistent with the ADR's `supersedes_in_part` framing.

### Verdict

One blocking line (`corpus.rs:457` wrong make target) and one non-blocking dormant-test nit. Fix the breadcrumb, decide on the run-all.sh guard, and Phase A is clean and faithful to ADR-139's keep-the-engine-dormant intent.

---

PR #182, opened 2026-06-22
feat(commands): add /sync-to-home command for subdirectory-clone installs

jonmuse9:
## Summary

- Adds a `/sync-to-home` Claude command for users who clone agent-ways into a subdirectory of `~/.claude` rather than using it as `~/.claude` itself
- Syncs skills, agents, commands, hooks, and binaries — skips copies when the destination is already symlinked to source
- Merges the hooks block and ways permissions into `settings.json` without touching the user's model/theme/plugins/credentials
- Quotes hook command paths for `$HOME` values containing spaces (Windows compatibility)
- Adds `docs/reference/ways-cli.md` — reference documentation for the ways CLI

## Use case

The canonical install clones agent-ways directly as `~/.claude`. This command supports the other topology: the repo lives at e.g. `~/.claude/directory` while `~/.claude` is an existing config dir with the user's own sessions and credentials.

## Test plan

- [ ] Run `/sync-to-home` from a subdirectory clone — verify skills, agents, commands, hooks land in `~/.claude`
- [ ] Run a second time — verify symlinked hooks/bins are skipped cleanly
- [ ] Verify `settings.json` hooks block is updated and existing keys are preserved

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ curl -sL https://raw.githubusercontent.com/aaronsb/agent-ways/main/scripts/insta
 
 The built-in ways cover software development, but the framework is domain-agnostic. Fork it, replace the ways, add your own domains.
 
-> **Already have `~/.claude/`?** The installer detects existing files and won't clobber them. See the **[install guide](docs/install-guide.md)** for how to back up, merge, or start fresh. If you're sure: `scripts/install.sh --dangerously-clobber`.
+> **Already have `~/.claude/`?** The installer detects existing files and won't clobber them. See the **[install guide](docs/install-guide.md)** for options — including a subdirectory install that keeps your existing config intact and syncs via `/sync-to-home`. If you're sure you want to replace: `scripts/install.sh --dangerously-clobber`.
 
 > **Stop and read this** if you're letting an AI agent run the installer. You are about to let an agent modify `~/.claude/` — the directory that controls how Claude Code behaves. The agent is editing its own configuration. Review the repo first. You are responsible for the result.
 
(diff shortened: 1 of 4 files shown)


aaronsb:
Thanks for this — the subdirectory-clone topology is exactly the gap we keep hitting as people adopt this on machines with an existing `~/.claude` they don't want to turn into a git repo. The command is careful in all the right places: the `$SRC == $DEST` guard, the timestamped backup before mutating, symlink-aware skips on re-run, and reusing `make update-binaries`. The `ways-cli.md` reference is great too — I checked it against the binary and it's accurate (subcommand list, `tune` flags, the `disable` → `.claude/ways.yaml` ADR-131 detail all line up).

I took you up on the offer of a maintainer edit and pushed one fix (76061bb) so you don't have to round-trip on it:

**Step 6 jq didn't compile.** The merge filter had `))) | from_entries)` — `from_entries` ended up *inside* the `map()` and the `.hooks |= ( … )` wrapper never closed, so jq bailed with a compile error. Because the step is gated by `&& mv`, it failed *safe* (your settings.json was preserved, not corrupted) — but the hooks block and ways permissions never merged in, so a synced install would copy every file yet never actually activate the framework. Fix is just paren balance: `)))) | from_entries` then close the wrapper.

I verified the corrected version against the real settings.json: compiles, valid JSON, merges all six permissions (deduped against existing), quotes both bare paths (`"${HOME}/…/check-setup.sh"`) and arg-bearing commands (`"${HOME}/.claude/bin/ways" init`) correctly, and is idempotent across re-runs. Nice touch quoting the literal `${HOME}` *before* expansion — that's the right way to survive a spaced home dir on Windows.

Two optional non-blockers, your call — happy to merge without them:
- Step 3 (skills/agents/commands) `cp -r`s unconditionally, while steps 4–5 skip when the dest is symlinked to source. Harmless `cp`-onto-self if someone symlinked those dirs, but inconsistent with the rest.
- `backups/sync-*` dirs accumulate with no pruning — maybe keep the last N.

Once CI's green I'll merge. Thanks again — this is a genuinely useful contribution.

PR #184, opened 2026-06-23
ADR-140 integration: sync-to-home as deterministic mechanism + topology observability

aaronsb:
Implements ADR-140 (and the follow-on it authorized) on top of the merged #182. Turns the subdirectory topology into a fully-supported, observable second install path across the ADR-138 triad — way (why) / skill (how) / make (mechanism).

## What changed

**Mechanism** — `scripts/sync-to-home.sh` + `make sync-to-home` (and `-link`)
- Extracts #182's inline command-bash into a real script (the bug class that hid in markdown now has a home and a test). Copy mode (default) + symlink mode in one script.
- Topology guard, backup with prune-to-last-5, symlink-skip across *all* trees (fixes #182's step-3 inconsistency), the fixed `settings.json` jq merge.
- `make sync-to-home-test` smoke-tests the merge: valid JSON, user settings preserved, perms deduped, hooks quoted (bare + arg-bearing), idempotent, symlink mode, topology guard.

**How** — `commands/sync-to-home.md` slimmed to an instructional wrapper (confirm topology → consent → run the make target → report). No inline bash.

**Why** — `hooks/ways/meta/deployment` way carries the topology decision tree (in-place vs subdirectory, copy vs symlink) and fires on install/update/existing-`~/.claude` talk. Verified it embeds and ranks #1 (0.6) for an install-without-clobbering query.

**Observability** — `.claude-source` marker + synced-HEAD stamp (written by the script)
- `check-config-updates.sh`: new Scenario 0 — when `~/.claude` isn't a repo but the marker points at one, check *that* repo; cache `type=subdirectory` with behind + `unsynced`.
- `metrics.rs`: extracted `render_update_status()` (pure, testable); subdirectory case nudges `git pull && make sync-to-home`, and `pulled-but-not-synced` nudges `make sync-to-home` even at behind==0. 6 unit tests.

**Installer** — `install.sh` conflict menu gains "install alongside (keeps your config)" as the recommended, non-destructive first option.

## Test plan

- `make sync-to-home-test` — ALL PASS (copy + symlink + merge + idempotency + guard)
- `cargo test -p ways` — 149 + 6 + 11 pass, 0 fail; `cargo clippy` clean
- `ways lint --global` 0/0; `adr lint` 0 errors (1 pre-existing ADR-127 warning, unrelated)

## Note (follow-up, not in this PR)

A way `description` containing a colon-space was silently dropped from the matching corpus by the strict YAML parser `ways corpus` uses, while `ways lint`/`ways tree` (lenient parser) accepted it. Worked around here; the parser divergence is a real gap — corpus should warn on parse-skip and/or lint should catch it.

Closes the ADR-140 follow-on. Builds on #182.


aaronsb:
## Review: ADR-140 sync-to-home + topology observability

Reviewed against ADR-140 intent. Ran the smoke test (ALL PASS), exercised `sync-to-home.sh` directly with spaces in `SRC`/`DEST`, drove the prune loop, and traced Scenario 0 of the update checker. The architecture is sound — extracting #182's command-prose bash into a tested script behind `make sync-to-home` is exactly the ADR-138 layering the ADR argues for, and the jq merge that shipped broken in #182 now preserves user keys, dedupes permissions, and quotes hook commands correctly (verified `${HOME}`-with-spaces stays a single token because the quote wraps the literal-up-to-first-space and the runtime shell expands `${HOME}` *inside* the quotes). Good work. A handful of findings, none catastrophic.

### Should-fix

**1. Copy-mode projection accumulates stale files — a deleted/renamed upstream way stays active. `scripts/sync-to-home.sh:82`**
`cp -r "$src/." "$dst/"` only ever adds/overwrites; it never deletes. I confirmed this empirically: remove a skill upstream, re-sync, and the old file is still in `~/.claude/skills/` (alongside the new one). For skills/commands/ways this is a correctness problem — Claude Code keeps loading an orphaned way or command that no longer exists in the repo. This is precisely the "pulled but didn't re-sync"-class drift the ADR claims copy-mode makes *observable*, except here it's invisible: the synced-HEAD stamp goes green while orphans linger. `make update` (in-place) doesn't have this because git tracks deletions.
Suggestion: mirror with delete semantics per tree, e.g. `rsync -a --delete "$src/" "$dst/"` when rsync is available, or for the no-rsync path `rm -rf "$dst" && cp -r "$src" "$dst"` for the *fully-owned* trees (skills/agents/commands/hooks/ways) — but **not** for `$DEST` itself or anything the user co-owns. Whatever the mechanism, the backup taken in step 2 is the safety net that makes a destructive-replace of these repo-owned subtrees acceptable.

**2. Backup failure is silently swallowed but reported as success. `scripts/sync-to-home.sh:50-53`**
`[[ -e ... ]] && cp -r ... 2>/dev/null || true` discards every `cp` error, then line 53 logs "Backed up to $BACKUP" unconditionally. On a full disk or a permission fault the backup can be empty/partial, yet the user is told it exists immediately before the projection overwrites their files. For the single highest-risk operation in this PR (mutating `~/.claude` + `settings.json`), a backup whose failure is invisible defeats its purpose.
Suggestion: at minimum verify the `settings.json` copy landed (`[[ -f "$BACKUP/settings.json" ]]` when the source existed) and `die`/warn loudly otherwise; don't print the reassuring "Backed up to" line unless something was actually copied.

**3. The "pulled but not synced" drift nudge is up to an hour stale despite needing no network. `hooks/check-config-updates.sh:107-115`**
The `UNSYNCED` computation (`CUR_HEAD != SYNCED_HEAD`) is a purely local `git rev-parse` comparison, but it lives entirely inside the `needs_refresh` (hourly) + `*github.com*` gate. So after `git pull`, a user who forgets `make sync-to-home` sees no drift nudge until the hourly cache expires — and the ADR sells this nudge as the mitigation for copy-mode drift. Worse, if the subdirectory repo's origin isn't GitHub (line 107), Scenario 0 writes *no cache at all*, so that install gets zero nudges including the local-only drift one.
Suggestion: compute and write the `unsynced` flag on every invocation (it's cheap and offline), gating only the network `fetch`/`BEHIND` on `needs_refresh`. And let the drift check run regardless of remote host — it doesn't need the network.

### Nits

**4. `KEEP_BACKUPS=0` deletes the just-created backup. `scripts/sync-to-home.sh:56`**
`tail -n +"$((0+1))"` = `tail -n +1` = "print everything", so the fresh backup is pruned immediately. The var is user-overridable, so it's reachable. Guard with `(( KEEP_BACKUPS >= 1 ))` or clamp.

**5. Second-granularity backup stamp can collide. `scripts/sync-to-home.sh:47`**
Two syncs in the same second land in the same `sync-$STAMP` dir (the test sidesteps this with `sleep 1`). Harmless today; add `%N` or a `.$$` suffix if you want guaranteed-distinct backups.

**6. `make -C "$SRC" update-binaries` runs even in symlink mode. `scripts/sync-to-home.sh:112-113`**
In symlink mode `bin/` points at the repo, so rebuilding is correct — but it also runs unconditionally before the per-binary symlink-skip check, which is fine, just worth a comment that the rebuild is intentional for both modes.

### Test coverage

The 6 Rust unit tests cover the new subdirectory branches well: behind, unsynced-while-not-behind, clean-silent, missing-repo placeholder, plus clone-behind and zero-behind-silent as guards. Two uncovered branches worth a line each: `behind > 0 && unsynced == true` (current code shows the behind message and suppresses the drift message — looks intentional, but it's the one interaction with two live flags and it's untested), and the `repo` field containing a path with spaces (the marker can legitimately hold one — see the `source=/tmp/space test/repo` case I hit). `render_update_status` being pure was the right call; it makes all of this trivially testable.

The shell smoke test is solid (valid JSON, user-keys preserved, dedupe, both quoting forms, idempotency, guard, symlink mode). The one assertion it can't make today — that a deleted upstream file disappears from the projection — is exactly finding #1.

---
*AI-assisted review via Claude (code-reviewer). Review only — no files modified.*

aaronsb:
Thanks — solid review. All three should-fix items + the nits are addressed in d13a4b8.

1. **Orphan accumulation** — you were right that this silently undercut the ADR's drift claim. Rather than `rsync --delete` (unsafe here: skills/agents/commands are *shared* with the user in this topology, so a blanket delete would nuke the user's own skills), I added a **projection manifest** (`.claude-source-manifest`): on re-sync, remove only files a *prior projection* wrote that no longer exist in source. User files were never in the manifest, so they're never touched. The smoke test now proves it — an upstream-removed skill is pruned while a user-owned file in the same dir survives.
2. **Silent backup failure** — now `die`s if a backup copy of an existing target fails, before any overwrite.
3. **Drift staleness / non-GitHub silence** — the unsynced check is purely local, so it now runs every session for any remote; only the behind-upstream fetch stays hourly-rate-limited (cache rewrites preserve the prior fetch timestamp so they don't reset it).

Nits: `KEEP_BACKUPS` floored at 1, `$$` in the backup stamp, the symlink-rebuild comment, and the repo path in the nudge is now quoted (space-safe).

Test gaps you flagged: added the `behind && unsynced` interaction and a space-bearing repo path (8 metrics tests now), plus the orphan-removal shell assertion. Also caught a `set -e`/`pipefail` footgun your fix surfaced — `build_manifest`'s trailing `[[ -f ]] && echo` returned 1 and aborted the caller; pinned with `return 0`.

`make sync-to-home-test` ALL PASS · `cargo test -p ways` green · clippy clean.

PR #189, opened 2026-06-23
feat(ways): non-interactive rethink --json + how-ways-works explanation cluster

aaronsb:
## Why

`ways rethink` could only show a session's way-firing timeline through the crossterm TUI — the reconstructed record of *how ways actually steered a session* was unreachable for agents, scripts, and CI. And while `cognitive-loop.md` explains how ways is **designed**, nothing showed it **operating** in a real session. This PR closes both gaps, and they reinforce each other: the new dump produces the very artifact the new docs explain.

## What

**`ways rethink --json`** — dumps the full reconstructed timeline as one JSON document:
- `summary` — trigger mix, top ways, and fire / re-disclosure / check / near-miss totals
- `frames` — the per-turn timeline (epoch, token position, active ways, what changed)
- `near_misses` — ways that scored close to threshold but didn't fire, with scores + margins (the TUI omits these entirely)

Runs without the `tui` feature, so it works headless. The frame-reconstruction prelude is extracted into `reconstruct_frames` (shared by TUI + dump, behavior-preserving); the new logic lives in a `rethink_dump` module rather than growing the already-oversized `rethink.rs`.

**`docs/explanation/how-ways-works/`** — a three-page Diátaxis explanation cluster (`01.017–019.E`), the observable companion to `cognitive-loop.md`:
1. **The model** — the five event types and four observable behaviours (first-fire, re-disclosure, near-miss, check)
2. **A long session, read through ways** — a real 97h / 78-way / 490-fire / 821-near-miss session, each behaviour shown in its actual numbers (including the compaction sawtooth in re-disclosure)
3. **Reading the session data yourself** — `ways list` / `stats` / `rethink --json`, with jq recipes and what the numbers mean

Reference doc and the docs README are updated to match.

## Verification

- `cargo build` + `cargo clippy` clean; `cargo test` 11/11 pass (incl. `session_sim`)
- Ran `--json` against the real 97h KG session — summary matches independent analysis exactly; most-recent-session and error paths return clean JSON
- `docs/scripts/doc lint` — 0 errors, 0 warnings

https://claude.ai/code/session_01TgPLxh8Ls2kKTN6AGaMeDM

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,6 +13,7 @@ Map of the documentation tree. For the project overview, see the [main README](.
 | [architecture.md](architecture.md) | System architecture diagrams (Mermaid) for the ways mechanics |
 | [architecture/](architecture/) | Architecture Decision Records (managed by `docs/scripts/adr`) |
 | [design-notes/](design-notes/) | Prose-first framing documents that justify multiple related decisions (complement to ADRs) |
+| [explanation/](explanation/) | Diátaxis explanation pages — conceptual walkthroughs grounded in real scenarios (install topologies, localization, attend messaging, and [how ways works](explanation/how-ways-works/how-ways-works-the-model.md) — the observable companion to cognitive-loop.md) |
 
 ## Guides vs Reference
 
--- a/docs/explanation/how-ways-works/how-ways-works-the-model.md
+++ b/docs/explanation/how-ways-works/how-ways-works-the-model.md
@@ -1,0 +1,152 @@
+---
+id: 01.017.E
+domain: system
+mode: explanation
+related:
+  - "[[ADR-104]]"
+  - "[[ADR-123]]"
+  - "[[ADR-134]]"
+  - "[[01.018.E]]"
+  - "[[01.019.E]]"
+aliases: []
+---
+
+# How ways works — the model
+
+When ways is doing its job, you don't notice it. A premise lands in Claude's
+context at the moment it's relevant, Claude reasons with it, and the work moves
+on. Nothing announces itself. That invisibility is the design working — but it
+makes the system hard to *believe in*, because the help leaves no mark on the
+conversation you can point to.
+
+It does leave a mark somewhere else. **Every time a way fires, the cheap
+substrate writes a line to `~/.claude/stats/events.jsonl`.** That append-only
+record is the observable shadow of the cognitive loop: a turn-by-turn account of
+which premises surfaced, why, and when. This cluster is about reading that
+shadow — what it records, what it reveals about how ways helps, and how to pull
+it out for a session of your own ([[01.019.E]]) or walk a real long one
+([[01.018.E]]).
+
+For the *design* — substrate separation, progressive disclosure, the ledger, the
+awareness layer — read [the cognitive loop](../../cognitive-loop.md). This page
+sits one level lower: not how the system is built, but how you *watch it run*.
+
+## The five things the record captures
+
+Each line is one event with a timestamp, a session id, a project, and a trigger.
+Five event types, and each one is a distinct kind of help made visible:
+
+| Event | What it means | What it tells you |
+|-------|---------------|-------------------|
+| `way_fired` | A premise crossed its threshold and was injected | The system decided this guidance was relevant *here* |
+| `way_redisclosed` | An already-seen way surfaced again after its cooldown | The premise had faded from attention and was refreshed |
+| `check_fired` | A depth-on-demand sub-way pulled in under a fired way | Claude got *more* detail because the situation warranted it |
+| `way_nearmiss` | A way scored close to its threshold but did **not** fire | The boundary — what the system *almost* surfaced, and held back |
+| `session_start` | A session began | The anchor every other event hangs off |
+
+The first four are the load-bearing ones, and they map onto four distinct
+behaviours worth understanding separately.
+
+## Four behaviours, made observable
+
+```mermaid
+flowchart TB
+    subgraph Loop["Claude's turn — the expensive substrate"]
+        direction TB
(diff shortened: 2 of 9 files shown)


aaronsb:
## Review: `rethink --json` + how-ways-works cluster

**What this changes**: Adds `ways rethink --json` — a headless, non-interactive dump of a session's reconstructed way-firing timeline (summary + per-turn frames + near-misses), runnable without the `tui` feature — plus a three-page Diátaxis explanation cluster grounded in a real 97h session. The frame-reconstruction prelude is extracted into a shared `reconstruct_frames`, and the dump logic lives in its own `rethink_dump` module rather than growing `rethink.rs`.

**Assessment**: Solid, well-factored change. I checked the five areas the most rigor was warranted in:

**1. `reconstruct_frames` extraction — behavior-preserving (verified).** The extracted function is a byte-for-byte move of the inlined `token_timeline` → `refire_cache` → `build_frames` sequence that previously lived in `run` (rethink.rs:404–423). The one variable that stayed behind — `context_window_k` (rethink.rs:159) — is still legitimately consumed downstream by the `Player` (lines 180, 306, 316, 334, 358), so it's not dead. The shared items (`WayEvent`, `ActiveWay`, `Frame`, `reconstruct_frames`, `load_session_events`, `find_session_project`, plus `build_frames`/`build_token_timeline`/`find_token_position`) are all correctly left *ungated* by `#[cfg(feature = "tui")]`, so the dump compiles headless; only the TUI loop and `run` carry the gate, and `run` has a proper `not(feature = "tui")` stub (rethink.rs:188). The TUI path is unchanged.

**2. JSON dump logic & edge cases — sound.** Error paths all return clean JSON-ish single-line objects rather than panicking: missing events file → `{"error":"no events recorded yet"}`, no sessions in scope → `{"error":"no sessions found"}`, empty/unknown session → `{"error":"no events for session",...}`. Empty `frames`/`near_misses` serialize as `[]`, single-frame and no-near-miss sessions fall out naturally. `summary.near_misses` is sourced from `near_misses.len()`, so the count and the array can't drift. `epoch_for_ts` relies on lexical `<=` over the frame `timestamp` (`cluster_ts`, raw ISO-8601 from `ev.ts`) vs the near-miss `ts` (same field, same format) — same-format ISO-8601 sorts lexically, so the mapping is correct.

**3. SOLID / structure — clean separation, no duplication of the load-bearing logic.** The genuinely complex part (frame reconstruction) is shared, not copied. `rethink_dump` only owns its own concerns: output shaping (`to_dump_frame`), summary aggregation, and near-miss assembly. Output structs are a deliberate serialization boundary distinct from the internal `Frame`/`ActiveWay` — the right call, since `#[derive(Serialize)]` on the internal types would couple the JSON contract to internal field names. One minor note below.

**4. Rust quality — idiomatic.** `Result`/`?` throughout, no `unwrap` on external data (the `unwrap_or`/`unwrap_or_else` defaults are all on parsed JSON fields), `field_f64` correctly tolerates both stringified (`"0.1693"`) and native numbers, iterator-based parsing keeps allocation reasonable.

**5. Docs ↔ code — accurate, no schema drift.** Cross-checked `reading-the-session-data.md` (the "What the dump contains" table + jq recipes) and `docs/reference/ways-cli.md` against the actual structs in rethink_dump.rs:19–80. Top-level (`session`/`project`/`context_window_k`), the nine `summary` fields, the per-frame `active_ways`/`new_events`, and the near-miss `score_en`/`score_multi`/`thr_en`/`thr_multi`/`margin`/`epoch` all match. The jq recipes reference only fields that exist (`.summary.trigger_breakdown`, `.frames[].new_events`, `.near_misses[].way/.margin`). The scenario doc's summary block maps cleanly onto the `Summary` fields.

### Findings

**Nits (non-blocking):**

- **`rethink_dump.rs:124–147` — `find_session_project` runs the file scan even for the most-recent-session path, which already resolved the project.** In the `session: None` branch, `most_recent_session` already inspected the `session_start` row that carries `project`, but `build_dump` then re-derives it via a second full-content scan in `find_session_project`. Harmless (and matches `rethink::run`'s structure), but if you ever profile multi-MB `events.jsonl`, this is a redundant pass. Not worth changing now.

- **`rethink_dump.rs:117` — error JSON is hand-rolled string interpolation.** `{{\"error\":\"no events for session\",\"session\":\"{session_id}\"}}` would emit invalid JSON if a session id ever contained a `"` or backslash. Session ids are UUID-shaped so this is purely theoretical, but a one-line `serde_json::json!({"error": ..., "session": session_id})` would make all four error paths uniformly escape-safe. Cosmetic.

- **`rethink_dump.rs:95–102` vs `rethink.rs:124–131` — the scope auto-detection block is duplicated verbatim.** Both correctly skip detection when `project` is explicitly passed (the asymmetry is intentional — explicit scope shouldn't be widened by detection). Minor candidate for a shared `resolve_scope(project)` helper, but at ~7 lines the duplication is low-cost and arguably clearer left inline.

**No blockers, no should-fix.** Quality gates pass: tests exist (session_sim, 11/11 per the PR body), no security surface (read-only observation command), behavior preserved, and the docs meet the change's own acceptance criteria. CI green.

### Considerations for future work

- The `near_misses` array is flat and unbounded — a multi-day session reports 821 of them. The docs already steer readers to `jq`-slice, which is the right mitigation, but if a `--top-near-misses N` or summary-only flag is ever wanted, the assembly is already isolated in `build_near_misses` and would be a clean extension point (nicely Open/Closed).
- `top_ways.truncate(15)` is a silent magic constant. Fine as a default; worth a one-word const name (`TOP_WAYS_LIMIT`) if it ever needs to be referenced or tuned.

---

PR #204, opened 2026-06-30
agent-ways 1.0 foundation: XDG taxonomy, reconciler, settings 3-way merge, migrator

aaronsb:
## What

The foundation of agent-ways 1.0 (ADR-142/143/144): turning `~/.claude` from *the clone* into a thin projection of an XDG application. Built incrementally in the worktree, each piece tested + demonstrated against sandbox `HOME`s — the live install is untouched.

## Landed (9 commits)

- **`paths.rs`** — the five-root location taxonomy (`data`/`config`/`state`/`cache` under `$XDG/agent-ways` + the `~/.claude` projection floor), resolving ADR-142's state-vs-Claude-Code boundary.
- **`ways manifest`** — git-derived projection manifest (allowlist subtrees, `git ls-files` contents → the app-vs-user disentangler).
- **`ways reconcile`** — symlink projection materializer (zero-step-update), idempotent, refuses to clobber a legacy in-place clone.
- **`settings.json` three-way merge** — kubectl-apply-style co-ownership (base/ours/theirs), self-auditing (reverts if it touches unmanaged fields), preserves all user content.
- **cache rename** `claude-ways` → `agent-ways` across the engine.
- **`ways migrate`** — `--plan` (read-only classification), `--what-if` (executor dry-run, asserts each phase contract), `--execute` (phased, backup-first, crash-safe/resumable, restore-on-failure). The bootstrap shim is the deliberate copied exemption.

## Verified

Transparency trio proven sandbox-vs-live: way enumeration (115 ways), embedding match (byte-identical hits), settings preservation. Full `--execute` against a sandboxed clone produces the correct XDG layout and the migrated binary reads ways through the projection.

## NOT yet in scope (follow-ups)

- Runtime `events.jsonl` call-sites → `$XDG_STATE` (migrator lifts the file; runtime still writes the old path)
- Three-root runtime scan (W2 — lifted user ways are preserved but not yet *loaded*)
- Bootstrap → SessionStart wiring, copy-mode fallback, deprecation lifecycle
- ADRs stay Draft until the above land

## Test plan

- `cargo test -p ways` — 180 unit + 17 integration green (1 ignored: the live-corpus recall test)
- `cargo clippy -p ways` — clean
- `ways migrate --what-if` against the live `~/.claude` is provably read-only (file count unchanged)

ADRs: 142, 143, 144.

### DIFF
--- a/hooks/ways/check-setup.sh
+++ b/hooks/ways/check-setup.sh
@@ -6,7 +6,13 @@
 # Checks: ways binary → corpus → embedding engine (functional probe)
 
 WAYS_BIN="${HOME}/.claude/bin/ways"
-XDG_WAY="${XDG_CACHE_HOME:-$HOME/.cache}/claude-ways/user"
+# Prefer the 1.0 cache name; fall back to the legacy one for un-migrated installs
+# (must match paths::cache_root() in the binary so the probe checks where the
+# binary actually reads).
+_CACHE="${XDG_CACHE_HOME:-$HOME/.cache}"
+if [[ -d "${_CACHE}/agent-ways/user" ]]; then XDG_WAY="${_CACHE}/agent-ways/user"
+elif [[ -d "${_CACHE}/claude-ways/user" ]]; then XDG_WAY="${_CACHE}/claude-ways/user"
+else XDG_WAY="${_CACHE}/agent-ways/user"; fi
 
 # Nothing to check if this isn't a ways-enabled install
 [[ ! -d "${HOME}/.claude/hooks/ways" ]] && exit 0
--- a/scripts/signal-report.py
+++ b/scripts/signal-report.py
@@ -66,8 +66,21 @@ class Score:
     score: float
 
 
+def _engine_dir() -> Path:
+    # Prefer the 1.0 cache name; fall back to legacy for un-migrated installs.
+    # Matches paths::cache_root() in the binary.
+    cache = Path(os.environ.get("XDG_CACHE_HOME") or (Path.home() / ".cache"))
+    new = cache / "agent-ways/user"
+    if new.is_dir():
+        return new
+    legacy = cache / "claude-ways/user"
+    if legacy.is_dir():
+        return legacy
+    return new
+
+
 def way_embed_path() -> Path:
-    xdg = Path(os.environ.get("XDG_CACHE_HOME", Path.home() / ".cache")) / "claude-ways/user/way-embed"
+    xdg = _engine_dir() / "way-embed"
     if xdg.is_file():
         return xdg
     fallback = Path.home() / ".claude/bin/way-embed"
@@ -102,7 +115,7 @@ def run_match(bin_path: Path, corpus: Path, model: Path, query: str) -> list[tup
 
 def gather(battery: list[tuple[str, str | None, str]]) -> list[Score]:
     bin_path = way_embed_path()
-    xdg = Path(os.environ.get("XDG_CACHE_HOME", Path.home() / ".cache")) / "claude-ways/user"
+    xdg = _engine_dir()
     en_corpus = xdg / "ways-corpus-en.jsonl"
     en_model = xdg / "minilm-l6-v2.gguf"
     mu_corpus = xdg / "ways-corpus-multi.jsonl"
--- a/tools/way-embed/Makefile
+++ b/tools/way-embed/Makefile
@@ -66,8 +66,10 @@ verify: cmake-build
 		exit 1; \
 	fi
 
-# Model download (XDG cache location)
-MODEL_DIR  = $(shell echo $${XDG_CACHE_HOME:-$$HOME/.cache})/claude-ways/user
+# Model download (XDG cache location). Prefer the 1.0 cache name; reuse the
+# legacy claude-ways dir if it already holds a model (un-migrated install) so we
+# don't re-download. Matches paths::cache_root() in the binary.
+MODEL_DIR  = $(shell C="$${XDG_CACHE_HOME:-$$HOME/.cache}"; if [ -d "$$C/agent-ways/user" ]; then echo "$$C/agent-ways/user"; elif [ -d "$$C/claude-ways/user" ]; then echo "$$C/claude-ways/user"; else echo "$$C/agent-ways/user"; fi)
 MODEL_EN   = $(MODEL_DIR)/minilm-l6-v2.gguf
 MODEL_MULTI = $(MODEL_DIR)/multilingual-minilm-l12-v2-q8.gguf
 SRCDIR     = $(CURDIR)/
--- a/tools/ways-cli/src/cmd/corpus.rs
+++ b/tools/ways-cli/src/cmd/corpus.rs
@@ -19,7 +19,7 @@ pub fn run(
         .unwrap_or_else(|| home_dir().join(".claude/hooks/ways"));
 
     // The engine dir holds the way-embed binary + GGUF models — always canonical.
-    let engine_dir = crate::util::normalize_path_sep(&xdg_cache_dir().join("claude-ways/user"));
+    let engine_dir = crate::util::normalize_path_sep(&xdg_cache_dir().join("agent-ways/user"));
     // Corpus artifacts (jsonl, splits, manifest) go to --output if given, else
     // the canonical engine dir.
     let out_dir = match &output_dir {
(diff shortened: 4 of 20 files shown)


aaronsb:
## Code review — agent-ways 1.0 foundation (XDG taxonomy, reconciler, settings merge, migrator)

Reviewed for correctness/safety over style, given this code mutates `~/.claude` and co-owns `settings.json`. The architecture is sound and the test coverage is genuinely good (the single-path what-if/execute design, the settings self-audit, reconcile idempotency + legacy-clone refusal are all strong). Two findings are merge-blockers; both are about claims the code makes but does not (yet) honor.

### BLOCKER 1 — The cache rename is split-brain: only the Rust read/write side moved

`claude-ways → agent-ways` was renamed in `tools/ways-cli/src/**` but **not** in the build/probe tooling that populates and checks that same cache:

- `tools/way-embed/Makefile:70` — `MODEL_DIR = …/claude-ways/user` (where `make setup` *downloads the GGUF models*)
- `hooks/ways/check-setup.sh:9` — `XDG_WAY=…/claude-ways/user` (the SessionStart health probe)
- `scripts/signal-report.py:70,105` — still `claude-ways/user`

Trace `make setup` after this PR (fresh install *or* an existing user who `git pull`s + rebuilds):
1. `tools/way-embed` downloads the model to `…/claude-ways/user/minilm-l6-v2.gguf` (Makefile:70).
2. `ways corpus` resolves `engine_dir = …/agent-ways/user` (`corpus.rs:22`) and looks for the model at `engine_dir/minilm-l6-v2.gguf` (`corpus.rs:428-429`).
3. `auto_embed` finds no model there → `en_model.is_file()` is false → it **silently skips embedding** and writes a corpus with no vectors → semantic matching is dead.

So it breaks for *everyone* who runs `make setup` with this PR, not only the pull-without-migrate path the prompt worried about — the model and the reader disagree on the directory. There is no read-fallback to the legacy path.

It's compounded by `check-setup.sh:9` still probing `claude-ways/user`: an existing user's *old* corpus is still sitting there, so the SessionStart probe reports healthy while the binary reads the empty `agent-ways/user`. The breakage is **masked**, not surfaced.

The migrator's `phase_cache` does rename the dir, but the Makefile would re-download to `claude-ways` on the next `make setup`/update even post-migration, so the split persists.

**Fix options (pick one, but close the split):** route every cache reference through `paths::corpus_dir()`/`cache_root()` (Makefile and shell can't call Rust, so they need their own constant kept in lockstep — or have `ways` print the path); **and** add a read-fallback in the engine: if `agent-ways/user` is absent but `claude-ways/user` exists, read (or one-time rename) the legacy path. At minimum, `make setup`'s `MODEL_DIR`, `check-setup.sh`, and `signal-report.py` must move in the same commit as the Rust rename. (Note `session.rs` / `sessions-root.sh` Windows `claude-ways/sessions` is a different tree — confirm it's intentionally out of scope.)

### BLOCKER 2 — "restore-from-backup on failure" is documented but not implemented

`migrate_exec.rs:9-12` states: *"A failed postcondition in execute restores from the backup and aborts — a half-migrated `~/.claude` is never left behind."* The executor does no such thing. The phase loop (`migrate_exec.rs:92-106`) only propagates the error via `?` with `.with_context(...)`. There is no restore path anywhere in the module; `ctx.backup` is referenced only by the success message. The user's suspicion is correct — only `settings_merge` self-reverts.

This matters most at `phase_projection` (`migrate_exec.rs:232-245`): it **removes** all app top-levels + `.git` + `bin` from `~/.claude` *first* (232-237), then calls `reconcile::run` (239). If reconcile errors partway (incomplete `$XDG_DATA`, a permission error mid-loop, a failing root), `~/.claude` is left gutted — real dirs deleted, symlinks not created — and nothing restores it. The data survives in `$XDG_DATA` + the backup, but recovery is manual, and the doc promises otherwise. For the single most destructive operation in the system, a misleading safety claim is worse than an honest "back up first, restore is manual."

**Fix:** either implement the restore (on any phase `Err`, wipe the partial `dest` and `copy_tree(backup → dest)` before bailing), or correct the module doc + the `--execute` output to say recovery is manual-from-backup and tell the user the exact restore command. Given the blast radius, I'd implement it.

### SHOULD-FIX

3. **`phase_projection` deletes before it verifies the replacement (`migrate_exec.rs:232-245`).** Even with Blocker 2 fixed, prefer build-then-swap: materialize/validate the symlinks (or stage them) before removing the real app dirs, so a reconcile failure can't strand the projection. The postcondition check (248-251) runs only *after* the destructive step.

4. **`git reset --hard HEAD` silently discards user edits to *tracked* app files (`migrate_exec.rs:156-157`).** An in-place clone is exactly the kind of thing people hand-edit (tweak a shipped way, patch a hook). Those edits live in the working tree; the fresh `$XDG_DATA` checkout is reset to pristine, so they vanish (recoverable only from the backup, unannounced). The tracked=app=replaceable model justifies it, but a `git diff --quiet HEAD` precondition that *warns and lists* locally-modified tracked files before discarding them would prevent a silent surprise.

5. **Backup verification is a count gate, not integrity (`migrate_exec.rs:132-135`).** `count_files` (297-303) counts regular files only (symlinks/dirs excluded) and bails only if `got < n`. A same-count-but-corrupt copy passes, and the count is racy against Claude Code writing `projects/` during the run. For "the escape hatch," consider at least verifying total bytes, or snapshotting under a marker that asserts completion. Not blocking, but the gate is weaker than the prose around it implies.

6. **Empty `XDG_*` env vars yield a *relative* root (`paths.rs:44-62`).** `std::env::var("XDG_DATA_HOME")` returns `Ok("")` when the var is set-but-empty (common), so `PathBuf::from("")` → `data_root()` resolves to a relative `agent-ways` against CWD. The XDG spec says treat empty/non-absolute as unset. This is pre-existing in `util::xdg_cache_dir` (so it's consistent, not a regression), but it's now load-bearing for a destructive *relocation* — the migrator could copy the app to a relative dir. Guard with `.filter(|s| !s.is_empty())` (ideally also `Path::is_absolute`).

### NICE-TO-HAVE

7. **`settings_merge` doc overstates "byte-identical" (`settings_merge.rs:22`).** The self-audit compares `serde_json::Value` equality (semantic), and `write_json_atomic` (311-317) re-serializes the whole doc with `to_string_pretty`. With `preserve_order` enabled (confirmed in Cargo.toml) key order is preserved, so churn is minimal — but it's not byte-identical (whitespace/normalization can shift), and the *entire* CC-owned file is rewritten on every change. Reword the doc to "the user's fields are semantically preserved" so the invariant isn't oversold. The merge logic itself (three-way per-event, deprecated-ours removal, the dual-base strip in the audit) looks correct and is well tested.

8. **`#![allow(dead_code)]` is module-wide in `paths.rs:28`.** The transitional rationale is fair, but a blanket allow will also hide *accidental* dead code added later. Per-item `#[allow(dead_code)]` on the not-yet-wired accessors keeps the net tight as call sites adopt them.

9. **Error-swallowing on destructive paths.** `remove_dir_all(&ctx.data).ok()` (`migrate_exec.rs:153`) and the `move_path` copy-then-remove fallback (365-374) drop or hide partial-failure states. Low probability, but on the migrator these are the places a partial failure becomes an inconsistent tree; prefer surfacing the error.

### What's solid
The settings three-way merge and its self-auditing revert; the single-code-path what-if/execute contract model; reconcile's idempotency and its refusal to clobber a legacy in-place clone; the planner being strictly read-only. The classification (git-tracked = app, untracked = user/CC) is a clean disentangler. Once Blockers 1 and 2 are closed and the doc claims match behavior, this is a strong foundation.

---
*Review by code-reviewer (Claude). Analysis only — no code changed.*

aaronsb:
Both blockers + all should-fixes addressed in 69a0f1d:

**BLOCKER 1 (cache split-brain):** `paths::cache_root()` is now fallback-aware (prefer `agent-ways`, reuse legacy `claude-ways` if present — no re-download for un-migrated installs); the 9 Rust sites route through `paths::corpus_dir()`; and the same prefer-new/fallback-legacy logic landed in `way-embed/Makefile`, `check-setup.sh`, and `signal-report.py`. No more split-brain across the pipeline.

**BLOCKER 2 (restore claim):** `phase_projection` now reconciles (builds symlinks) **before** removing any app-internal dir, with a postcondition gate before the destructive step — a reconcile failure can no longer gut `~/.claude`. The doc now describes the real recovery model (resume via markers + manual backup restore) and the executor prints the exact restore command on failure.

**Should-fixes:** dirty-tracked-file warning in `phase_relocate`; empty/relative `$XDG_*` guard; settings-merge doc corrected to 'Value (semantic) equality'; removed the module-wide `allow(dead_code)`.

Verified end-to-end against a sandbox with a dirty tracked file: correct layout, user edits preserved in backup, warning fires. 181 unit + 17 integration green, clippy clean.

PR #205, opened 2026-06-30
agent-ways 1.0 runtime: events→$XDG_STATE + three-root way runtime

aaronsb:
## What

The two correctness gaps the #204 review flagged as follow-ups, closing the loop on the migrator's lifts.

## Item 1 — events.jsonl → `$XDG_STATE`

The telemetry writer (`session::log_event`) + 8 readers now resolve the event log via `paths::events_log()` instead of the hardcoded `~/.claude/stats` path. Fallback-aware (mirrors `cache_root`): prefer `$XDG_STATE/agent-ways/events.jsonl`, reuse the legacy `~/.claude/stats` file if present so an un-migrated install keeps its history; the migrator lifts it, then the preferred path wins. Writer + readers share the resolver.

## Item 2 — three-root way runtime (ADR-143)

Promote two roots → three (core/user/project). The operator's own ways load from `$XDG_CONFIG/agent-ways/ways`, precedence `project > user > core`, **dedup-by-name** so a user way shadows a same-named shipped way without forking.

- `corpus.rs`: scan the user root, dedup user-shadows-core (both passes), `user_hash` staleness, manifest + counts split core/user
- `session.rs`: `resolve_way_file`/`resolve_check_file` insert the user tier
- `check-post.sh`: user root in `WAYS_ROOTS`
- `status`: `Core ways` + `User ways` lines

This is what makes the migrator's lifted user ways actually **load** — without it they were preserved but dead.

## Verified

- Three-root corpus: a user way shadows a same-named core way (user's content wins, core dropped, no double-embed); a unique user way embeds alongside — new `user_way_shadows_core_way` integration test.
- Both fallbacks have deterministic pure-resolver tests.
- `cargo test -p ways` — 182 unit + 18 integration green (1 ignored); `cargo clippy` clean.

Both items carry the same legacy-fallback discipline as the foundation, so un-migrated installs keep working unchanged.

Refs ADR-142, ADR-143.

### DIFF
--- a/hooks/ways/check-post.sh
+++ b/hooks/ways/check-post.sh
@@ -29,9 +29,15 @@ export CLAUDE_PROJECT_DIR="${PROJECT_DIR}"
 # Project-local is kept symmetric with predictive hooks; trust is
 # enforced at the `ways show way` gate (which honors
 # trusted-project-macros for project-local macros) rather than here.
+# Precedence project > user > core (ADR-143). User root is the operator's own
+# ways in $XDG_CONFIG; core is the shipped tree (projected at ~/.claude/hooks/ways).
 WAYS_ROOTS=()
 if [[ -d "${PROJECT_DIR}/.claude/ways" ]]; then
   WAYS_ROOTS+=("${PROJECT_DIR}/.claude/ways")
+fi
+_USER_WAYS="${XDG_CONFIG_HOME:-$HOME/.config}/agent-ways/ways"
+if [[ -d "${_USER_WAYS}" ]]; then
+  WAYS_ROOTS+=("${_USER_WAYS}")
 fi
 WAYS_ROOTS+=("${HOME}/.claude/hooks/ways")
 
--- a/tools/ways-cli/src/cmd/corpus.rs
+++ b/tools/ways-cli/src/cmd/corpus.rs
@@ -65,12 +65,30 @@ pub fn run(
     };
 
     let excluded = crate::util::load_excluded_segments();
+    let empty_skip: std::collections::HashSet<String> = std::collections::HashSet::new();
 
-    // Scan global ways
-    let global_count = scan_ways_dir(&global_dir, "", &excluded, &mut w)?;
+    // Scan USER ways first (ADR-143): the operator's own root in $XDG_CONFIG.
+    // Its ids become the skip set for core, so a user way shadows a same-named
+    // shipped way (precedence project > user > core).
+    let user_dir = crate::paths::user_ways_root();
+    let mut user_ids: std::collections::HashSet<String> = std::collections::HashSet::new();
+    let user_count = if user_dir.is_dir() {
+        let c = scan_ways_dir(&user_dir, "", &excluded, &mut w, &empty_skip, &mut user_ids)?;
+        if c > 0 {
+            log(&format!("User ways: {c} ({})", user_dir.display()));
+        }
+        c
+    } else {
+        0
+    };
+    let user_hash = content_hash(&user_dir);
+
+    // Scan CORE (shipped) ways, dropping any id a user way already claimed.
+    let mut core_written = user_ids.clone();
+    let global_count = scan_ways_dir(&global_dir, "", &excluded, &mut w, &user_ids, &mut core_written)?;
     let global_hash = content_hash(&global_dir);
     log(&format!(
-        "Global ways: {global_count} (hash: {}...)",
+        "Core ways: {global_count} (hash: {}...)",
         &global_hash[..16.min(global_hash.len())]
     ));
 
@@ -156,9 +174,9 @@ pub fn run(
     // Atomic move
     std::fs::rename(&tmpfile, &corpus_path)?;
 
-    let total = global_count + project_total;
+    let total = global_count + user_count + project_total;
     log(&format!(
-        "Generated {}: {total} ways ({global_count} global, {project_total} project)",
+        "Generated {}: {total} ways ({global_count} core, {user_count} user, {project_total} project)",
         corpus_path.display()
     ));
 
@@ -169,6 +187,8 @@ pub fn run(
     let manifest = json!({
         "global_hash": global_hash,
         "global_count": global_count,
+        "user_hash": user_hash,
+        "user_count": user_count,
         "total_count": total,
         "projects": manifest_projects,
     });
@@ -208,7 +228,11 @@ fn embed_one_project(
     }
 
     let prefix = format!("{key}/");
-    let local_count = scan_ways_dir(ways_path, &prefix, excluded, w)?;
+    // Project ids are namespaced ({key}/…), so they can't collide with core/user;
+    // pass fresh dedup sets.
+    let skip = std::collections::HashSet::new();
+    let mut written = std::collections::HashSet::new();
+    let local_count = scan_ways_dir(ways_path, &prefix, excluded, w, &skip, &mut written)?;
 
     if local_count > 0 {
         let local_hash = content_hash(ways_path);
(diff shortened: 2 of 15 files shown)


aaronsb:
## Review — events→$XDG_STATE + three-root way runtime (ADR-142/143)

Reviewed the full diff plus the surrounding match/resolve machinery (`scan/candidates.rs`, `scan/mod.rs`, `scan/scoring.rs`, `show/mod.rs`) because the dedup correctness can only be judged against the path that actually *fires* ways. Item 1 (events log) is sound. Item 2 has one blocker: the user tier was added to the corpus, the resolver, status, and the reactive postcheck hook — but **not** to the predictive scan path, which is where most ways actually fire.

---

### BLOCKER — the user root is invisible to predictive firing (`collect_candidates`/`collect_checks`)

`tools/ways-cli/src/cmd/scan/candidates.rs:12-29` (and `:31-44`) scan **project + core only** — there is no user root:

```rust
let project_ways = PathBuf::from(project_dir).join(".claude/ways"); // project
...
let global_ways = super::scoring::home_dir().join(".claude/hooks/ways"); // core
// (user root $XDG_CONFIG/agent-ways/ways is never walked)
```

The predictive engine in `scan/mod.rs:91-119` iterates these candidates and looks each up in the embed-score map **by `way.corpus_id`** (`:103`). `check-prompt.sh` (`ways scan prompt`), `check-task-pre.sh`, `check-bash-pre.sh`, `check-file-pre.sh` all flow through here. Consequences:

- **A unique user way never fires predictively.** It is embedded in the corpus (`corpus.rs` scans the user root), but no candidate carries its `corpus_id`, so the score is computed and discarded. This contradicts the PR's stated goal ("makes the migrator's lifted user ways actually load") for the primary firing channel. The new `user_way_shadows_core_way` test passes only because it asserts the way is *in the corpus*, not that it *fires* — so the gap is invisible to CI.
- **A shadowing user way fires through the core candidate.** For `meta/foo` in both roots: the only candidate built is from the **core** file (`id=foo`, `corpus_id=foo`); the embed score at `foo` is the user's (corpus dedup kept user's), and `resolve_way_file` renders the user's file. So the embedding vector and the rendered text are the user's, but every **gating field is read from the core file**: `pattern`/`commands`/`files` keyword channels, `when:` (`scan/mod.rs:95`), `scope` (`:92`), `embed_threshold` via `effective_thresholds` (`:104`), `trigger`, `repeat`. A user override of any of those is silently ignored.

The reactive postcheck path (`check-post.sh:38-42`, with the `FIRED[way_id]` cross-root dedup at `:47/:62`) and direct `ways show way <id>` resolution are correctly three-root. It is specifically the predictive scan that was missed. Fix: teach `collect_candidates`/`collect_checks` to walk `paths::user_ways_root()` between project and core, with the same user-shadows-core dedup-by-`id` the corpus uses, so MATCH and RENDER agree.

---

### SHOULD-FIX — corpus dedup keys on *semantic* ways; resolve dedups on *directory existence* (asymmetry → core leak)

In `corpus.rs` Pass 1 the semantic-field gate (`:353`, `if fm.description.is_empty() || fm.vocabulary.is_none() { continue; }`) runs **before** the dedup bookkeeping (`:364-367`, `written.insert(id)`). So a **non-semantic** user way (pattern/files/commands-only — legitimate and common) is never added to `user_ids`, core `foo` is **not** shadowed, and core's semantic `foo` stays in the corpus. Yet `resolve_way_file` (`session.rs:568-585`) prefers the user dir for *any* way dir. Net: corpus matches on core's `foo` embedding, then renders the user's `foo` file — a match/render divergence, and a concrete case of a "shadowed" core way leaking into the corpus (your concern #1). Reconcile the two dedup notions (semantic-keyed vs directory-keyed), or document the limitation.

### SHOULD-FIX — project + user same bare id: match vs render diverge (concern #2)

Project ways are namespaced (`{key}/foo`); user/core are bare (`foo`). If a project **and** a user way share bare id `foo`, the corpus holds `{key}/foo` (project) and `foo` (user, core dropped). The bare-`foo` entry matches on the user embedding, but `resolve_way_file` checks the project dir first (`session.rs:569`) and renders the **project** `foo`. So a user-content match renders project content. This is a pre-existing class for core too (core `foo` matched, project `foo` rendered), now extended to the user tier. Narrow, but worth a guard or a documented precedence note.

### SHOULD-FIX — the delicate Pass 2 (localized) dedup is untested

The Pass 2 skip check (`corpus.rs:400`) is the most subtle part of the change, yet `user_way_shadows_core_way` runs in **English mode** (no `localized_language`, no `.locales.jsonl`), so neither the locale-alias suppression nor the multilingual-anchor path is exercised. Add a localized-mode case where the shadowed core way carries `.locales.jsonl` and assert its aliases + `lang:"en"` anchor do **not** leak. (The logic reads correct — shadowed core ids are absent from `en_roots` so the anchor can't emit, and `:400` drops the aliases — but "correct and untested" on the corpus matcher is exactly where a future regression lands unnoticed.)

---

### NICE-TO-HAVE

- **`core_written` is dead** (`corpus.rs:87`): initialized to `user_ids.clone()`, mutated via `written.insert`, never read (core is the last unnamespaced scan). Drop the clone; pass a throwaway `&mut HashSet::new()`. The doc comment on `written` ("every id actually written") also overstates — Pass 2 never inserts into it.
- **`is_stale` can't see deletions; `user_hash` isn't consulted** (`corpus.rs:690-709`, manifest at `:190`). Staleness is mtime-only; deleting a user way bumps no surviving file's mtime, so the corpus stays stale until an unrelated edit. `user_hash` is written but, like `global_hash`, never compared for staleness. Pre-existing, but bites harder for user ways since the operator edits/deletes them expecting firing to change.
- **Shadow loses localization**: a user way shadowing a *localized* core way that lacks its own `.locales.jsonl` drops all multilingual entries — in a localized install the override silently degrades multilingual matching. Worth a one-line note that user overrides should carry their own locales.
- `is_stale` duplicates the walkdir block (core/user loop vs the project block) — minor DRY.

---

### Item 1 (events_log) — correct, no action

`resolve_events` (`paths.rs:370-380`) "prefer state → else legacy → else state" is right: a fresh in-place install (neither file exists) resolves to the `$XDG_STATE` preferred path, and `log_event` (`session.rs:425-430`) creates the parent from `events_log().parent()` and writes there — new installs go straight to XDG_STATE, no legacy dir created. Within a single short-lived CLI process the resolution is stable (the first write creates the state file, so later calls keep resolving to state); a cross-process flip after the migrator lifts the file is benign for append-only telemetry. Writer and all 8 readers share the resolver, so they never diverge within a call. The pure-resolver factoring + test is the right call.

### Good to see

Atomic corpus write via `rename` (`corpus.rs:175`); the `FIRED` cross-root dedup in `check-post.sh`; the documented `id` vs `corpus_id` split on `WayCandidate`; pure resolver extracted for testing without touching `$HOME`.

---

**Bottom line:** Item 1 is mergeable. Item 2's corpus/resolve/status/postcheck plumbing is correct, but the predictive scan path (`collect_candidates`/`collect_checks`) was not extended to the user root, so unique user ways stay dead in the channel that matters most and shadowing user ways read their gating metadata from the core file. I'd treat the `candidates.rs` gap as a merge blocker for the ADR-143 goal, with the semantic-vs-directory dedup asymmetry close behind.

*AI-assisted review via Claude. Review only — no files modified.*

aaronsb:
Blocker + should-fixes addressed in e7d987f:

**BLOCKER (user root invisible to predictive firing):** `scan/candidates.rs` (collect_candidates/collect_checks) now walks the user root with full **project > user > core** dedup-by-name. A unique user way becomes a candidate (fires); a shadowing user way wins with **its own** gating fields, not core's. New `user_candidate_fires_and_shadows_core` test.

**Should-fix 1 (semantic-vs-directory asymmetry):** corpus core-skip now uses a shared `candidates::way_ids()` (all user ids by directory, incl. non-semantic), so a pattern-only user override suppresses the core way in the corpus too — and the scanner + corpus dedup identically.

**Should-fix 2 (project+user/core bare-id divergence):** the candidate dedup now spans the full project>user>core chain, so what fires matches what `resolve_way_file` renders. This also tightens the pre-existing project-vs-core case.

**Nice-to-have:** dropped the dead `core_written` clone.

**Deferred (low-risk, noted):** a localized Pass-2 dedup test (logic mirrors the tested Pass-1 skip), and hash-based staleness/deletion detection in `is_stale` (pre-existing mtime mechanism; `user_hash` matches the existing `global_hash` convention).

183 unit + 18 integration green, clippy clean.

PR #213, opened 2026-07-01
fix(install): native projection install (closes the 1.0.0 fresh-install bug)

aaronsb:
## Summary

agent-ways 1.0.0's installer still **cloned into `~/.claude`**, producing the pre-1.0 in-place shape — so a brand-new 1.0 user landed in exactly the topology 1.0 replaced, then had to migrate off it. This reworks `install.sh` to the native 1.0 path, wiring the machinery ADR-144/145 already anticipated (fresh = *materialize the manifest into an empty tree* = `ways reconcile`).

**New flow:** stage app → `$XDG_DATA_HOME/agent-ways`, `make setup`, link binaries onto PATH, `ways reconcile` the projection into `~/.claude` (which also seeds `settings.json` and preserves Claude Code's own files). No new logic — `ways reconcile` already does projection + settings merge.

## Behaviour

| State | Action |
|---|---|
| fresh (empty or CC-only `~/.claude`) | stage → build → reconcile |
| already installed (app dir is our repo) | pull + rebuild + reconcile |
| legacy pre-1.0 in-place clone at `~/.claude` | point to gated `ways migrate` |
| app dir occupied by non-agent-ways | stop (don't clobber) |

Native projection made the old **four-option conflict menu disappear** — an existing `~/.claude` with CC files is no longer a problem to resolve, it's the normal case we project into.

## Validation

Sandbox `HOME`/`XDG` run of the reworked installer:
- ✅ app staged to `$XDG_DATA/agent-ways`
- ✅ `~/.claude` has **no `.git`** (not the in-place shape)
- ✅ `skills`/`agents`/`commands`/`hooks/ways` are symlinks into the app dir
- ✅ `settings.json` seeded via the three-way merge (hooks + 6 ways perms, backup written)

(The sandbox couldn't fetch/build the binary — `gh` auth is broken under an overridden `HOME` — so the reconcile step was validated with a supplied binary. The installer correctly *skips* reconcile and warns when no binary is present, rather than half-installing.)

## Follow-ups (not this PR)

- Binary acquisition for **unauthenticated** fresh installs (`download-ways.sh` uses `gh release view`).
- The `ways` make target leaves a **dangling `bin/ways` symlink** when the build fails (same class as #210).
- Makefile `help` text + the remaining doc/way sweep (`.claude/todo-1.0-path-doc-audit.md`).

## Test plan

- [x] `bash -n` clean; sandbox projection verified (no `.git`, symlinked trees, seeded settings)
- [ ] Real fresh install on a clean machine (cube / sandbox with gh auth) to exercise binary fetch

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,22 +1,29 @@
 #!/usr/bin/env bash
-# Install agent-ways into ~/.claude
+# Install agent-ways as an XDG application projected into ~/.claude (1.0, ADR-142/144).
 #
-# This is a gateway, not a butler. It checks readiness, detects complexity,
-# and either proceeds with a clean install or tells you what to sort out first.
+# The application lives in $XDG_DATA_HOME/agent-ways; ~/.claude becomes a thin
+# projection (symlinks + Claude-Code-owned files). This is a gateway, not a butler:
+# it stages the app, builds it, and reconciles the projection.
 #
 # Usage:
 #   scripts/install.sh                           # install from repo root
 #   scripts/install.sh --bootstrap               # clone latest + install
 #   curl ... | bash -s -- --bootstrap            # self-bootstrap from internet
-#   scripts/install.sh --dangerously-clobber     # overwrite existing ~/.claude/
+#   scripts/install.sh --dangerously-clobber     # nuke ~/.claude + app dir, reinstall
 #
-# The happy path: clone → make setup. Everything else is guardrails.
+# The happy path: stage app → make setup → ways reconcile. Everything else is guardrails.
 
 set -euo pipefail
 
 UPSTREAM_REPO="aaronsb/agent-ways"
 UPSTREAM_URL="https://github.com/${UPSTREAM_REPO}"
 DEST="${HOME}/.claude"
+
+# The application root (1.0 XDG layout). The repo is staged here; ~/.claude is a
+# projection of it. Honours XDG_DATA_HOME so sandbox installs stay contained.
+XDG_DATA="${XDG_DATA_HOME:-${HOME}/.local/share}"
+APP_DIR="${XDG_DATA}/agent-ways"
+XDG_BIN="${HOME}/.local/bin"
 
 # --- Colors ---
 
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: native-projection installer

Bash hygiene is solid — `set -euo pipefail` holds up against the `[[ ... ]] && ... && ...` top-level chains (leading-test-false doesn't trip `-e`, verified), the dangling `bin/ways` symlink is handled consistently (`[[ -e ]]` in `link_path_binaries` and `[[ -x ]]` at the reconcile guards both evaluate false on a dangling link, so it's skipped rather than half-linked), and the old `local`-outside-a-function latent bug is gone. No destructive path can eat a user's real `~/.claude`/`projects/` without a recoverable backup. Findings below, most-severe first.

### 1. Dead `SRC == $APP_DIR` branch; guard at :262 pre-empts it (moderate)
`scripts/install.sh:262` (`[[ -e \"$APP_DIR\" ]] → exit 1`) runs before the fresh-install `if [[ \"$SRC\" == \"$APP_DIR\" ]]` at `:272-273`. To reach :272, `$APP_DIR` must not exist (:262 false) — but if `SRC == APP_DIR`, then `APP_DIR` *is* the dir we're running from, so it exists and :262 exits first. The `:272-273` true-branch is therefore unreachable, and the `else` (git clone) always runs.

Real impact: a contributor whose working clone lives at `$XDG_DATA_HOME/agent-ways` with an origin we don't recognize (local clone, renamed remote, or a fork while `gh` is unauthed) fails `is_agent_ways_repo` at :222, isn't legacy at :245, then hits :262 → "exists but is not an agent-ways clone… move it aside or clobber." Non-destructive, but confusing, and the :272 code shows the author meant to support exactly this. Fix: move the `SRC == APP_DIR` short-circuit above the :262 guard (or fold it into :222's branch).

### 2. `$APP_DIR`-is-repo precedence reconciles into an undetected legacy `~/.claude` (moderate)
`is_agent_ways_repo \"$APP_DIR\"` (:222) is checked before `is_agent_ways_repo \"$DEST\"` (:245). If both hold — a 1.0 app dir exists **and** `~/.claude` is still a pre-1.0 in-place clone (full git working tree) — the update branch runs `ways reconcile --dest \"$DEST\"` onto that clone and never routes to `ways migrate`. Worth confirming reconcile is safe when `$DEST` contains its own `.git`/working tree (symlinking projection roots over a populated repo), or add a legacy-`$DEST` check inside the update branch before reconciling. Odd state, but reachable for someone who set up the app dir manually without migrating.

### 3. Silent "Updated." when the rebuild fails on update (low)
`:237` guards reconcile as `[[ -x … ]] && … || true`, and `:233` `make … setup || true`. So a failed rebuild (no fresh `bin/ways`) still falls through to `:239` "Updated. Restart Claude Code" with no warning that reconcile was skipped. Consider echoing a warning in the not-`-x` case, mirroring the fresh path's explicit message.

### 4. Degraded fresh path leaves an empty `~/.claude` (low)
`:300` `mkdir -p \"$DEST\"` runs before the `:301` `[[ -x bin/ways ]]` check, so a build failure leaves a freshly-created empty `~/.claude` plus the app dir. Recoverable and the `ways reconcile` hint is correct (its defaults resolve to this app dir + `~/.claude`) — but `ways` isn't on PATH yet since `link_path_binaries` linked nothing. Minor; a note that they must build first / call `$APP_DIR/bin/ways` would tighten it.

### 5. Non-interactive clobber (note, by design)
`curl … | bash -s -- --dangerously-clobber` takes the `:212` non-interactive branch and moves `~/.claude` aside with only a warning, no `clobber` confirmation. Backed up, and the flag name earns it — flagging only for awareness.

Net: ship-able. #1 and #2 are the ones worth a look — both are edge states rather than fresh-install-path breakage, and the happy path (fresh, already-installed, legacy→migrate, occupied→stop) routes correctly.

---

PR #214, opened 2026-07-01
fix(ways): rewrite injected install doctrine to the 1.0 projection model

aaronsb:
## Summary

- The **session-injected** deployment doctrine still asserted the pre-1.0 identity (`~/.claude *is* the clone`, "this repo IS `~/.claude`") and gave repair commands (`cd ~/.claude && make setup`) that **break** under the 1.0 native projection — there's no Makefile in the projection.
- Per doc-sweep decision **D1 = supersede** (native projection is THE 1.0 model; in-place/subdirectory are legacy), rewrites the three highest-stakes surfaces:
  - `hooks/ways/meta/deployment/deployment.md` — no longer a two-topology chooser; teaches the projection model (source in `$XDG_DATA`, `~/.claude` as thin symlink projection) and the one real fork: a legacy in-place clone that must `ways migrate`, not `git pull`.
  - `hooks/ways/meta/skills/skills.md` — "this repo IS `~/.claude`" → "`skills/` projects into `~/.claude/skills/`"; the tight-triggers lesson is unchanged.
  - `hooks/ways/check-setup.sh` — the three repair messages point at `$XDG_DATA_HOME/agent-ways` + `ways reconcile`, not `cd ~/.claude && make setup`.

This is **Category A (A1/A2/A3)** of the 1.0 `~/.claude` path & doctrine audit — the injected-every-session doctrine + one broken command. Model matched to the known-correct `docs/development.md`.

## Test plan

- `bash -n hooks/ways/check-setup.sh` — clean; heredocs render the resolved app path with literal backticks (verified).
- `deployment.md` frontmatter is well-formed; `pattern`/`vocabulary` still match install/update/`~/.claude`/reconcile/migrate so the way still fires.
- Prose only elsewhere — no runtime behavior change.

### DIFF
--- a/hooks/ways/check-setup.sh
+++ b/hooks/ways/check-setup.sh
@@ -6,6 +6,9 @@
 # Checks: ways binary → corpus → embedding engine (functional probe)
 
 WAYS_BIN="${HOME}/.claude/bin/ways"
+# The app source (with the Makefile) lives in $XDG_DATA, not in the ~/.claude
+# projection — repairs run there, not in ~/.claude (ADR-142).
+APP_DIR="${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways"
 # Prefer the 1.0 cache name; fall back to the legacy one for un-migrated installs
 # (must match paths::cache_root() in the binary so the probe checks where the
 # binary actually reads).
@@ -18,17 +21,19 @@ else XDG_WAY="${_CACHE}/agent-ways/user"; fi
 [[ ! -d "${HOME}/.claude/hooks/ways" ]] && exit 0
 
 if [[ ! -x "$WAYS_BIN" ]]; then
-  cat <<'MSG'
+  cat <<MSG
 
-⚠️  Ways setup incomplete — the `ways` binary is not installed.
+⚠️  Ways setup incomplete — the \`ways\` binary is not installed.
 
-Hooks will be inactive until setup completes. Run:
+Hooks will be inactive until setup completes. Build the app source and
+reproject (the Makefile lives in the app dir, not in ~/.claude):
 
-    cd ~/.claude && make setup
+    cd "$APP_DIR" && make setup && ways reconcile
 
-This downloads the ways binary, embedding model, and generates
-the matching corpus. If you don't have a Rust toolchain, pre-built
-binaries are downloaded automatically.
+If the app dir doesn't exist, re-run the installer one-liner. Setup
+downloads the ways binary, embedding model, and generates the matching
+corpus; if you don't have a Rust toolchain, pre-built binaries are
+downloaded automatically.
 
 MSG
   exit 0
@@ -37,13 +42,13 @@ fi
 # Binary exists — check corpus
 CORPUS="${XDG_WAY}/ways-corpus.jsonl"
 if [[ ! -f "$CORPUS" ]]; then
-  cat <<'MSG'
+  cat <<MSG
 
 ⚠️  Ways corpus not generated — semantic matching is inactive.
 
-Run:
+Rebuild it from the app source:
 
-    cd ~/.claude && make setup
+    cd "$APP_DIR" && make setup
 
 MSG
   exit 0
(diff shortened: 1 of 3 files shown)


aaronsb:
## Code Review — doctrine sweep to the 1.0 projection model

Reviewed the three touched files against the projection model (manifest.rs, install.sh, the `ways` CLI subcommands, and the reference `docs/development.md`). The doctrine rewrite is accurate and the prose lands. One real shell bug in `check-setup.sh`, plus a wording nit and one out-of-scope heads-up.

### Finding 1 — Medium: `ways reconcile` in the "binary not installed" repair message can fail with command-not-found

**Location**: `hooks/ways/check-setup.sh:31` (primary), and the same pattern at `:85`.

**Problem**: The first message fires precisely when `~/.claude/bin/ways` is not executable — i.e. the projection is not materialized. In that state `ways` is generally **not on PATH**: `make setup` builds binaries into `$APP_DIR/bin` and generates the corpus, but it does **not** link them onto `~/.local/bin` — that PATH bootstrap is a separate step the installer performs (`Makefile` `install:` target, and `scripts/install.sh:317`). So chaining `&& ways reconcile` after `make setup` hits `command not found` in exactly the greenfield/broken case this message targets (its own body says "Setup downloads the ways binary" — the reader has no binary yet).

**Reference**: The project's own authoritative recipe invokes reconcile by full path for this reason — `scripts/install.sh:337-338`:

```
cd $APP_DIR && make setup      # build the binary
"$APP_DIR/bin/ways" reconcile   # materialize the projection
```

(also `install.sh:254`, `:326` — always `"$APP_DIR/bin/ways" reconcile`, never bare `ways`).

**Suggestion**: match the installer.

```
    cd "$APP_DIR" && make setup && "$APP_DIR/bin/ways" reconcile
```

Apply the same at `:85` for consistency (lower stakes there — that branch only runs when `$WAYS_BIN` is already executable, so bare `ways` is *more* likely resolvable, but not guaranteed since `~/.claude/bin` is rarely on PATH).

### Finding 2 — Nit: `ways migrate --what-if` labeled "preview: what moves where"

**Location**: `hooks/ways/meta/deployment/deployment.md` (migrate block).

Per the CLI, the "what moves where" read-only preview is `--plan` (the default); `--what-if` is a dry-run of the *executor* that asserts each phase's contract without mutating. Both are read-only, and `install.sh:273` also frames `--what-if` as "preview (read-only)", so this is **consistent with the established convention** — flagging only as an optional clarity tweak, not a correctness error.

### Out of scope (non-blocking, for the follow-up categories)

`README.md` still teaches the pre-1.0 model — `cd ~/.claude && make setup` (`:100`), the `sync-to-home` / `--dangerously-clobber` conflict menu (`:113`). Not part of this PR (the body scopes this as Category A of a broader audit), just noting it so a later category captures it.

### What's solid

- **Doctrine is factually correct.** Projected roots in `deployment.md` (`skills/`, `agents/`, `commands/`, `hooks/ways/`, built binaries) match `manifest.rs` `PROJECTED_TREES` + `PROJECTED_BINS`; the "source in `$XDG_DATA`, `~/.claude` as thin symlink projection, everything else stays unprojected" framing matches `development.md` and the manifest header. No lingering "this repo IS `~/.claude`" / "in-place is the default" in the touched files (grep-clean).
- **Commands exist and behave as described.** `ways reconcile` and `ways migrate --what-if/--execute` are real subcommands; bare `ways reconcile` is argument-correct (source/dest default to `$XDG_DATA/agent-ways` and `~/.claude`). The one gap is PATH resolvability, per Finding 1.
- **Shell is sound.** `bash -n` clean. The `<<'MSG'` → `<<MSG` conversion is safe: the only `$` in any heredoc is the intended `$APP_DIR`; the backticks at `:26` are escaped (`\`ways\``); no `$(...)` or unescaped backticks remain, so no unwanted expansion or command substitution. `APP_DIR="${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways"` matches `install.sh:24-25`.

Recommendation: fix Finding 1 (one-line change, already modeled by `install.sh`) before merge; Finding 2 is optional.

---

PR #215, opened 2026-07-01
fix(skills,ways): resolve app-internal script refs via XDG, not ~/.claude

aaronsb:
## Summary

Three skills/ways invoked helper scripts at `~/.claude/scripts/…` and `~/.claude/tools/…`. Under the 1.0 projection, **only** `skills/ agents/ commands/ hooks/ways/ bin/` are projected into `~/.claude`; `scripts/` and `tools/` deliberately stay in `$XDG_DATA` (`manifest.rs`). So these commands **fail when run today** — the paths don't exist in the projection.

Per doc-sweep decision **B-form = resolve at app location**, they now point at `"${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways/{scripts,tools}/…"`:

- `skills/project-pulse/SKILL.md` — the run command
- `skills/gh-monitor/SKILL.md` — ci-watch (×2) + inbox-watch Monitor commands (quoted)
- `hooks/ways/softwaredev/visualization/diagrams/diagrams.md` — the mmaid downloader

**Deferred (task #6):** diagrams.md's `~/.cache/claude-ways/user/mmaid` cache-name refs are left as-is — they're still *functional* (the download tooling writes `claude-ways` today). Flipping them to `agent-ways` now would break the fallback until the downloader is fixed; both move in lockstep in the embedding/tooling chunk.

This is **Category B (B1/B2/B3)** — the real runtime breakage — of the 1.0 `~/.claude` path & doctrine audit.

## Test plan

- `grep` confirms no `~/.claude/scripts` or `~/.claude/tools` refs remain in the three files.
- Path form matches `scripts/install.sh` and `docs/development.md` (`${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways`).
- Prose/instruction change only — no code behavior change; the referenced scripts are unmodified.

### DIFF
--- a/hooks/ways/softwaredev/visualization/diagrams/diagrams.md
+++ b/hooks/ways/softwaredev/visualization/diagrams/diagrams.md
@@ -25,7 +25,7 @@ command -v mmaid || ~/.cache/claude-ways/user/mmaid --version
 If not installed:
 - **Arch Linux**: `yay -S mmaid` (AUR, when published)
 - **Go**: `go install github.com/aaronsb/mmaid-go/cmd/mmaid@latest`
-- **Download**: `bash ~/.claude/tools/mmaid/download-mmaid.sh`
+- **Download**: `bash "${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways/tools/mmaid/download-mmaid.sh"` (the `tools/` tree is app-internal, not projected into `~/.claude` — ADR-142)
 
 ## Usage
 
--- a/skills/gh-monitor/SKILL.md
+++ b/skills/gh-monitor/SKILL.md
@@ -39,9 +39,10 @@ to leave running afterward.
 
 Launch with the **Monitor tool** (not Bash):
 
-- **command**: `bash ~/.claude/scripts/gh-monitor-ci-watch.sh` — append a PR
-  number/branch/URL only if watching something other than the current branch,
-  e.g. `bash ~/.claude/scripts/gh-monitor-ci-watch.sh 142`
+- **command**: `bash "${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways/scripts/gh-monitor-ci-watch.sh"` —
+  append a PR number/branch/URL only if watching something other than the current
+  branch, e.g. `bash "${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways/scripts/gh-monitor-ci-watch.sh" 142`
+  (the `scripts/` helpers live in the app dir, not the `~/.claude` projection — ADR-142)
 - **description**: `gh-monitor ci: <repo>#<pr>` (fill in the PR you're watching)
 - **persistent**: `false`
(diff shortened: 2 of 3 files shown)


aaronsb:
## Review: app-internal script refs via XDG

Verified all five points from the review brief. No blocking issues; this is a correct, well-scoped fix.

### 1. Path form matches install.sh — correct
`${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways` is the exact semantic equivalent of `scripts/install.sh`'s `APP_DIR`:
```sh
XDG_DATA="${XDG_DATA_HOME:-${HOME}/.local/share}"
APP_DIR="${XDG_DATA}/agent-ways"
```
The only cosmetic difference is `$HOME` vs install.sh's `${HOME}` — both expand identically. Good.

### 2. Referenced helpers exist — confirmed
All four resolve in-repo, so the app-dir path resolves post-install:
- `scripts/project-pulse`
- `scripts/gh-monitor-ci-watch.sh`
- `scripts/gh-monitor-inbox-watch.sh`
- `tools/mmaid/download-mmaid.sh`

### 3. gh-monitor Monitor `command` quoting — shell-correct
`bash "${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways/scripts/gh-monitor-ci-watch.sh"` is sound: `${VAR:-default}` expands inside double quotes, the nested `$HOME` in the default expands, and the whole path being quoted keeps it space-safe. The trailing ` 142` sits outside the quotes as a separate argument — correct. This is no more shell-dependent than the prior `~/.claude/...` form (tilde also requires shell interpretation of the command), so there's no regression in the expansion-context assumption.

### 4. No stale refs left behind — clean
Zero `~/.claude/scripts` or `~/.claude/tools` occurrences remain in the three touched files. (The one hit elsewhere, `hooks/ways/meta/deployment/deployment.md:42`, is descriptive prose about detecting a legacy pre-1.0 in-place clone, not a helper invocation — correctly untouched.)

### 5. Deferred `~/.cache/claude-ways/user/mmaid` refs — defensible
`tools/mmaid/download-mmaid.sh:30` still writes `${XDG_CACHE_HOME:-$HOME/.cache}/claude-ways/user`, which is exactly what `diagrams.md` lines 17/22/38 point at. Leaving the doc as-is keeps it truthful to the tooling; changing only the doc would create a doc/tooling mismatch. Deferring until the download tooling migrates off `claude-ways` is the right call.

### Nice touches
The inline `— ADR-142` annotations explaining *why* the path resolves to the app dir rather than the projection are the right move — they pre-empt the exact confusion that produced the bug.

### Optional, non-blocking
The `agent-ways` (app dir) vs `claude-ways` (cache dir) naming asymmetry is pre-existing debt surfaced but correctly not addressed here. Worth a tracking issue if one doesn't already exist, so the deferred `diagrams.md` cache refs get closed out when the tooling migrates.

Assessment: correct fix, accurately scoped, honest about what it defers. Ship it.

---

PR #218, opened 2026-07-01
fix(skills): rewrite ways-update for the 1.0 projection update flow

aaronsb:
## Summary

`ways-update` was **broken on every 1.0 install**: its pre-flight resolved `ROOT="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"` and required a Makefile + `.git` *there*. Under native projection `~/.claude` is not a repo and has no Makefile, so the skill aborted with "Not an agent-ways checkout" before doing anything.

Rewritten to the projection model (ADR-142):
- Resolves the app source at **`$XDG_DATA_HOME/agent-ways`** (where the checkout + Makefile actually live), not `~/.claude`.
- **Legacy guard:** if `~/.claude` *is* a pre-1.0 in-place clone (own `.git` + ships `tools/`+`docs/`), routes to `ways migrate` instead of updating in place.
- **Flow:** `git pull` (or fetch/merge `upstream` for a fork) → `make setup` → `ways reconcile` to refresh the projection + re-merge `settings.json`.
- Preserves the pre-flight discipline (ff-only, dirty/diverged handling, never stash/discard without asking).

This is **Category A (A7)** of the audit — upgraded from a one-line fix to a full rewrite once it was clear the skill was non-functional under projection.

## Test plan

- `grep` confirms no residual `CLAUDE_CONFIG_DIR` / `$ROOT` / in-place `make update` / "repo that backs" refs.
- Frontmatter `name: ways-update` matches the directory; description keeps the trigger phrasing ("update agent-ways", "pull the latest ways/hooks", "refresh their ways framework install").
- Commands align with the reconcile/migrate/XDG conventions used by `scripts/install.sh`, `docs/development.md`, and the just-rewritten `install-guide.md`.

### DIFF
--- a/skills/ways-update/SKILL.md
+++ b/skills/ways-update/SKILL.md
@@ -1,85 +1,110 @@
 ---
 name: ways-update
-description: Update the agent-ways installation to the latest version — pull, rebuild the Rust binaries (ways/attend/way-embed), re-install symlinks, and rebuild the corpus. Use when the user wants to update agent-ways, pull the latest ways/hooks, run "make update", or refresh their ways framework install. Not for editing or authoring individual ways (that is the ways skill) or upgrading project dependencies.
+description: Update the agent-ways installation to the latest version — pull the app source in $XDG_DATA, rebuild the Rust binaries (ways/attend/way-embed) and corpus, and refresh the ~/.claude projection with `ways reconcile`. Use when the user wants to update agent-ways, pull the latest ways/hooks, or refresh their ways framework install. Not for editing or authoring individual ways (that is the ways skill) or upgrading project dependencies.
 allowed-tools: Bash, Read
 ---
 
 # ways-update: Update the agent-ways install
 
-Brings the local **agent-ways** checkout (the repo that backs `~/.claude`) up to
-the latest version and rebuilds everything that depends on it. The canonical
-entry point is `make update`; this skill wraps it with the pre-flight checks
-`make` itself doesn't do.
+In 1.0, `~/.claude` is a thin **projection** of an XDG application whose source
+lives in `$XDG_DATA_HOME/agent-ways` (ADR-142). Updating means refreshing *that*
+checkout, rebuilding what depends on it, and reprojecting — not touching `~/.claude`
+directly. This skill wraps the flow with the pre-flight checks a bare `git pull`
+doesn't do.
 
-## What `make update` does
+## The update flow
 
 ```
-make update
-  ├─ git pull --ff-only          # fast-forward to latest; refuses if history diverged
-  ├─ make update-binaries        # force-rebuild ways, attend, attend-chat, way-embed
-  └─ make install                # re-symlink binaries into ~/.local/bin, mark hooks executable
+cd "$XDG_DATA_HOME/agent-ways"
+git pull --ff-only        # (or fetch/merge upstream for a fork — see step 2)
+make setup                # rebuild ways/attend/way-embed + regenerate the corpus
+ways reconcile            # refresh the ~/.claude projection + re-merge settings.json
 ```
 
-Binaries are symlinked, so a rebuild takes effect immediately for *new* shells —
-but a **running Claude Code session keeps the old hooks and ways in memory.**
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: ways-update rewrite for 1.0 projection flow

**Assessment:** The rewrite is correct and lands the skill squarely on the 1.0 model. I verified each claim against source; there are no blocking issues. The flow (`git pull --ff-only` → `make setup` → `ways reconcile --source --dest`) is a faithful mirror of the canonical update path in `scripts/install.sh` (the "Already installed... Updating" branch). Three low/medium robustness notes below.

### What checks out
- **App-dir resolution** — `APP="${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways"` matches `install.sh` (`XDG_DATA`/`APP_DIR`) and `docs/development.md` exactly.
- **`make setup` regenerates the corpus** — `Makefile:65` (`setup: ways attend attend-chat`) ends in `$(WAYS_BIN) corpus --quiet`. Dropping the old separate `ways corpus` step is legitimate.
- **`ways reconcile` accepts `--source`/`--dest`** — confirmed in `main.rs:214-217` and `reconcile.rs:56-64` (defaults: source→`data_root`, dest→`projection_root`, so the bare form in the summary block also works).
- **Ordering** — pull source, rebuild, reproject: correct and matches install.sh.
- **Re-runnable sub-targets** (`update-binaries`, `ways-rebuild`) exist (`Makefile:148,172`).
- **Migrate routing** — `ways migrate --what-if`/`--execute` flags exist (`main.rs:230-244`). Referenced docs (`docs/migration-1.0.md`, `docs/development.md`) both exist.
- No lingering pre-1.0 framing (the Makefile-at-`~/.claude` / `make update` assumption is gone).

### Findings

**1. [Medium] Step 4 calls bare `ways`; canonical flow uses the absolute path.**
`install.sh:254` runs `"$APP_DIR/bin/ways" reconcile --source "$APP_DIR" --dest "$DEST"`, deliberately *not* bare `ways` — precisely to dodge the `~/.local/bin`-not-on-PATH command-not-found hazard that bit #214/#216. For an update the PATH symlink normally exists (and points at `$APP/bin/ways`, so on-PATH == freshly-rebuilt — no divergence risk). But if `~/.local/bin` isn't on the session's PATH, bare `ways reconcile` fails after a successful rebuild. Recommend `"$APP/bin/ways" reconcile --source "$APP" --dest "$HOME/.claude"` for parity with install.sh: immune to PATH gaps and guarantees you run the binary `make setup` just built. (The bare `ways --version`/`ways status` in the verify block are fine as-is — they're convenience checks, not load-bearing.)

**2. [Low] Fork path doesn't verify the `upstream` remote exists.**
`git -C "$APP" fetch upstream && git -C "$APP" merge upstream/main` fails confusingly if no `upstream` remote is configured. Note a standard install sets *origin* to the upstream URL with no separate `upstream` remote (`install.sh:302`), so the Fork branch only applies when the user added `upstream` themselves — and the skill gives no way to detect which branch applies. Suggest a guard: `git -C "$APP" remote get-url upstream >/dev/null 2>&1` before offering the fork path (install.sh's `is_agent_ways_repo` uses origin-owner / `gh` parent as a detection heuristic if you want to auto-route).

**3. [Low] Legacy guard's git check is looser than the source-of-truth.**
Step 0 uses `git -C "$HOME/.claude" rev-parse --git-dir` + `-d tools` + `-d docs`. The authoritative `is_legacy_in_place` (`reconcile.rs:212-215`) tests `dest/.git` *exists* — strictly `~/.claude/.git`. `rev-parse --git-dir` walks *up* the tree, so if `~/.claude` isn't a repo but an ancestor (e.g. `$HOME`) is, it succeeds. The `tools/` + `docs/` conjunction makes a real false-positive unlikely, but tightening to `[ -d "$HOME/.claude/.git" ]` gives exact parity with the Rust guard and removes the ancestor-walk edge entirely.

None of these block merge. (1) is the one worth taking before release since it's the same class of hazard as #214/#216 and install.sh already models the fix.

---

PR #220, opened 2026-07-01
ci(way-embed): publish prebuilt binaries as a release (the #6 linchpin)

aaronsb:
## Summary

Fresh installs land with **semantic matching down** because `way-embed` (the embedding engine) has no prebuilt binary to download — reproduced on cube (stock Arch, no cmake): source build fails → model never downloads → `ways status` shows Engine: none, Model MISSING, 107/115 ways can't fire.

Root cause: `build-way-embed.yml` builds all 4 platform binaries but **only uploads them as ephemeral CI artifacts** — it has no `tags:` trigger and no `release:` job, so `gh release create` never runs. `download-binary.sh` resolves the latest **`way-embed-v*` release** (line 69-70), which never exists → always falls through to source build → needs cmake.

`build-ways.yml` and `build-attend.yml` both publish (tags + release job → `ways-v1.0.0`, `attend-v0.7.1`); `build-way-embed` was the lone workflow missing both. This adds them, mirroring `build-attend.yml:114-133`:
- `tags: ['way-embed-v*']` push trigger
- a `release:` job (`needs: collect`, gated on the tag) that downloads `way-embed-all-platforms` + `gh release create`s it

## Effect

Once merged and a `way-embed-v1.0.0` tag is cut → CI publishes the release → `download-binary.sh` finds it → the installer downloads the platform binary → `setup-binary` succeeds → model download proceeds → **semantic matching works with no C++ toolchain**, exactly like `ways`. This is **fix A** of the embedding-acquisition task; fixes B (a `make deps` toolchain-installer fallback) and C (cache-name align) follow.

## Test plan

- `python3 -c "import yaml; yaml.safe_load(...)"` → valid; jobs = build/collect/**release**, trigger includes `way-embed-v*`.
- Release job structurally identical to the proven `build-attend.yml` release job (same `download-artifact` + `gh release create` shape).
- Full end-to-end validation is the tag cut + a fresh cube install (no cmake) — to be done after merge.

### DIFF
--- a/.github/workflows/build-way-embed.yml
+++ b/.github/workflows/build-way-embed.yml
@@ -3,6 +3,7 @@ name: Build way-embed
 on:
   push:
     branches: [main]
+    tags: ['way-embed-v*']
     paths:
       - 'tools/way-embed/**'
       - '.github/workflows/build-way-embed.yml'
@@ -88,3 +89,30 @@ jobs:
         with:
           name: way-embed-all-platforms
           path: binaries/
+
+  release:
+    needs: collect
+    if: startsWith(github.ref, 'refs/tags/way-embed-v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/download-artifact@v5
+        with:
+          name: way-embed-all-platforms
+          path: binaries
+
+      - name: Create GitHub Release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          TAG="${GITHUB_REF#refs/tags/}"
+          gh release create "$TAG" \
+            --repo "$GITHUB_REPOSITORY" \
+            --title "way-embed ${TAG#way-embed-}" \
+            --notes "Pre-built way-embed (semantic matcher) binaries for 4 platforms.
+
+          Downloaded automatically by \`make setup\` / the installer, so semantic
+          matching works with no C++ toolchain. To build from source instead,
+          install a C++ toolchain (cmake + compiler), then \`make setup\`." \
+            binaries/*


aaronsb:
## Review: CI-only, ships clean

Verified the new `release:` job and tag trigger against the proven siblings (`build-attend.yml` lines 114-137, `build-ways.yml`) and against `download-binary.sh`. **No blocking issues.** One operational note worth recording; the rest is confirmation.

### Structural parity (build-attend.yml release job) — PASS
The new job is a faithful clone of the attend/ways template:
- `needs: collect` ✓
- `if: startsWith(github.ref, 'refs/tags/way-embed-v')` ✓
- `permissions: contents: write` ✓
- `gh release create "$TAG" ... binaries/*` ✓

**Artifact name round-trips correctly.** The `collect` job in *this* file uploads `way-embed-all-platforms` (line 90); the `release` job downloads that exact name (line 102). Confirmed match — not just copied from attend.

### Asset naming vs. `download-binary.sh` — PASS
`collect` publishes `way-embed-<platform>` for `{linux-x86_64, linux-aarch64, darwin-x86_64, darwin-arm64}`. `download-binary.sh` (line 21) builds `BIN_NAME="way-embed-${OS}-${ARCH}"` from `uname -s`/`uname -m`, which resolves to those same four names on each target (notably macOS `uname -m` reports `arm64`, matching `darwin-arm64`, not `aarch64`). The `grep -q "^${BIN_NAME}$"` lookup at line 82 and the `--pattern "$BIN_NAME"` download at line 94 will resolve. `checksums.txt` is generated by `collect` and swept into the release by the `binaries/*` glob, so the optional checksum-verify path (lines 100-121) also works. Chain is complete.

### Release-notes heredoc / shell — PASS
Multi-line `--notes` inside the `run: |` block is correct. Continuation lines (115-117) sit at the block's base indent, so YAML strips them flush-left into a well-formed double-quoted string. Backticks are backslash-escaped, so no command substitution fires; the parens in `(semantic matcher)` / `(cmake + compiler)` are inert inside the quotes. Mirrors attend's construction exactly.

### Tag trigger + `paths` interaction — LOW (operational note, not a defect)
`tags: ['way-embed-v*']` under the same `push:` as `paths:` means GHA requires **both** the ref filter and the path filter to match. A tag push only kicks off `build` → `collect` → `release` if the tagged commit also touches `tools/way-embed/**` or the workflow file. Tag an unrelated commit and the whole chain silently no-ops — no build, so `release`'s `needs: collect` never runs.

This is **real but not a bug in this PR**: it is byte-for-byte the same combination `build-attend.yml` and `build-ways.yml` use, and both have shipped many releases (attend up to v0.7.1, ways to v1.0.0), so the pattern is proven when tags land on version-bump commits (which touch the relevant paths). The existing stray `way-embed-v0.1.0` tag is a live illustration of the failure mode — it points at an old `ways.json`-only commit, doesn't touch `tools/way-embed/**`, and has no release.

**Recommendation (non-blocking):** capture the release procedure in the runbook — cut `way-embed-vX.Y.Z` on a commit that touches `tools/way-embed/**` (e.g. the version bump). No workflow change needed to stay consistent with the established pattern.

**Assessment:** merge-ready. Faithful to the linchpin template, wiring verified end-to-end against the installer's download path.

---

PR #222, opened 2026-07-01
fix(way-embed): correct model repo, align cache name, fix stale ~/.claude strings

aaronsb:
## Summary

Three fixes to the `way-embed` acquisition scripts — covers the `GH_REPO` bug (#3), the cache-name align (#2/fix C), and the way-embed slice of the stale-string sweep (#10):

- **Repo bug (headline):** `download-model.sh:26` had `GH_REPO="aaronsb/claude"` (the pre-rename name), so the GitHub-release model fetch always 404'd and silently fell back to HuggingFace. Corrected to `aaronsb/agent-ways` (matches `download-binary.sh`). The release path now works once a `v0.1.0-model` release exists.
- **Cache name:** both scripts defaulted to `$XDG_CACHE/claude-ways/user`; the `ways` binary reads `$XDG_CACHE/agent-ways/user` (prefer) with a `claude-ways` fallback. Default to **agent-ways**, and pass the Makefile's `MODEL_DIR` (already computes prefer-agent-ways-fallback-claude-ways) into both scripts so binary/scripts/Makefile resolve the **same** dir. The intentional legacy fallback in `MODEL_DIR` stays for un-migrated installs.
- **Stale strings:** build-from-source hints (`cd ~/.claude && make setup`, `~/.claude/tools/way-embed/…`) were broken under the projection (`tools/` isn't projected). Point them at `$XDG_DATA_HOME/agent-ways`. Same in `RELEASE.md`.

## Test plan

- `grep` confirms no stray `~/.claude` / `claude-ways` remain except the deliberate `MODEL_DIR` fallback.
- `bash -n` clean on both scripts; `download-model.sh` arg parsing handles the positional dir after `--multilingual`/`--upstream`.
- Validates end-to-end on the next cube `make setup` (should write/read a consistent cache dir; model fetch prefers the GitHub release once bundled).

### DIFF
--- a/tools/way-embed/Makefile
+++ b/tools/way-embed/Makefile
@@ -75,21 +75,21 @@ MODEL_MULTI = $(MODEL_DIR)/multilingual-minilm-l12-v2-q8.gguf
 SRCDIR     = $(CURDIR)/
 
 model:
-	@[ -f "$(MODEL_EN)" ] || bash "$(SRCDIR)download-model.sh"
+	@[ -f "$(MODEL_EN)" ] || bash "$(SRCDIR)download-model.sh" "$(MODEL_DIR)"
 
 model-multilingual:
-	@[ -f "$(MODEL_MULTI)" ] || bash "$(SRCDIR)download-model.sh" --multilingual
+	@[ -f "$(MODEL_MULTI)" ] || bash "$(SRCDIR)download-model.sh" --multilingual "$(MODEL_DIR)"
 
 # Download directly from HuggingFace (for provenance verification)
 model-upstream:
-	bash $(SRCDIR)download-model.sh --upstream
+	bash $(SRCDIR)download-model.sh --upstream "$(MODEL_DIR)"
 
 # Full setup: download (or build) binary + English model, regenerate corpus.
 # Tries pre-built binary from GitHub Releases first, builds from source if unavailable.
 # The multilingual model (127MB) is on-demand per ADR-139 — fetched via
 # `make model-multilingual` by the ways-localize skill, not at default setup.
 setup: setup-binary
-	@[ -f "$(MODEL_EN)" ] || bash "$(SRCDIR)download-model.sh"
+	@[ -f "$(MODEL_EN)" ] || bash "$(SRCDIR)download-model.sh" "$(MODEL_DIR)"
 	@echo "Regenerating corpus with embeddings..."
 	../../bin/ways corpus --quiet
 	@echo ""
@@ -101,7 +101,7 @@ setup: setup-binary
 
 # Try download, fall back to build
 setup-binary:
-	@if bash "$(SRCDIR)download-binary.sh"; then \
+	@if bash "$(SRCDIR)download-binary.sh" "$(MODEL_DIR)"; then \
 		echo "Pre-built binary installed."; \
 	else \
 		echo "No pre-built binary available, building from source..."; \
--- a/tools/way-embed/RELEASE.md
+++ b/tools/way-embed/RELEASE.md
@@ -5,7 +5,7 @@ Pre-built binaries and model for ADR-108 embedding-based way matching.
 One command — downloads the right binary for your platform + model, generates corpus:
 
 ```bash
-cd ~/.claude && make setup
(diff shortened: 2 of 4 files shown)


aaronsb:
## Review — way-embed acquisition fixes

Reviewed the full diff plus the referenced surfaces (`paths.rs::cache_root`/`resolve_cache`, both script arg parsers, `Makefile` MODEL_DIR). Verified by simulating the parsers and expanding MODEL_DIR under fresh vs. cube conditions. **No blocking issues — the change is correct and consistent.**

### What this fixes (all confirmed)
- **`GH_REPO` bug (headline):** `download-model.sh:26` `aaronsb/claude` → `aaronsb/agent-ways`. `GH_URL` (line 75) now resolves to `https://github.com/aaronsb/agent-ways/releases/download/v0.1.0-model/minilm-l6-v2.gguf`. `MODEL_NAME` default is `minilm-l6-v2.gguf`, so a `v0.1.0-model` release providing that asset satisfies the fetch. Matches `download-binary.sh`'s repo.
- **Cache-name align + MODEL_DIR wiring:** both scripts default to `agent-ways/user`, and all four call sites now pass `$(MODEL_DIR)`.
- **Stale strings:** build-from-source hints now point at `$APP_DIR` (`$XDG_DATA_HOME/agent-ways`). No stray `~/.claude` / `claude-ways` remain in the four touched files except the deliberate `MODEL_DIR` legacy fallback.

### Verification

**1. Arg parsing (MODEL_DIR after flags) — correct.** Simulated both parsers:
- `download-model.sh`: plain / `--multilingual <dir>` / `--upstream <dir>` all route the positional to `OUTPUT_DIR` via `*)`. Flags `shift` (single), positional is captured next iteration. No flag consumes it as a value.
- `download-binary.sh`: only `--release` takes `$2` (shift 2); it's never passed at any call site, so `$(MODEL_DIR)` always lands in `*)` → `OUTPUT_DIR`.

**2 & 3. Consistency / no regression — correct in both states:**
- *Fresh (neither dir):* MODEL_DIR → `agent-ways/user`; scripts `mkdir -p` + write there; binary's `resolve_cache` then sees `agent-ways` and reads `agent-ways/user`. All three agree.
- *Cube (`claude-ways/user` exists, no `agent-ways`):* MODEL_DIR → `claude-ways/user`; scripts write there; binary prefers `agent-ways` (absent) → falls back to `claude-ways`. All three agree, `MODEL_EN` check hits the existing file → **no needless re-download.**

**4. bash -n clean** on both scripts; **RELEASE.md quotes balanced** (the mid-edit issue is resolved — `cd` targets quoted, `bash <path>` args unquoted-but-space-free).

### Non-blocking observations (neither introduced by this PR)

- **Low / latent — check-granularity asymmetry.** The binary's `resolve_cache` keys on the existence of `$XDG_CACHE/agent-ways` (the dir), while `MODEL_DIR` keys on `$XDG_CACHE/agent-ways/user`. A split-brain would require `agent-ways/` to exist *without* `agent-ways/user/` inside while `claude-ways/user` holds the model. In practice unreachable: the only thing ever created under `cache_root()` is `user/` (via `corpus_dir()` and the scripts' `mkdir -p .../user`), and the migrator renames `claude-ways`→`agent-ways` wholesale. Flagging only so the two fallback contracts are known to differ in granularity; no action needed for this PR.
- **Low / pre-existing — `--f16` + GitHub source.** The `--f16` branch changes `QUANT`/`HF_FILENAME`/checksum but not `MODEL_NAME`, and leaves `SOURCE=github`. Today the `v0.1.0-model` release doesn't exist so the `curl --head` check (line 107) falls back to HuggingFace with the correct f16 filename — works. Once a GitHub model release exists, `--f16` would request the Q5-named asset and fail checksum. Adjacent to the `GH_URL` this PR touches, but out of scope here — worth a follow-up when the model release is cut.

### Assessment
Tightly scoped, correct bug fix. The `aaronsb/claude` → `aaronsb/agent-ways` correction removes the silent 404→HuggingFace fallback, and threading `MODEL_DIR` through every call site closes the binary/script/Makefile split-brain in both fresh and un-migrated installs. Good preservation of the intentional legacy fallback with a clear comment. Ready to merge.

---

PR #225, opened 2026-07-01
feat(install): ADR-146 launch-verify + halt-and-instruct recovery card

aaronsb:
## Summary

Second slice of **ADR-146**: the installer now **verifies** semantic matching actually came up (not just "the binary downloaded") and, if it didn't, prints a **recovery card** instead of leaving a silently-degraded install.

- `embedding_engine_ok()` — checks `ways status` reports `Engine: embedding`. A binary that downloaded/built but won't embed is treated as *not acquired*.
- `print_recovery_card()` (ADR-146 halt-and-instruct) — leads with an **agent handoff**: Claude Code is by definition installed (this is being set up *for* it), so it prints a **paste-ready prompt** with the app-dir path + platform substituted — recovery is one copy-paste. Exact `make deps`/`make setup` commands are the backstop. The installer never builds or installs anything itself.
- Wired into **both** the fresh-install and update paths (after each reconcile).
- Adds **`docs/finish-install.md`** — a *discoverable* context doc (explicitly **not** a way or skill, so it never loads into a session on its own) that the card + paste-prompt reference.

Completes ADR-146's runtime behavior (the `make deps` target landed in #224). Local compilation stays a consented last resort; the happy path (prebuilt binary) is unaffected.

## Test plan

- `bash -n scripts/install.sh` clean.
- The recovery card's heredoc is unquoted (so `$APP_DIR`/`$PLATFORM`/colors expand) but all backticks are **escaped** — render test confirms no command substitution fires and the paths substitute correctly.
- `embedding_engine_ok` uses the same `ways status` "Engine:" line validated on cube.
- Happy path unchanged (card only prints when the engine is down).

### DIFF
--- a/docs/finish-install.md
+++ b/docs/finish-install.md
@@ -1,0 +1,75 @@
+# Finishing an agent-ways install (semantic matching)
+
+You're reading this because the installer got everything in place **except** the
+semantic-matching engine — `way-embed` (the embedding binary) or its model isn't
+working on this machine, so ways currently fall back to keyword/pattern matching
+only. This is the guide to finish the job. It's written so **Claude Code can follow
+it** — if a human pasted the installer's recovery prompt, that's you; do the steps
+below and report back.
+
+> This is a **discoverable** doc, not a way or a skill — it does nothing until
+> someone opens it. It won't load into any session on its own.
+
+## The situation
+
+- The app source (with the Makefile and the `way-embed` source) is at
+  **`$XDG_DATA_HOME/agent-ways`** (default `~/.local/share/agent-ways`).
+- `way-embed` builds from a bundled `llama.cpp`, which needs a **C++ toolchain**
+  (cmake + a C++ compiler). The likely reason semantic matching is off: no prebuilt
+  binary exists for this platform *and* the toolchain isn't installed, so the
+  source build couldn't run.
+
+## Steps
+
+Resolve the app dir once:
+
+```bash
+APP="${XDG_DATA_HOME:-$HOME/.local/share}/agent-ways"
+```
+
+**1. Check whether the toolchain is present.**
+
+```bash
+command -v cmake && command -v c++ 2>/dev/null || command -v g++ || command -v clang++
+```
+
+**2a. If cmake / a compiler is MISSING — install the toolchain.**
+
+`make deps` auto-detects the package manager (pacman/apt/dnf/zypper/brew) and
+installs cmake + a compiler + git. **It uses `sudo`.** If you're an agent: *ask the
+human before running it* — installing system packages is their call.
+
+```bash
+cd "$APP" && make deps          # asks for sudo; confirm with the human first
+```
+
+**2b. If the toolchain is already present**, skip straight to build.
+
+**3. Build + regenerate the corpus.**
+
+```bash
+cd "$APP" && make setup
+```
+
+This downloads/builds `way-embed`, fetches the model (~21MB), and regenerates the
+corpus with embeddings.
+
+**4. Verify.**
+
+```bash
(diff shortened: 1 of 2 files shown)


aaronsb:
## Code Review — ADR-146 launch-verify + recovery card

Reviewed against the six verification points. **No blocking issues; no real bugs found.** Verified behaviorally, not just by reading. Details below.

### 1. `set -e` safety — CONFIRMED SAFE
Ran the actual `embedding_engine_ok || print_recovery_card` expression under `set -euo pipefail` across all three engine states:

| Case | Result |
|---|---|
| no binary (`[[ -x ]]` fails → `return 1`) | card prints, script survives, `rc=0` |
| `Engine: none` (grep no-match → non-zero) | card prints, survives, `rc=0` |
| `Engine: embedding` | short-circuits, no card, `rc=0` |

- (a)/(b) confirmed: the `cat` heredoc exits 0, so the `||` expression's status is 0 — `set -e` never fires.
- (c) confirmed: `grep -q` is the function's last command, so its non-zero result is the function's return value. Because the function is *always* invoked in a tested (`||`) context, `set -e` is suppressed inside it — the internal `[[ ]] || return 1` and the `status | grep` pipe (even under `pipefail`) can't trip it.

### 2. `embedding_engine_ok` correctness — CONFIRMED
`status.rs:123` emits `println!("Engine:    {engine}")` — plain text, no ANSI prefix, `engine` is literal `"embedding"` / `"none"` (status.rs:23/25). Against live `ways status`, the line is `Engine:` + 4 spaces + `embedding`. The regex `^Engine:[[:space:]]*embedding` matches; `[[:space:]]*` absorbs the spacing; `-i` covers casing. The `^` anchor correctly excludes the `Ways Engine Status` header line (doesn't start with `Engine:`). Solid.

### 3. Heredoc safety — CONFIRMED
Unquoted `<<CARD`; rendered it in isolation with vars set. `$APP_DIR`/`$PLATFORM`/colors expand correctly; **all five backticks are escaped** (`\`make deps\``, `\`make setup\``, `\`ways status\``) and stay literal — no command substitution fires. No unescaped `$(...)` or stray backtick. Color vars (YELLOW/BOLD/CYAN/DIM/RESET) defined at install.sh:31-39, well before use.

### 4. Call-site placement — CONFIRMED
Both probes sit inside the `if [[ -x "$APP_DIR/bin/ways" ]]` branch — update path (right after reconcile, before `exit 0`) and fresh-install path (after the Done banner). Neither is in the no-binary `else`, so the probe only runs when `ways` exists. `PLATFORM` is defined at top level ahead of both sites.

### 5. `docs/finish-install.md` — CONFIRMED accurate + non-loading
Path is `docs/`, not `hooks/ways/` or `skills/` — plain doc, won't auto-load. Flow checks out against the Makefile: `make deps` auto-detects pacman/apt/dnf/zypper/brew and uses sudo (Makefile:71-90); `make setup` builds way-embed + fetches model + regenerates corpus (Makefile:94); model size `~21MB` matches `download-model.sh`; `Engine: embedding` / `Engine: none` success/failure signals match `status.rs`.

### 6. Other bugs
None found. `bash -n` clean.

### Assessment
Tight, well-scoped slice. The design choice to treat "binary present but engine down" as *not acquired* (rather than trusting the download) is the right call — it's what makes the verify step meaningful. Recovery card leads with the agent handoff, which fits the install-for-Claude context. Happy path is untouched (card only prints when the engine is actually down). Ship it.

One optional future consideration (non-blocking): the two call sites are identical one-liners guarded by identical `[[ -x ]]` conditions — if a third path ever appears, worth a tiny wrapper. Not worth it at two.

---

PR #219, opened 2026-07-01
docs: config-path refs → canonical XDG config.yaml + knowledge.md three-root nuance

aaronsb:
## Summary

**Category C + D** — the final, low-urgency chunk of the 1.0 `~/.claude` path & doctrine audit. Nothing here was *broken* (D3: `~/.claude/ways.json` is still read as a legacy layer — `config.rs:131-135`), so this is a correctness/canonicalization pass.

**Category C (config-path):** the refs that presented `~/.claude/ways.json` as *the* config location now point at the canonical **`$XDG_CONFIG_HOME/ways/config.yaml`** (YAML; disable via `disabled_domains:`, vs the legacy JSON key `disabled`), each noting the legacy file is still honored:
- README "Configuration", `docs/hooks-and-ways.md`, `docs/hooks-and-ways/extending.md`, and the `knowledge.md` disable line.
- **Left intentionally:** the localization docs (`languages.md`, mode-gate, adopter-lifecycle) already describe `ways.json` correctly — as the legacy *Layer 1* overridden by `config.yaml`.

**Category D (authoring nuance):** `knowledge.md` Locations rewritten to the three-root model (ADR-143) — distinguishing the **read-only `~/.claude` projection** (a symlink into `$XDG_DATA`, wiped on update → don't author there) from **user-scope ways** at `$XDG_CONFIG/agent-ways/ways/` (survives updates; verified scanned by `corpus.rs`/`scan/candidates.rs`) and project ways.

The config-file path cited is the **functional loader path** (`config.rs:138`); the `config.rs`-vs-`paths.rs` location disagreement is a separate latent bug, tracked.

## Test plan

- `docs/scripts/doc lint` → 0 errors, 0 warnings.
- User-scope ways path + `disabled_domains` YAML key verified against `paths.rs`/`config.rs`/`corpus.rs` (not aspirational).
- Docs/way prose only.

## Follow-up tasks surfaced across the whole sweep (code, not docs)

- Native-projection update detection missing from `check-config-updates.sh`.
- Subagent domain-disable (`inject-subagent.sh`) reads only legacy `ways.json`, not `config.yaml`.
- Canonicalize the user config-file path (`config.rs` vs `paths.rs` disagree; one is dead).

https://claude.ai/code/session_01TFyiQNDZ8RTMHX4Tnmp1wY

PR #229, opened 2026-07-01
fix(config): canonicalize user config to $XDG_CONFIG/agent-ways/config.yaml (#9)

aaronsb:
`config.rs` read user config from `$XDG_CONFIG_HOME/ways/config.yaml`, but `paths::user_config()` returned `$XDG_CONFIG_HOME/agent-ways/config.yaml` — **dead code nothing read** (flagged in the #219 review). User *ways* live under `agent-ways/` (`user_ways_root`), so the config file sat under a different parent — a latent trap.

**Fix:** add the canonical layer. Order is now defaults → `~/.claude/ways.json` → `$XDG_CONFIG/ways/config.yaml` (legacy) → **`$XDG_CONFIG/agent-ways/config.yaml`** (canonical, via now-live `paths::user_config()`) → project. Both legacy locations stay honored fallbacks — no existing config breaks. Docs updated in lockstep; immutable ADRs left historical.

## Test plan
- `cargo build -p ways` clean; 15 config unit tests pass.
- `apply_ways_json` (legacy `disabled`) and `apply_yaml` (`disabled_domains`) unchanged; only the file-resolution layer added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ For the complete system guide — trigger flow, state machines, the pipeline fro
 
 ## Configuration
 
-User config lives in `$XDG_CONFIG_HOME/ways/config.yaml` (YAML). For example, to disable whole domains:
+User config lives in `$XDG_CONFIG_HOME/agent-ways/config.yaml` (YAML). For example, to disable whole domains:
 
 ```yaml
 disabled_domains:
--- a/docs/hooks-and-ways.md
+++ b/docs/hooks-and-ways.md
@@ -449,7 +449,7 @@ flowchart TD
 
 ## Domain Enable/Disable
 
-`$XDG_CONFIG_HOME/ways/config.yaml` controls which domains are active (a legacy `~/.claude/ways.json` with `{"disabled": [...]}` is still honored):
+`$XDG_CONFIG_HOME/agent-ways/config.yaml` controls which domains are active (a legacy `$XDG_CONFIG_HOME/ways/config.yaml`, and `~/.claude/ways.json` with `{"disabled": [...]}`, are still honored):
 
 ```yaml
 disabled_domains:
--- a/docs/hooks-and-ways/extending.md
+++ b/docs/hooks-and-ways/extending.md
@@ -138,7 +138,7 @@ ways:
 
 ### Disabling an entire domain (global)
 
-Add the domain to `disabled_domains` in `$XDG_CONFIG_HOME/ways/config.yaml`:
+Add the domain to `disabled_domains` in `$XDG_CONFIG_HOME/agent-ways/config.yaml`:
 
 ```yaml
 disabled_domains:
--- a/docs/hooks-and-ways/languages.md
+++ b/docs/hooks-and-ways/languages.md
@@ -24,7 +24,7 @@ There are **two** different language settings; conflating them is the common mis
 
 | Config | Controls | Set by |
 |--------|----------|--------|
-| `language` in `~/.config/ways/config.yaml` (overrides `output_language` in `~/.claude/ways.json`) | the **ways** intl mode + the output-language directive | `ways-localize` |
+| `language` in `~/.config/agent-ways/config.yaml` (overrides `output_language` in `~/.claude/ways.json`) | the **ways** intl mode + the output-language directive | `ways-localize` |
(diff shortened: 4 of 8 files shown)


aaronsb:
## Review: canonicalize user config path

The core change in `config.rs::load()` is correct. Verified against all six review points:

- **Resolution order (correct).** Canonical layer (Layer 3, `agent-ways/config.yaml` via `paths::user_config()`) is applied *after* the legacy `ways/config.yaml` (Layer 2), so it overrides as intended. `user_config()` resolves to `config_root()/config.yaml` = `$XDG_CONFIG/agent-ways/config.yaml` (`config_root()` = `xdg_config_base()/APP`, `APP="agent-ways"`) — confirmed at `paths.rs:171`.
- **Guard is harmless.** `if user_config != legacy_xdg` is always true (`agent-ways/config.yaml` vs `ways/config.yaml` can never be equal), so no double-read risk. Minor: the two sides resolve XDG through *different* helpers (`paths::config_root` vs config.rs-local `xdg_config_dir`), but since the subdirs differ the guard can never fire — effectively dead, but benign.
- **No regression.** Both legacy layers still read; `apply_yaml`/`apply_ways_json` untouched.
- **Project overlay still last (Layer 4)** and still the sole `disabled_ways` populator via `apply_project_ways_overlay` (ADR-131). The canonical layer uses `apply_yaml` only, so it cannot set `disabled_ways` — correct.
- **15 config tests pass** (`config::tests::*` all green locally).

That said, the change is **incomplete** — `load()` now prefers the canonical path, but the two sibling functions that *write* and *report* the config path were not updated.

---

### Issue 1 (medium): `ways config init` still writes to the legacy path

**Location**: `tools/ways-cli/src/config.rs:277` (and the embedded comment at `:285`)

`init_user_config()` still creates the file at `xdg_config_dir().join(\"ways/config.yaml\")` — the path this PR just relabeled *legacy*. The scaffolded comment body also still reads `# User scope: $XDG_CONFIG_HOME/ways/config.yaml`.

**Why it matters**: A fresh user running `ways config init` writes their config to the legacy location. `load()` picks it up via the Layer 2 fallback (so it *works*), but the canonical Layer 3 file the PR introduced stays empty forever. The dead code the PR set out to eliminate (\"`user_config()` nothing read\") is only half-fixed — `user_config()` is now *read* but still never *written* by the tool. This perpetuates legacy-path installs indefinitely, contradicting the PR's canonicalization goal.

**Suggestion**: Point `init_user_config()` at `crate::paths::user_config()` and update the embedded comment string to the `agent-ways/config.yaml` path.

### Issue 2 (medium): `ways config path` reports the wrong file

**Location**: `tools/ways-cli/src/config.rs:305-310`

`config_path()` prints the legacy `ways/config.yaml` as `user:` and never mentions `agent-ways/config.yaml`.

**Why it matters**: This is the exact diagnostic a user runs to locate their config, and it points at the now-legacy file. Combined with Issue 1, `init` writes legacy, `path` reports legacy, and the canonical layer is invisible through the CLI.

**Suggestion**: Report `paths::user_config()` as `user:`, and demote `ways/config.yaml` to a `legacy:` line alongside `ways.json`.

### Issue 3 (medium): doc sweep missed two how-to/reference pages

The docs cited in the PR (README, install-guide, hooks-and-ways, extending, knowledge) were updated, but these still present the legacy path as canonical:

- `docs/hooks-and-ways/matching.md:119` — \"Configure via `~/.config/ways/config.yaml`\"
- `docs/hooks-and-ways/languages.md:27` — ``language`` in `~/.config/ways/config.yaml`

Softer (stale but describing the still-existing legacy Layer 2, lower priority): `docs/design-notes/adopter-localization-lifecycle-and-tuning.md:91` and `docs/explanation/localization/the-mode-gate-mechanism-under-the-scenarios.md:22`. ADRs 115/116/126 correctly left as historical.

### Issue 4 (low): no test for the new layer's precedence

The 15 config tests pass, but none exercise canonical-over-legacy precedence or that `load()` reads `user_config()`. `load()` is filesystem/env-driven, but `cmd::disable::tests::round_trip_through_real_config_load` shows env-driven config tests are feasible here — a precedence test would have caught the init/read path mismatch in Issues 1-2.

---

*Note: the 9 `session_sim` scenario test failures observed locally are environmental (missing corpus/embedding markers), pre-existing on `main`, and unrelated to this change.*

PR #231, opened 2026-07-01
feat(update): native-projection update detection (#7)

aaronsb:
`check-config-updates.sh` detected only **legacy** layouts (git-at-`~/.claude`, or an ADR-140 `.claude-source` marker). A **1.0 native projection** has neither — `~/.claude` is a symlink projection and the app source is a repo at `$XDG_DATA_HOME/agent-ways` — so the session-start 'behind upstream' nudge **silently never fired** for the canonical install.

- **Detector:** new Scenario 0b — `~/.claude` not a repo, no marker, but `$XDG_DATA/agent-ways` is a repo whose origin is upstream → fetch (rate-limited) + behind-count, write `type=native repo=<app>`. Native-fork behind-upstream needs its upstream remote (follow-up; falls through to no nudge, no regression).
- **Reader:** `metrics.rs` `native` case renders `cd "<app>" && git pull && make setup && ways reconcile`.

**Validated end-to-end** on a real native install (this dev box): detector wrote `type=native behind=47`; `ways show core --session …` rendered the correct nudge. `bash -n` + `cargo build` clean.

### DIFF
--- a/hooks/check-config-updates.sh
+++ b/hooks/check-config-updates.sh
@@ -5,6 +5,8 @@
 # Detection order:
 #   0. Is ~/.claude NOT a repo but a .claude-source marker points at one?
 #      → subdirectory projection (ADR-140): check the projecting repo, not ~/.claude
+#   0b. Is ~/.claude NOT a repo, no marker, but $XDG_DATA/agent-ways is a repo?
+#      → native XDG projection (1.0, ADR-142): check the app source, not ~/.claude
 #   1. Is ~/.claude a git repo? If not, exit.
 #   2. Is origin aaronsb/agent-ways? → direct clone
 #   3. Is origin a fork of aaronsb/agent-ways? → fork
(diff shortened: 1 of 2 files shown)


aaronsb:
## Review: native-projection update detection (Scenario 0b)

Reviewed against all six verification points. **No blocking issues** — the implementation is correct and faithfully mirrors the Scenario 0 (subdirectory) design it's modeled on.

### 1. Detection precedence + false-positives — correct
- **(a) Ordered after Scenario 0.** Scenario 0 (`hooks/check-config-updates.sh:107`) fires on `!git ~/.claude && -f SOURCE_MARKER` and always `exit 0`s (even when the inner marker is malformed, `:132`). Scenario 0b additionally requires `[[ ! -f "$SOURCE_MARKER" ]]` (`:144`), so a subdirectory install can never reach or be misclassified by 0b. Double-guarded.
- **(b) `exit 0` at `:158`** is outside the inner remote-match `if`, so 0b never falls through to Scenario 1. Correct — and as noted, in native topology `~/.claude` isn't a repo, so Scenario 1 (`:163`) couldn't match anyway.
- **(c) Legacy in-place clone** (git AT `~/.claude`): first condition `! git -C "$CLAUDE_DIR"` (`:143`) is false → 0b skipped → reaches Scenario 1. Correct.
- **(d) Fork** (origin != upstream): inner `if` false → no cache write → still `exit 0` at `:158`. No nudge, no fall-through. Correct, and the deferral is documented in the block comment.

### 2. Cache write — correct
`write_cache "native" "$BEHIND" "repo=${APP_DIR}" "$FETCH_TS"` matches the signature `type behind extra fetched` (`:53`). PREV_FETCH→FETCH_TS handling (`:149-154`) mirrors Scenario 0 exactly: preserves the prior fetch timestamp when `needs_refresh` is false, so the once-per-hour fetch limit is not reset, while BEHIND is recomputed every run (responsive to local HEAD moves). Consistent.

### 3. behind-count — correct
`git rev-list HEAD..origin/main --count` against APP_DIR, whose origin is validated to equal `aaronsb/agent-ways` (default branch `main`) before this runs. Stale/missing `origin/main` falls back to `0` via `|| echo 0` — acceptable, same as Scenario 0.

### 4. metrics.rs native case — correct
`tools/ways-cli/src/cmd/show/metrics.rs:221` sits below the `if behind == 0 { return }` guard at `:201`, so no nudge at behind 0. Type string `"native"` matches the bash writer exactly. Rendered command `cd "<repo>" && git pull && make setup && ways reconcile` matches `docs/install-guide.md:33` verbatim; `make setup` (Makefile:94) and `ways reconcile` (bare-arg defaults, main.rs:623) both exist. The `&repo` borrow pattern matches the existing subdirectory branch.

### 5. Cache path — matches
Bash `/tmp/.claude-config-update-state-$(id -u)` (`:33`) == Rust `/tmp/.claude-config-update-state-{getuid}` (metrics.rs:150); Windows both `LOCALAPPDATA/.claude-config-update-state`.

### 6. set-safety / quoting — clean
No `set -e`/`set -u` in the hook, so partial git failures don't abort. All `git -C "$APP_DIR"` and variable expansions are quoted; `APP_REMOTE`/`APP_OWNER_REPO` empty-remote case is guarded by the `*github.com*` test.

### Traceability
Backed by ADR-142 (`docs/architecture/system/ADR-142-agent-ways-1-0-xdg-application-distribution.md`) and rendered command matches the install guide. Good.

---

### One LOW-severity note (not blocking — already the documented follow-up)
**Stale cache on a fork→native migration.** In the fork/non-upstream branch, 0b detects native topology but writes no cache and `exit 0`s (`:157-158`). If a prior `~/.claude`-as-repo install had written a `fork` (or `clone`) cache entry, that stale entry survives, and the reader would render a fork nudge whose command (`cd ~/.claude && git fetch upstream && git merge ...`) is wrong once `~/.claude` is a symlink projection. Narrow: requires a specific migration path, and `/tmp` clears on reboot. The block comment already scopes native-fork support to a follow-up, so this is consistent with intent — flagging only so it's captured. A one-line `write_cache "native" "0" ...` (behind 0 → no nudge) in the else path would also clear any stale entry, if you want to close it here.

PR #233, opened 2026-07-01
fix: sweep residual pre-1.0 ~/.claude config-surface references

aaronsb:
## Summary

Follow-up to the #2/#10 doctrine sweeps. A three-agent audit classified **every** `~/.claude` reference in the tree (~450) as 1.0-correct vs stale. Verdict: the Rust path/config core is **fully migrated** — every surviving `cd ~/.claude && git pull` / "repo at ~/.claude" phrasing is correctly fenced inside legacy-migration context. The residue was in shell helpers and docs.

**Live path fixes (broke on native installs):**
- `download-ways.sh` — `cd ~/.claude && make ways` ×4 → `cd "$REPO_ROOT" && make ways` (`~/.claude` has no Makefile in 1.0; `REPO_ROOT` self-locates the app source, works for dev checkout + XDG install alike)
- `commands/ways.md` — `Read ~/.claude/docs/...` → XDG app path (`docs/` is not a projected root — ADR-142 — so the Read failed on native installs)
- `docs/hooks-and-ways/itops.md` — presented `~/.claude/ways.json` as the *current* global domain-disable knob → `disabled_domains` in `$XDG_CONFIG/agent-ways/config.yaml` (legacy `ways.json` noted as still-honored, matching sibling docs)

**Dead-code removal (obsolete, not just stale-pathed):**
- `compare-engines.sh` — BM25-vs-Embedding comparison. Requires the `way-match` binary (removed, ADR-111 folded it into `ways-cli`), compares BM25 (removed, ADR-125 *Accepted*), and reads deleted fixtures. Its premise — comparing two engines — no longer exists.
- `calibrate.sh` — orphaned (zero references), deleted fixtures, superseded by `ways tune` / `ways tune-curves` (ADR-123/134).
- Repointed the `tests/README.md` test-matrix row at `test-embedding.sh` (the live verification `make setup` runs); dropped the dead "compare against BM25" echo in `way-embed/Makefile`.

## What this is NOT
- `settings.json` statusLine (`~/.claude/statusline.sh`) — inert today, resolved by ADR-147; left alone.
- `metrics.rs` legacy-clone update branches — legit (gated on detected install `type`); not touched.
- The `make setup` "rebuilds binaries" claim in `ways-update`/README — that's bug #13 (rebuild), tracked separately, not a config-surface issue.

## Follow-up (not in this PR)
- `hooks/ways/meta/project-health/project-health.md` gates on `when: project: ~/.claude` — dead-for-purpose (framework dev now happens in `$XDG_DATA`, never `~/.claude`). Fixing it needs a new `when: git_remote:` matcher condition — its own change.

## Test plan
- `bash -n` clean on `download-ways.sh`
- Confirmed no lingering references to the deleted scripts
- `docs/scripts/adr lint` unaffected (docs prose only)


aaronsb:
## Review: pre-1.0 config-surface sweep

Scoped verification of the four points this change hinges on. Three pass cleanly; one shell-quoting regression in `download-ways.sh` is worth fixing before merge.

---

### Finding: quoting regression in the `download-ways.sh` echoes (all 4 sites)

**Location**: `tools/ways-cli/download-ways.sh:63, 76, 91, 140`

**Problem**: The replacement is written with nested *unescaped* double quotes:

```sh
echo "  cd "$REPO_ROOT" && make ways" >&2
```

The shell reads `"  cd "` and `" && make ways"` as the two quoted segments and leaves `$REPO_ROOT` **unquoted** in between. So the intended `cd "$REPO_ROOT"` protection is silently dropped. Demonstrated:

```
# as written (space in path)
$ REPO_ROOT="/home/user/my project/agent-ways"
$ echo "  cd "$REPO_ROOT" && make ways"
  cd /home/user/my project/agent-ways && make ways     # <- quotes gone

# escaped form (intended)
$ echo "  cd \"$REPO_ROOT\" && make ways"
  cd "/home/user/my project/agent-ways" && make ways
```

**Why it matters**:
- The printed guidance loses the quotes around the path, so a user with a space in their app-source path copy-pastes a broken `cd`.
- `$REPO_ROOT` is unquoted at echo time, so it is subject to word-splitting and (no `set -f` is in effect) glob expansion — a path containing a space or a glob metacharacter mangles the emitted line.
- `bash -n` passes because the line is syntactically valid, which is why the test plan didn't catch it. This is a semantic quoting bug, not a syntax error.

**Suggestion**: escape the inner quotes at all four sites:

```sh
echo "  cd \"$REPO_ROOT\" && make ways" >&2
```

This both preserves the quotes in the user-facing output and keeps the expansion safely double-quoted.

---

### Verified clean

**(a) `$REPO_ROOT` scope** — PASS. Defined unconditionally at line 25 (`REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"`), well before all four edited echoes (63/76/91/140). No early-exit path references it before definition.

**(b) Deleted scripts unreferenced** — PASS. The pre-change tree carried exactly two references (`tests/README.md` matrix row + the `way-embed/Makefile` echo), and this PR removes both. `calibrate.sh` had zero references. No CI/workflow or other doc invokes either script. Post-merge the tree is clean.

**(c) `tests/README.md` repoint** — PASS. `way-embed/Makefile`'s `setup` target runs `bash "$(SRCDIR)test-embedding.sh"`, so `test-embedding.sh` is genuinely the live verification the matrix row should point at.

**Docs** (`commands/ways.md` XDG resolution, `itops.md` `disabled_domains`/`config.yaml` repoint with the legacy-`ways.json`-still-honored note, ADR-147, INDEX.md row) read correctly and are consistent with the 1.0 XDG/projection model.

---

**Assessment**: Substantively sound sweep — the dead-code removals are fully de-referenced and the doc repoints are accurate. The one blocker is the nested-quote regression in the four `download-ways.sh` echoes: trivial to fix (escape the inner quotes), but it defeats the exact path-quoting the change set out to get right, so worth correcting before merge.

---

PR #234, opened 2026-07-01
refactor: relocate project-pulse/project-health dev harness out of the operator surface

aaronsb:
## Summary

Follow-on from the `~/.claude` config-surface sweep (#233). While auditing project-pulse/project-health, the real problem surfaced: they're **dev harness leaking into the operator product surface**, not stale paths. They're maintainer tools for developing agent-ways itself (track Claude Code upstream vs agent-ways' commits, reconcile agent-ways' own ADRs), yet they sat in projected roots (`skills/`, `hooks/ways/`) — so they symlinked into every operator install. An operator has no use for either.

The tell: `project-health` gated on `when: project: ~/.claude`, which pre-1.0 *meant* "only when developing agent-ways" (because `~/.claude` **was** the repo then). 1.0 broke the gate — but its intent was always dev-only.

### Two commits

**1. `.gitignore`: pre-1.0 deny-all overlay → conventional.** The deny-all-then-allowlist posture was a fossil of when the repo *was* an in-place overlay on a live CC install (it had to ignore the install's runtime state). In 1.0 the repo is a normal app that *projects into* `~/.claude` (ADR-142), so the deny-all is obsolete. Flipped to conventional "track by default, ignore build/runtime junk." **Verified:** tracked set unchanged (620 files); `target/`, `.claude` runtime, `__pycache__`, `bin/`, build trees, generated corpus all still ignored; llama.cpp submodule untouched; no junk newly surfaced. Un-ignores `.claude/ways/` + `.claude/skills/` so the repo can track its own project scope.

**2. Relocate the dev harness to project scope.**
- `skills/project-pulse/` → `.claude/skills/project-pulse/` (script now resolved via `CLAUDE_PROJECT_DIR`)
- `hooks/ways/meta/project-health/` → `.claude/ways/meta/project-health/` — **drops the broken `when: project:` gate entirely** (a project-scoped way fires only in its own project by construction, so the planned `when: git_remote:` matcher condition is no longer needed)
- `scripts/project-pulse` unchanged (already non-projected)

agent-ways' **first dogfooding of its own ADR-143 project tier**. Recorded as **ADR-148 (Draft)**: *the projected framework surface ships operator content only; agent-ways' own dev harness lives in project scope.*

## Notes
- ADR-148 is numbered around #232's ADR-147 (which isn't on main yet).
- ADR-106 (original project-pulse) linked as related, left unedited (historical record).
- Caveat in the ADR: a project-scoped *semantic* way needs the project corpus to include project-local ways to fire; the *skill* (`/project-pulse`) works immediately via native discovery.

## Test plan
- `.gitignore` flip verified against tracked baseline (see above)
- `docs/scripts/adr lint` clean for ADR-148
- No dangling references to old locations except ADR-106 (historical, linked)


aaronsb:
## Review: independent verification of the two commits

Checked out the branch and verified the description's claims against the repo rather than taking them on faith. Headline: **the mechanics all hold up** — the relocation is clean, no dangling references, no ADR collision, and the `.gitignore` correctly ignores every junk category tested while un-ignoring project scope. The one thing worth acting on is a *safety-by-default* regression that's inherent to the deny-all → conventional flip.

### Verified ✓

**(a) `.gitignore` — junk still ignored / project scope surfaced.** `git check-ignore` confirms all of these are ignored: `tools/target/x`, `.claude/sessions/x`, `foo.pyc`, `bin/ways`, `tools/way-embed/build/x`, `ways-corpus.jsonl` (root *and* `hooks/ways/`), plus `tools/attend/target`, `tools/ways-cli/target`, `.claude/state`. `.claude/ways/foo.md` and `.claude/skills/foo/SKILL.md` are correctly **not** ignored, and the `tools/way-embed/llama.cpp` submodule path is **not** ignored. The generated-corpus glob `ways-corpus*.jsonl` matches the real filenames (`ways-corpus-en.jsonl`, `ways-corpus-multi.jsonl` per `scripts/signal-report.py`). 

**Tracked-set delta is exactly the intended set — nothing unexpected surfaced.** `comm` of `git ls-tree main` vs the branch: **+3** (`.claude/skills/project-pulse/SKILL.md`, `.claude/ways/meta/project-health/project-health.md`, `ADR-148`), **−2** (the two old locations). One wording nit: the count is **621**, not 620 — ADR-148 is net-new (the two moves are net-zero). The *claim behind* the number (the flip surfaced no unexpected files) is verified true.

**(b) skill script resolution.** `${CLAUDE_PROJECT_DIR:-$PWD}/scripts/project-pulse` is the idiomatic resolution used throughout `tools/ways-cli/src` (session.rs, scan/scoring.rs, list.rs, init.rs all resolve project scope via `CLAUDE_PROJECT_DIR`). `scripts/project-pulse` is still tracked at repo root and non-projected. Correct. Only edge: the `:-$PWD` fallback is wrong if `CLAUDE_PROJECT_DIR` is unset *and* cwd isn't repo root — but Claude Code always exports it, so this is theoretical.

**(c) dropped `when:` gate.** Confirmed project ways load exclusively via `CLAUDE_PROJECT_DIR/.claude/ways/` (`session.rs:556` — *"Project-scope only — sourced exclusively from {project}/.claude/ways.yaml"*). So a project-scoped way is project-gated *by construction*; removing `when: project: ~/.claude` doesn't broaden scope. The ADR's caveat about the semantic way needing the project corpus to include project-local ways to actually *fire* is honest and correct — worth keeping visible since the way is otherwise silently inert until the project corpus picks it up.

**(d) no stranded references.** Grep across `*.md/*.sh/*.py/*.rs/*.json/*.jsonl/*.yaml/*.toml` for the old paths returns only ADR-148's own move-description lines (the `→` arrows). No code, corpus, test, or doc still points at `skills/project-pulse` or `hooks/ways/meta/project-health`. Clean.

**(e) ADR numbering.** No ADR-147 *or* 148 on `main`; 147 is correctly absent from `INDEX.md` (it's on #232's branch). No on-main collision.

### Finding — safety-by-default regression (non-blocking, worth a proactive line)

The flip's real cost: the old `*`-then-allowlist ignored **everything not explicitly tracked**, so stray secrets/junk were ignored *by default*. The conventional list inverts that to track-by-default, and it doesn't yet cover several common categories. `git check-ignore` confirms these are all **NOT** ignored now:

- `.env`, `.env.local`, `secrets.env` — the security-relevant one
- `*.log`, `.venv/`, `venv/`, `.pytest_cache/`, `.mypy_cache/`, `.ruff_cache/`, `.coverage`, `coverage/`
- `node_modules/`, generic `dist/`, `build/`
- `*.orig`, `*.rej` (conflict leftovers)

Practical risk today is low — no `package.json` and no `pyproject/requirements` are tracked, so node/python-venv output won't appear. But `.env`/secrets is worth adding proactively: a maintainer dropping a `.env` in the checkout is now one `git add -A` away from committing it, whereas the old posture caught it silently. Suggest a small block:

```gitignore
# --- Secrets / local env ---
.env
.env.*
# --- Python virtualenvs / tool caches ---
.venv/
venv/
.pytest_cache/
.mypy_cache/
.ruff_cache/
.coverage
# --- Misc ---
*.log
*.orig
*.rej
```

### Minor / forward-looking

- `/.claude/*` + `!/.claude/ways/` + `!/.claude/skills/` un-ignores only those two roots. If a project-scoped *command* or *agent* is ever added under the repo's `.claude/`, it'd be silently ignored. Matches ADR-148's deliberate scoping (ways + skills only), so fine now — just a thing to remember when the project tier grows.

### Assessment

The reasoning in ADR-148 is sound — the `when: project: ~/.claude` gate really was a pre-1.0 dev-only tell that the projection model broke, and moving the harness into project scope (dogfooding ADR-143) is the right correction. Mechanics verified end-to-end. Not blocking. The only thing I'd fold in before this becomes the repo's steady-state `.gitignore` is a proactive `.env`/secrets ignore, since the flip removes the implicit net that used to cover it.

PR #235, opened 2026-07-01
fix(settings-merge): idempotent re-apply when the persisted base under-records

aaronsb:
## Summary

Fixes the recurring `settings merge changed unmanaged fields — reverted` that fired on **every** `ways reconcile` (tracked as #12). Earlier notes guessed a `permissions.allow` reorder — the real cause is **hooks**, and it's self-perpetuating.

### Root cause
A settled install stores our hook commands **quoted** (`"${HOME}/.claude/bin/ways" init` — Windows-space safety). But the merge dedup'd `theirs` against the **raw (unquoted)** `ours`. When the persisted base under-records what we own — on this box its `hooks` map was empty while `perms` was fine — the filter matches nothing, so:
1. every hook is **duplicated** → `merged != live` → a write happens
2. the self-audit strips our contribution using the **empty base** for the before-view but the **correct new base** for the after-view → asymmetry → **revert**
3. it reverts *before* persisting the corrected base → the base never heals → fails forever

### Fixes
- **merge**: dedup `theirs` against the **quoted** form we actually write, so a lost/empty base can't cause duplication — re-apply collapses to a true no-op.
- **self-audit**: strip the pre-write view with the **union** of prior + new base, so an under-recording base can't make `live` look like it holds user hooks that `after` correctly drops.

### Validation
- New repro test `reapply_is_idempotent_when_base_hooks_were_lost` (base file exists with empty hooks + quoted live hooks) — fails before, passes after.
- All 9 `settings_merge` unit tests + 186 `ways` unit tests green. (`session_sim` integration failures are pre-existing on `main` — unrelated marker-pipeline/env issue.)
- **Live on this box**, where it reverted every reconcile: now reports `settings.json already up to date`, idempotent across repeated runs, and settings.json keeps both the user's `Bash(git:*)` and our `Bash(ways:*)`.

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml settings_merge`
- `ways reconcile` twice → second run is a clean no-op

### DIFF
--- a/tools/ways-cli/src/cmd/settings_merge.rs
+++ b/tools/ways-cli/src/cmd/settings_merge.rs
@@ -106,13 +106,15 @@ pub fn merge(live: &Value, desired_hooks: &Value, base: &Owned) -> Result<Merged
         // stores our hooks quoted, so comparing only against raw `ours` fails to
         // recognize them: with an under-recording base (e.g. one whose hooks were
         // lost) we'd then duplicate every hook on re-apply and trip the self-audit.
+        // NB: this byte-match assumes `quote_entry_commands` stays in lockstep with
+        // the jq quoting the hook-install path applies (both quote the first token).
         let mut merged: Vec<Value> = theirs
             .into_iter()
             .filter(|e| {
                 !base_entries.contains(e) && !ours.contains(e) && !ours_quoted.contains(e)
             })
             .collect();
-        merged.extend(ours_quoted.iter().cloned());
+        merged.extend(ours_quoted);
 
         if !merged.is_empty() {
             new_hooks.insert(event.clone(), Value::Array(merged));
--- a/tools/ways-cli/src/cmd/settings_merge.rs
+++ b/tools/ways-cli/src/cmd/settings_merge.rs
@@ -98,17 +98,21 @@ pub fn merge(live: &Value, desired_hooks: &Value, base: &Owned) -> Result<Merged
         let base_entries =
             base.hooks.get(event).and_then(|v| v.as_array()).cloned().unwrap_or_default();
         let ours = ours_hooks.get(event).and_then(|v| v.as_array()).cloned().unwrap_or_default();
+        // The form we actually write (command paths quoted, Windows-space safety).
+        let ours_quoted: Vec<Value> = ours.iter().map(quote_entry_commands).collect();
 
         // User entries = theirs minus what we wrote last (base) minus what we're
-        // about to write (ours) — so a re-apply doesn't duplicate.
+        // about to write — matched in BOTH raw and quoted form. A settled install
+        // stores our hooks quoted, so comparing only against raw `ours` fails to
+        // recognize them: with an under-recording base (e.g. one whose hooks were
+        // lost) we'd then duplicate every hook on re-apply and trip the self-audit.
         let mut merged: Vec<Value> = theirs
             .into_iter()
-            .filter(|e| !base_entries.contains(e) && !ours.contains(e))
+            .filter(|e| {
+                !base_entries.contains(e) && !ours.contains(e) && !ours_quoted.contains(e)
+            })
             .collect();
-        // Append ours, with hook command paths quoted (Windows-space safety).
-        for e in &ours {
-            merged.push(quote_entry_commands(e));
-        }
+        merged.extend(ours_quoted.iter().cloned());
 
         if !merged.is_empty() {
             new_hooks.insert(event.clone(), Value::Array(merged));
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review — settings-merge idempotency fix

**What this changes**: two coordinated fixes so a re-apply against a base that under-records our owned hooks stops reverting forever — (1) `merge` dedups `theirs` against `ours_quoted` as well as raw `ours`; (2) the self-audit strips the pre-write `live` view using `union_owned(&base, &merged.base)` instead of `&base` alone; plus the `union_owned` helper and a repro test.

**Assessment**: correct and well-scoped. I verified each concern against the code rather than the description.

### (a) Dedup change — no wrongful user-hook loss
`ours_quoted` is a fresh `Vec` (`ours.iter().map(quote_entry_commands)`); `quote_entry_commands` clones the entry, so `ours` stays raw and `!ours.contains(e)` still means what it says. The new `!ours_quoted.contains(e)` can only drop a `theirs` entry that is *byte-identical* to the quoted form of one of ours — i.e. effectively our own hook — and it's immediately re-appended via `ours_quoted`, so exactly one copy survives. No user data is lost even in the pathological "user independently authored our hook" case. `quote_first_token` edge cases (already-quoted → untouched; no space → whole-token quote; multiple spaces → quote-to-first-space) are all idempotent, so re-quoting a stored base is stable.

### (b) `union_owned` does NOT weaken the audit (the important risk)
The union is a *superset* strip-set for `before`, so the question is whether it can strip genuinely-user content and mask a real clobber. It can't: `union_owned` only adds `merged.base` over `base`, and `merged.base` is derived purely from our shipped `desired_hooks` (quoted) + the fixed `WAYS_PERMS`. Every entry it adds is ours by construction; the only user entry it could touch is one byte-identical to a quoted-ours hook or equal to a `WAYS_PERMS` string — both already indistinguishable from ours and stripped symmetrically on the `after` side too. A genuine unrelated user field is never in the union strip-set, so a real clobber still surfaces as `before != after`. Note: the pre-existing masking directions (an *over-recording* base that mis-records a user hook as ours; a deprecated perm the user also wants) are unchanged by this PR — they already followed from using `&base` for the before-view, and are inherent to three-way ownership, not introduced here.

### (c) Idempotency
Settled installs append ours *last*, so `[user…, ours_quoted]` re-merges to the identical order → `merged.settings == live` → early return, a true no-op. One nuance worth stating plainly: the *first* post-bug reconcile (lost/empty base, or ours stored non-suffix) is **not** a no-op — the merge canonicalizes order to `[user…, ours_quoted]`, so it writes once, passes the audit, and heals the persisted base; *subsequent* runs are no-ops. So "collapses to a true no-op" is true steady-state but the healing run still writes once. Correct and convergent, just not zero-write on the transition.

### (d) Repro test genuinely gates the fix
`reapply_is_idempotent_when_base_hooks_were_lost` exercises exactly the fixed path and needs *both* fixes:
- Without the dedup fix: quoted `theirs` entry doesn't match raw `ours`, so it's kept *and* re-appended → `SessionStart` len 3, tripping `assert_eq!(ss.len(), 2)`; and with an empty base the audit strip is asymmetric → `apply_to_files` bails → the `.expect(...)` panics first.
- Without the union fix (dedup only): no duplication, but empty-base before-view retains our quoted hook while after drops it → audit aborts → `.expect(...)` panics.
Both fixes present → green. Good coverage of the interaction.

### (e) No residual `permissions.allow` reorder trips the audit
Merge filters user perms stably then appends `WAYS_PERMS` as a suffix, and `stripped_user_view` preserves user-perm relative order, so interleaved / non-suffix user perms strip to the same sequence on both sides. Worth calling out that `union_owned` *also* fixes the perms-lost-base case symmetrically (empty `base.perms` ∪ `WAYS_PERMS` → correct before-view), which the PR description under-sells — it frames the fix as hooks-only but it's the same helper covering perms. Non-canonical perm ordering canonicalizes with a single healing write, then no-ops.

### Minor (non-blocking)
- `merged.extend(ours_quoted.iter().cloned())` clones the whole vec needlessly; `ours_quoted` is unused afterward, so `merged.extend(ours_quoted)` (move) is equivalent and cheaper.
- The dedup relies on `ours_quoted` byte-matching the *stored* quoted form. The Rust path is now self-consistent (writes and dedups via the same `quote_first_token`), but installs whose live hooks were last written by the jq transform depend on that transform quoting identically ("Mirrors the jq transform"). Worth a one-line note/assert somewhere that the two quoting paths stay in lockstep, since a divergence would silently reintroduce duplication.

**Verdict**: no correctness or safety-net regression. The union broadens the strip-set only over provably-owned content, the dedup can't lose user hooks, and the test meaningfully gates both changes. `cargo test settings_merge` → 9/9 green locally. Good to merge.

---

PR #236, opened 2026-07-01
fix(update): document force-rebuild recipe for native updates

aaronsb:
## Summary

Fixes #13 — the documented native-update recipe left users on a **stale binary**. `make setup` short-circuits any binary that already exists and runs (`ways already installed`), so `git pull && make setup && ways reconcile` pulled new source but never rebuilt the compiled binary. (Confirmed live during a machine update: `make setup` printed *"ways already installed"* / *"attend already built"* and skipped the rebuild.)

The infrastructure already existed — `make update` (pull + `update-binaries` force-rebuild + relink + corpus regen) and `make update-binaries` (force-rebuild standalone). The docs just pointed at the wrong verb.

## Surgical, not a blanket swap
Only genuine **update** recipes change. The many `make setup` references in **acquire/repair** contexts are correct and left alone — fresh install, missing/broken-engine repair (`make setup` rebuilds a binary that doesn't run, and downloads a prebuilt when absent, whereas `update-binaries` forces a source build needing cargo), `finish-install`, the deployment way (already uses `make update`).

## The recipes
- **Direct install:** `make update && ways reconcile` (robust pull that autostashes around machine-local settings.json drift, force-rebuild, corpus regen, then reproject).
- **Fork / surfaces that already pull** (the `ways-update` skill does its own guarded pull): `make update-binaries` + `ways reconcile`. The corpus self-heals on the next session via the `ways corpus --if-stale` `SessionStart` hook.

## Surfaces changed
- `README.md` — update section + the sync-command table (native + fork rows)
- `docs/install-guide.md` — 1.0 update + fork update
- `skills/ways-update/SKILL.md` — the flow block + step 3 (was already naming `update-binaries` as a fallback; now it's the primary)
- `tools/ways-cli/src/cmd/show/metrics.rs` — the `native` branch, matched to the `clone`/`fork` branches that *already* used `make update` / `make update-binaries`

## Test plan
- `cargo build` + `cargo test metrics` green
- Verified no stale `make setup` remains in the update recipes (only intentional "use `make update`, not `make setup`" explanatory prose)

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,17 +287,17 @@ For the attention mechanics: [context-decay.md](docs/hooks-and-ways/context-deca
 In 1.0 the app source lives in `$XDG_DATA_HOME/agent-ways` (not `~/.claude`). Update by refreshing that checkout, then rebuilding and reprojecting — or just re-run the installer one-liner, which is idempotent:
 
 ```bash
-cd "$XDG_DATA_HOME/agent-ways" && git pull && make setup && ways reconcile
+cd "$XDG_DATA_HOME/agent-ways" && make update && ways reconcile
 ```
 
-`make setup` rebuilds the binaries and semantic-matching corpus; `ways reconcile` refreshes the `~/.claude` projection. A fork fetches and merges upstream in the app dir first (`git fetch upstream && git merge upstream/main`), then the same `make setup && ways reconcile`.
+`make update` pulls (robustly — it autostashes around machine-local settings drift), **force-rebuilds** the binaries, regenerates the corpus, and relinks; `ways reconcile` then refreshes the `~/.claude` projection. Use `make update`, not `make setup` — `setup` skips binaries that already exist, so on an update it would leave you on the *old* binary. A fork fetches and merges upstream in the app dir first (`git fetch upstream && git merge upstream/main`), then `make update-binaries && ways reconcile` (force-rebuild without re-pulling origin; the corpus self-heals via the `SessionStart` hook).
 
 At session start, `check-config-updates.sh` flags when you're behind upstream (`aaronsb/agent-ways`), rate-limited to once per hour. It currently recognizes **legacy** git-based layouts — an in-place clone at `~/.claude`, and the ADR-140 subdirectory (`.claude-source` marker). Native-projection update detection is being wired (it reads the app source in `$XDG_DATA`); until then, use the manual command above.
 
 | Scenario | How detected | Sync command |
 |----------|-------------|--------------|
-| **Native projection (1.0)** | app source at `$XDG_DATA_HOME/agent-ways` | `cd "$XDG_DATA_HOME/agent-ways" && git pull && make setup && ways reconcile` |
-| **Fork** | GitHub API reports `parent` is `aaronsb/agent-ways` | fetch/merge upstream in the app dir, then `make setup && ways reconcile` |
+| **Native projection (1.0)** | app source at `$XDG_DATA_HOME/agent-ways` | `cd "$XDG_DATA_HOME/agent-ways" && make update && ways reconcile` |
+| **Fork** | GitHub API reports `parent` is `aaronsb/agent-ways` | fetch/merge upstream in the app dir, then `make update-binaries && ways reconcile` |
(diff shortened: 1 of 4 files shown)


aaronsb:
## Review — verified against the Makefile, update.sh, settings.json, and the tree

Reviewed the four changed surfaces against the actual build recipes. The fix is correct, surgical, and the reasoning in the PR body holds up on every claim I checked. Approving on substance; notes below.

### (a) Recipe correctness — confirmed
- `update:` (Makefile:155) runs `scripts/update.sh` (pull) → `update-binaries` → `install`. `install: hooks-executable setup hooks-install`; `setup` (Makefile:94) re-depends on `ways attend attend-chat` (already fresh, so they short-circuit), then runs way-embed setup and `ways corpus --quiet`. So `make update` = **pull + force-rebuild + corpus regen + PATH relink**, and it does **not** run `ways reconcile` — the trailing `ways reconcile` is genuinely required. ✓
- `update-binaries: ways-rebuild attend-rebuild attend-chat-rebuild way-embed-rebuild` (Makefile:178) is a **pure force-rebuild** — no pull, no corpus regen, no reconcile. So `make update-binaries && ways reconcile` correctly leans on the corpus self-heal (see c). ✓
- `ways-rebuild` (Makefile:202) relinks `bin/ways` after the in-place cargo build, so the `XDG_BIN/ways → bin/ways → target/release/ways` chain stays valid without a `make install` — the fork recipe omitting `make install` is fine. ✓

### (b) acquire-vs-update judgment — confirmed, nothing missed
Grepped every `make setup` in the tree and classified the untouched ones. All remaining references are fresh-install / repair / model-fetch / historical, not update recipes:
- `hooks/ways/check-setup.sh` (32/52/86) — missing/broken-engine repair. Correct: `setup` downloads a prebuilt when the binary is absent; `update-binaries` forces a source build needing cargo.
- `docs/finish-install.md` (51) and `tools/way-embed/RELEASE.md` (8/47) — download-prebuilt-and-fetch-model recipes, not updates.
- `scripts/install.sh`, the `*.rs` error hints (status/tune/match/corpus/scan/language), test SKIP hints, ADRs, model-fetch prose — all acquire/repair context.
No genuine update/refresh recipe still points at `make setup`. The "surgical, not a blanket swap" framing is accurate.

### (c) Corpus self-heal — confirmed
`settings.json:133` SessionStart (`matcher: "startup"`) includes `${HOME}/.claude/bin/ways corpus --if-stale --quiet` (line 159). The hook exists and runs exactly the command the PR claims, so the fork/skill recipe omitting an explicit corpus regen is sound. (It fires on `startup` only, which is the relevant post-update path.)

### (d) metrics.rs — compiles, consistent
- The added `out.push_str("…")` (line 228) is a plain literal with no `{}` — no format-arg mismatch; the interpolated line above it uses `format!("…{repo_disp}…")` with `repo_disp: &str`. Clean. ✓
- Consistency with siblings is topology-appropriate rather than identical: `clone` uses `make update`; `fork`/`renamed_clone` (legacy in-place) use `make update-binaries && make install`; `native` uses `make update && ways reconcile`. The reproject-via-`ways reconcile` vs `make install` split is correct — native is the projection topology. The PR only touched `native` and matched the verb pattern the other arms already used. ✓

### (e) Fork nuance — confirmed
`update.sh:32` is `git pull --ff-only`, i.e. it pulls **origin**. For a fork, origin is the fork, so `make update` would pull the wrong remote. All three fork recipes (README table + prose, install-guide, ways-update SKILL) correctly use a manual `git fetch upstream && git merge upstream/main` then `make update-binaries` — never `make update`. ✓

### Minor (non-blocking)
- `skills/ways-update/SKILL.md` — "(binaries are already fresh, so it only rebuilds way-embed and regenerates the corpus)" is slightly loose: `make setup`'s way-embed target can also short-circuit if the binary is present; what it reliably does is regenerate the corpus. The load-bearing claim (eager corpus regen) is correct, so this is just prose precision.

### Assessment
Root cause (setup short-circuits existing binaries → stale binary after pull) is correctly diagnosed, and the fix repoints only real update recipes while leaving acquire/repair paths on `make setup` for the right reason. Recipes are complete and topology-correct on both the direct and fork paths, the self-heal hook backing the fork path actually exists, and the Rust change is sound. No blockers.

*AI-assisted review via Claude Code.*

PR #239, opened 2026-07-01
feat(settings): ways settings lint — config-fragment linter (ADR-147 slice 1)

aaronsb:
## Summary
First implementation slice of ADR-147 (composable settings.json config fragments): the **authoring gate**. `ways settings lint <dir>` reads a tree of `NN-*.md` fragments (YAML `settings:` block + rationale body) and reports three failure classes at author time — the thing the enterprise console's raw textarea cannot do.

- **fragment loader** (`cmd/settings/fragment.rs`) — parses `NN-*.md` frontmatter, alphabetical-by-filename to match Claude Code's `managed-settings.d` merge order; `settings:` → `serde_json::Value`.
- **curated schema table** (`cmd/settings/schema.rs`) — top-level key → (ScopeClass, ValueType). Unknown key → warning, never error (CC version-gates ~90 keys). Managed-only list seeded from the ADR-147 interop research.
- **three checks** (`cmd/settings/lint.rs`) — schema-valid (type mismatch → error, unknown → warn), scope-legal (managed-only at user/project → error; managed-overridable outside managed → warn), duplicate-scalar (→ warn; objects/arrays excluded as non-lossy). Human + `--json`; exits 1 on errors.

Named `ways settings` (not `config`): `ways config` already manages the ways runtime config, and CC calls this surface "settings" anyway. Resolves ADR-147 open-Q2 direction; Q1/Q3/Q4 remain open (store path is an argument, so Q1 stays unbaked).

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml` — 209 unit (incl. 23 settings) + 7 project_ways + 4 settings_lint integration, all green.
- Fixture stores under `tests/fixtures/settings-store/{clean,dirty}` drive the real binary end-to-end (findings + exit code + `--json`).
- Pre-existing `session_sim` failures are environment-dependent (need the embedding corpus) and reproduce identically on clean `main` — unrelated to this change.


aaronsb:
## Review — ADR-147 slice 1: `ways settings lint`

Reviewed against the ADR's stated semantics, with focus on the three checks, the frontmatter splitter, the YAML→JSON fidelity boundary, and reachable panics. This is a clean, well-documented slice — module boundaries are crisp, the loader/schema/lint split is right, and the fidelity boundary is tested honestly. Findings ranked by severity.

### High — `deniedMcpServers` is classed `ManagedOnly`, contradicting the ADR's own interop section

`schema.rs:104` puts `deniedMcpServers` in the `ManagedOnly` bucket:

```rust
| "allowedMcpServers" | "deniedMcpServers" => (ManagedOnly, Any),
```

But ADR-147 "Managed-scope interop" (lines 129–130) names `deniedMcpServers` as the canonical *list key that concatenates* and explicitly states user-scope fragments "still take effect — a user may broaden a managed allowlist… not dead under management."

Consequence: a user-scope fragment setting `deniedMcpServers` produces a scope-legal **ERROR** — "at user scope Claude Code ignores it" — which is the exact opposite of what the ADR says happens, and it flips the exit code to 1. For a linter whose value prop is "don't reject valid config" and "be honest about managed interop," this is a false-positive error on legitimate config.

Suggestion: `deniedMcpServers` should be `Normal` (or a new list-concat class), not `ManagedOnly`. Also re-check `allowedMcpServers` — the ADR only names `deniedMcpServers` + `permissions.allow/deny` for concatenation, so `allowedMcpServers` may be genuinely managed-only, but it's worth confirming against the same interop source rather than pairing it with the denied list by reflex. The unit test at `schema.rs:133` currently pins the wrong classification, so the test suite is green *because* it encodes the bug.

### Medium — duplicate-scalar is not partitioned by scope → false-positive "dropped" warnings

`lint.rs` `check()` builds one global `seen: HashMap<&str, &Path>` across every fragment regardless of scope. But user / project / managed fragments compile to *different destination artifacts* (`~/.claude/settings.json`, project `settings.json`, `managed-settings.d/*.json`). A scalar like `model` set once at `scope: user` and once at `scope: project` does not "last-win / drop the earlier value" — the two land in different files and both apply (runtime precedence, not compile-time loss). The current code emits a duplicate warning saying the earlier value "is dropped," which is misleading across scopes.

Suggestion: key the `seen` map on `(scope, key)` so duplicate-scalar only fires within a single compilation target — matching the ADR's "last-wins by filename order" framing, which is a within-one-settings.json property.

### Low — 3+ occurrences of a scalar always name the *first* fragment, and `seen` isn't updated

In `check()`'s duplicate loop, once a key is in `seen` the entry is never replaced. With the same scalar in three fragments, both later fragments warn "also set in <first>", when the immediately-preceding fragment is the more accurate "earlier value" being dropped. Minor diagnostic accuracy; consider updating `seen` to the latest path on each hit.

### Low — `.map()` used for its side effect (clippy `option_map_unit_fn`)

`lint.rs`:

```rust
scope_finding(spec.class, frag.scope, key, &file)
    .map(|f| findings.push(f));
```

`Vec::push` returns `()`, so this is `Option::map` driven for effect — `clippy::option_map_unit_fn` flags it by default. Prefer `if let Some(f) = scope_finding(...) { findings.push(f); }` for consistency with the rest of the crate.

### Low — `json_kind` duplicated across `fragment.rs` and `lint.rs` with divergent wording

Two `json_kind` helpers exist; one says `"a list"` / `"a mapping"`, the other `"an array"` / `"an object"`. Same concept, two strings a user sees in different diagnostics. Consider one shared helper (or at least aligned wording).

### What holds up well

- **Frontmatter splitter (`split_frontmatter`)** — I traced CRLF, empty frontmatter (`---\n---\n`), unclosed block, `---` appearing inside the body, and BOM. All handled correctly: the closer requires both line-start and line-end, so body horizontal rules are preserved; CRLF openers/closers work; the lone `unwrap()` on `find('\n')` is genuinely unreachable because `starts_with("---\n")` already guarantees a newline. No reachable panic on malformed input reaching the CLI — good. One nit: a `----` line won't match as a closer (falls through to "not closed"), which is acceptable but undocumented.
- **YAML→JSON fidelity (yes/no)** — handled honestly and tested at both layers: `yaml_yes_no_are_strings_not_booleans` pins the loader behavior, and `yaml_yes_becomes_schema_error_on_bool_key` proves the end-to-end path where `autoUpdates: yes` surfaces as a schema error. The doc comment on `FrontMatter` calls the caveat out explicitly. This is the right way to handle it.
- **unknown-key = warning, managed-only at user/project = error, managed-overridable outside managed = warning, duplicate only for scalars** — all four match the ADR, and each has a direct unit test. The `Any` value-type escape hatch to avoid false type-positives on uncertain keys is a sound call.

Net: one data-classification bug that inverts the ADR (deniedMcpServers), one scope-partitioning gap in duplicate detection, and a few idiom nits. The splitter and fidelity handling are solid.

---
*Reviewed by code-reviewer (AI-assisted). Strictly advisory — no code modified.*

aaronsb:
Review addressed in 7290cf1: reclassed deniedMcpServers/allowedMcpServers to Normal (the ADR-147 interop bug — thanks, this was the right block) + regression test; scope-partitioned duplicate-scalar on (scope,key) with most-recent predecessor; Option::map→if let; unified json_kind. 26 unit + 4 integration green, clippy clean.

PR #240, opened 2026-07-01
feat(settings): vendored schema + configurable source + scaffold (ADR-147 slice 2)

aaronsb:
## Summary
Second ADR-147 slice: acquire Claude Code's config surface deterministically and turn it into a fill-in-the-blank authoring aid.

- **Vendored schema** — no official Anthropic machine-readable schema exists (anthropics/claude-code#11795); the native binary embeds keys only as minified strings. We vendor the community **SchemaStore** schema (`json.schemastore.org/claude-code-settings.json`, 84 keys, all with descriptions) as a repo file bundled via `include_str!` (offline, lockfile pattern). `schema_doc.rs` parses it to key → {type, description, placeholder}.
- **Configurable source** (per request) — the refresh *source* is a configuration surface, not a constant: `source.rs` resolves `$WAYS_SETTINGS_SCHEMA_URL` → ways config `settings_schema_url` → SchemaStore default. `ways settings schema` surfaces it (`--source` prints the bare URL); `refresh-settings-schema.sh` honors the same resolution.
- **`ways settings new <key>`** — scaffolds a valid `NN-<key>.md` with a type-correct placeholder and the schema description pre-filled in the body. Only schema-known keys scaffold; managed-only keys default to `managed` scope; won't clobber; NN auto-increments.
- **Lint consolidation** — schema-valid now sources the key set + types from the vendored schema (all 84 keys), and `schema.rs` shrinks to *only* the scope-class overlay (which also covers 5 keys SchemaStore still lags, e.g. `fallbackModel`, `disableSideloadFlags`).

Records the schema-source decision in ADR-147.

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml` — 221 unit + 7 settings integration, all green; clippy clean.
- Round-trip guaranteed by test: `ways settings new` → `ways settings lint` clean.
- Pre-existing `session_sim` failures are environment-only (need the embedding corpus), unchanged on this branch.

### DIFF
--- a/docs/architecture/system/ADR-147-composable-settings-json-config-fragments.md
+++ b/docs/architecture/system/ADR-147-composable-settings-json-config-fragments.md
@@ -97,6 +97,25 @@ Let Claude run git/gh unprompted — constant use, prompts are pure friction.
   into `~/.claude` and three-way-merges `settings.json`, using the provenance manifest
   as its base (the same base hardened in the ADR-145 settings-merge work).
 
+### Schema source
+
+The lint and compile stages need Claude Code's settings schema. Anthropic does
+not publish a machine-readable one (anthropics/claude-code#11795), and the native
+binary embeds keys only as minified string literals — not cleanly extractable.
+The community **SchemaStore** schema
+(`json.schemastore.org/claude-code-settings.json` — 84 keys, every one carrying a
+`description`) is the de-facto definition editors already use for autocomplete, so
+we adopt it as the shape source: **vendored** — pinned and bundled into the binary
+(`include_str!`), offline, the lockfile pattern — not fetched at runtime. It
+supplies the key set, types, and descriptions (enough to generate fill-in-the-blank
+fragment templates); it deliberately does *not* encode scope-class (managed-only /
+managed-overridable), which stays the small hand-curated overlay.
+
+Because it is community-maintained and can lag the CLI, the *source* is a
+**configuration surface** (`settings_schema_url`, also env-overridable), not a
+constant — an org can point at an internal mirror or a version-pinned URL, or an
+official Anthropic schema if one is ever published. project-pulse tracks the drift.
+
 ### Independence and the shape contract
 
 *Managing* the configuration is independent of the ways matching engine — someone can
--- a/tools/ways-cli/refresh-settings-schema.sh
+++ b/tools/ways-cli/refresh-settings-schema.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# Refresh the vendored Claude Code settings JSON Schema (ADR-147).
+#
+# Maintainer action: re-vendors the schema, after which you rebuild so the
+# `include_str!` in schema_doc.rs embeds the new copy. Normal operation never
+# needs this — the schema ships bundled in the binary.
+#
+# Source resolution (mirrors cmd/settings/source.rs):
+#   1. $WAYS_SETTINGS_SCHEMA_URL          — one-shot override
+#   2. `ways settings schema --source`    — the configured/default URL, if a
+#                                            built `ways` is on PATH
+#   3. the SchemaStore default            — hardcoded fallback below
+set -euo pipefail
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+OUT="$HERE/src/cmd/settings/claude-code-settings.schema.json"
+DEFAULT="https://json.schemastore.org/claude-code-settings.json"
+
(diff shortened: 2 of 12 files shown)


aaronsb:
## Review — ADR-147 slice 2 (vendored schema + source + scaffold)

Solid slice: the `include_str!` + `LazyLock` bundling is sound (parse panics only on a build-time-vendored malformed file, never on runtime input), source precedence env > config > default is correct and consistently trimmed, and the `scope_finding` match handles managed scope correctly now that `ScopeClass` dropped `Normal` (managed-only/overridable *at* managed scope both fall through to `_ => None` = clean). Two real bugs and a few minor notes below.

### 1. (Medium) `derive_type` mis-types union properties → false lint ERROR on `strictPluginOnlyCustomization: true`

`strictPluginOnlyCustomization` is the one property in the vendored schema defined via a combiner:

```json
"anyOf": [ { "type": "boolean" }, { "type": "array", ... } ]
```

Its own description says *"true locks all four; false is an explicit no-op"* — so the boolean form is first-class and common. But `derive_type` (schema_doc.rs:138-153) prefers a structured branch:

```rust
if matches!(t, PropType::Object | PropType::Array) { return t; } // Array branch wins
```

so it reduces the key to `PropType::Array`. In lint.rs:69 the type check then runs `Array.matches(true)` → false → **`Severity::Error` "expects array, got a boolean"** on a legitimately-correct managed config. Errors gate the exit code, so `ways settings lint` fails a valid store.

Root cause: `derive_type`'s single-type reduction is fine for *placeholder* generation (you want one concrete skeleton) but wrong for *validation* (a union must accept any branch). Same latent trap for any future `type: ["boolean","null"]` nullable key — `matches(null)` would error. Suggest splitting the two needs: keep the structured-preferred type for the scaffold placeholder, but have the validator treat combiner/multi-type properties as permissive (`PropType::Other` / match-any), or make `matches` union-aware. No test authors this key with a value, so it's uncaught.

### 2. (Medium-low) `autoUpdates` regressed from known → "unrecognized" warning

The old curated table in `schema.rs` knew `autoUpdates` (a current, valid Claude Code boolean setting). SchemaStore lags it — the vendored schema has `autoUpdatesChannel` but not `autoUpdates` — and the overlay supplement was not updated to carry it. Net effect: a fragment with `autoUpdates: false` now emits `unrecognized settings key` (warning only, no exit-code impact, but a false nag on a common key). This is exactly the "keys SchemaStore still lags" case the overlay is meant to cover (alongside `fallbackModel`, `disableSideloadFlags`); `autoUpdates` was simply missed. Suggest adding it to the overlay's known-key supplement.

### 3. (Low) `next_prefix` arithmetic overflow panic on a crafted filename

scaffold.rs:90 does `max + 10`. A file with a ~10-digit prefix near `u32::MAX` (e.g. `4294967290-x.md`) parses into `max`, and `max + 10` panics in debug builds (wraps in release). Reachable from a directory listing. `max.saturating_add(10)` closes it.

### Notes (non-blocking)
- **`existing_fragment_for` is correct** for the flagged `model` vs `fallbackModel` case — the `-{key}.md` suffix requires a literal hyphen before the key, and `ends_with` is case-sensitive, so `-model.md` does not match `...fallbackModel.md`. It holds for all camelCase/no-hyphen settings keys, but the guarantee rests on that key shape; a defensive exact filename parse would be more robust than suffix matching.
- lint.rs top doc (check 1) still says keys must exist in "the curated schema"; the key set now comes from the **vendored schema OR the overlay**. Worth a wording refresh.
- `derive_type` treats `allOf` like `oneOf`/`anyOf` (first structured branch). `allOf` is intersection, not union — semantically wrong, but currently unexercised (no top-level `allOf` in the vendored schema). Fine as a documented simplification.
- `SettingsSchema::len` is `pub` with no `is_empty` — worth confirming clippy's `len_without_is_empty` doesn't fire (or add `#[allow]`/`is_empty`).

Findings 1 and 2 are the ones I'd act on before merge; neither is caught by the current tests, so each wants a regression test (`strictPluginOnlyCustomization: true` lints clean; `autoUpdates: <bool>` is known).

aaronsb:
Review addressed in 79288dc: (1) union types (anyOf/oneOf/multi-type) now validate permissively — analyze() decouples the validation type (Other for unions) from the scaffold placeholder (concrete branch), fixing the false ERROR on `strictPluginOnlyCustomization: true` + regression test; (2) autoUpdates added to a KNOWN_LAGGED overlay supplement so it's not false-flagged 'unrecognized' + regression test; (3) next_prefix saturating_add; (4) doc wording. 37 settings unit + 7 integration green, clippy clean.

PR #241, opened 2026-07-01
refactor(settings): externalize schema to runtime file + resolve ADR-147 open questions

aaronsb:
## Summary
Externalize the Claude Code settings schema so it updates without recompiling ways (CC's settings surface changes often), and resolve ADR-147's three open questions.

**Externalization:**
- Drop `include_str!`. The schema is now a data file at `share/claude-code-settings.schema.json`, riding the app's data-dir git clone; read **lazily at runtime** by `schema_doc::active()` only when a `ways settings` command needs it — never on the every-turn hook path.
- `paths::settings_schema_file()` resolves by precedence: `$WAYS_SETTINGS_SCHEMA_FILE` › `$XDG_CONFIG` durable copy (the `--refresh` target) › shipped `share/` copy (refreshed by `make update`).
- **Graceful degradation:** if the file is absent, `lint` skips schema-valid with one notice (scope-legal + duplicate still run); `scaffold` errors with guidance. `check()` now takes an injected `Option<&schema>` — degrades explicitly, stays testable.
- `refresh-settings-schema.sh` writes the durable user copy (`--shipped` updates the repo copy); no rebuild. `ways settings schema` shows the resolved file path + key count.

**ADR-147 open questions resolved:** store location (path is a parameter — a shared org repo is just a different argument); test depth (static-only; decline the live-boot gate — CC parses tolerantly, so a boot is low-signal); manager/ways boundary (`ways` subcommand; independence is a clean code boundary, standalone binary deferred).

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml` — 225 unit + 9 settings integration (incl. schema-absent degradation for both lint and scaffold), clippy clean.
- Live: `ways settings schema` reads the external file (84 keys); scaffold → lint round-trips clean.
- Pre-existing `session_sim` failures are environment-only (embedding corpus), unchanged.


aaronsb:
## Code Review — externalize settings schema (ADR-147)

Solid refactor. `include_str!` → runtime data file is the right call given CC's churn, the `Option<&schema>` injection makes `check()` honestly testable, and dropping the old `.expect(\"vendored… must parse\")` in favor of degrade-to-`None` removes a reachable abort. Control flow in the degraded path is clean (verified below). Findings ranked by severity.

### 1. (Medium) Recovery guidance points at a `--refresh` flag that doesn't exist
`lint.rs:65` and `scaffold.rs:23` tell the user to run `ways settings schema --refresh`. The `Schema` subcommand only defines `--source` (`main.rs:593-596`), so clap rejects `--refresh` with \"unexpected argument.\" This message is shown *exactly* when the schema is missing — the moment correct guidance matters most — and it sends the user to a dead end. `mod.rs:36,43` gets it right, naming `refresh-settings-schema.sh`. Fix: unify all three on the script, or actually add a `--refresh` flag to the `Schema` command.

### 2. (Low-Med, latent) `active()` caches `None` for the whole process
`LazyLock<Option<…>>` caching `None` is safe *today* because `ways` is a one-shot CLI and no in-process refresh exists. But it's coupled to finding #1: the instant someone implements `ways settings schema --refresh` to write-then-report in one process, the pre-refresh probe will have cached `None` and the post-refresh confirmation will lie. When that lands, have refresh call `load()`/`settings_schema_file()` directly, or document `active()` as a per-process snapshot. The `exists()`-then-`load()` TOCTOU is benign — a file vanishing between them just yields the \"unusable\" eprintln instead of a silent `None`, no panic.

### 3. (Low) `settings_schema_file()` env-override edge handling
`paths.rs:708-712`: emptiness is tested on `p.trim()` but the `PathBuf` is built from the raw `p`, so `WAYS_SETTINGS_SCHEMA_FILE=\" /x \"` yields a path with surrounding spaces; the env branch also skips `normalize_path_sep` that both other branches get (Windows parity). Minor. Separately, the doc comment says \"first existing wins,\" but the env branch returns unconditionally — that's *correct and deliberate* (lets errors name the chosen path, and lets the degradation tests force absence), so tighten the comment to call it an unconditional override rather than existence-gated.

### 4. (Nit) Absent-schema `Finding.file` carries the schema path, not a fragment path
`lint.rs:62`. Serializes cleanly (empty `key`, valid JSON) and the message explains it, so acceptable — but JSON consumers keying `file` as a fragment path get a surprise. Worth a one-line doc note.

### 5. (Nit) Stale \"vendored\" wording
`main.rs:592` still says \"Show the vendored … schema\"; a few module-doc/inline comments in `lint.rs`/`schema_doc.rs` retain \"vendored.\" The PR de-vendors — sweep for accuracy.

### Control-flow verification (no action needed)
- **Degradation:** schema `None` → exactly one notice, per-key schema block guarded by `if let Some(sc)` so no unrecognized-flood; scope-legal and duplicate run outside that guard, so they can't be skipped; can't double-emit (notice guarded by `is_none`). Empty-key notice serializes fine.
- **Shipping assumption holds:** `data_root` is a full git checkout (`git rev-parse --show-toplevel` == data_root), `share/…schema.json` is committed and not gitignored, so `make update`'s `git pull` lands it at `data_root/share/`. `download-ways.sh` only swaps the binary *inside* that existing clone, so it never strips `share/`. The only absent-`share/` case is a release binary run entirely outside a clone (outside ADR-140 topology) — and there degradation is correct (lint notice + exit 0; scaffold errors). Assumption is sound; consider a reconcile-time smoke check so a partial pull (new binary, missing `share/`) is caught rather than silently degraded.
- **Test integrity:** `test_schema()` path (`CARGO_MANIFEST_DIR/../../share`) is correct and `.expect()`s loud on absence; integration tests force the schema via `WAYS_SETTINGS_SCHEMA_FILE` (repo copy) or `/nonexistent`, so nothing rides ambient `data_root/share` — no spurious green. The unconditional env override (#3) is what makes the degradation tests robust.

_AI-assisted review via Claude Code_

aaronsb:
Review addressed in 21fc3c1: (1) BLOCKER — implemented `ways settings schema --refresh` (fetch configured source → durable user copy, no rebuild) so the degradation guidance is now true and unified across lint/scaffold/mod, instead of pointing at a nonexistent flag; (2) refresh validates via load() not active() so no pre-refresh None is cached; (3) trimmed the env override + fixed the precedence comment; (4) doc-noted the synthetic notice's file field; (5) killed stale 'vendored' wording. New hermetic file:// --refresh integration test. 225 unit + 10 integration green, clippy clean.

PR #242, opened 2026-07-01
feat(settings): ways settings compile — merge engine + baked output (ADR-147 slice 3)

aaronsb:
## Summary
Third ADR-147 slice: compile a fragment store into baked `settings.json`.

- **Merge engine** (`compile.rs`): `merge_scope()` deep-merges a scope's fragments in filename order per CC's documented law — objects deep-merge, the documented concat paths (`permissions.allow/deny/ask`, `deniedMcpServers`) union+dedupe, everything else overrides (last wins). Yields a **provenance** map (dotted path → contributing fragments). Minimal-correct 3-strategy law (decided); extend as CC edge cases surface.
- **Orchestration**: lint-gates (refuses to bake a store with errors), groups by scope, merges each. `--out <dir>` writes `<scope>.settings.json` + `provenance.json`; else prints a single scope's baked blob to stdout. Managed scope = single pre-merged blob (decided).
- **CLI**: `ways settings compile <store> [--scope] [--out]`.

The baked user blob is what the reconciler will project (next slice); the managed blob is what you paste into the enterprise console.

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml` — 231 unit (incl. 6 merge-law conformance) + 13 settings integration (compile: --out+provenance, stdout, lint-gate refusal), clippy clean.
- Pre-existing `session_sim` failures are environment-only (embedding corpus), unchanged.

### DIFF
--- a/tools/ways-cli/src/cmd/settings/compile.rs
+++ b/tools/ways-cli/src/cmd/settings/compile.rs
@@ -1,0 +1,193 @@
+//! Compile a fragment store into baked `settings.json` (ADR-147).
+//!
+//! Each scope's fragments are deep-merged in filename order into one baked
+//! object, following Claude Code's documented merge law. The law is small — a
+//! path-keyed strategy, not a merge engine:
+//!
+//! - **objects deep-merge** (recurse key by key);
+//! - the documented **concat** paths (`permissions.allow/deny/ask`,
+//!   `deniedMcpServers`) union their arrays with de-duplication;
+//! - **everything else overrides** — the last fragment in filename order wins,
+//!   which is exactly the case the duplicate-scalar lint warns about.
+//!
+//! Merging also yields a **provenance** map (`dotted path → contributing
+//! fragment(s)`): the effective (last) source for an overridden key, every
+//! contributor for a concatenated list. That is both `git blame`-for-config and
+//! the future reconciler's merge base.
+
+use super::fragment::Fragment;
+use serde_json::{Map, Value};
+use std::collections::BTreeMap;
+
+/// `dotted path → fragment file name(s)` that produced the effective value.
+pub type Provenance = BTreeMap<String, Vec<String>>;
+
+/// The documented concat-and-dedupe paths. Everything else overrides (objects
+/// deep-merge). Kept explicit and minimal; extend as Claude Code's edge cases
+/// surface (see ADR-147 Q3 — the merge law is a spec to mirror, not invent).
+fn is_concat(path: &str) -> bool {
+    matches!(
+        path,
+        "permissions.allow" | "permissions.deny" | "permissions.ask" | "deniedMcpServers"
+    )
+}
+
+/// Deep-merge a scope's fragments (already filename-ordered) into one baked
+/// settings object, plus the provenance of every leaf.
+pub fn merge_scope(frags: &[&Fragment]) -> (Value, Provenance) {
+    let mut acc = Value::Object(Map::new());
+    let mut prov = Provenance::new();
+    for frag in frags {
+        let name = frag
+            .path
+            .file_name()
+            .and_then(|s| s.to_str())
+            .unwrap_or("<fragment>")
+            .to_string();
+        if let Some(obj) = frag.settings.as_object() {
+            merge_object(&mut acc, obj, "", &name, &mut prov);
+        }
+    }
+    (acc, prov)
+}
+
+/// Merge `incoming` into `acc` (both objects) at `prefix`, recording provenance.
+fn merge_object(
+    acc: &mut Value,
+    incoming: &Map<String, Value>,
+    prefix: &str,
+    frag: &str,
+    prov: &mut Provenance,
+) {
+    let acc_obj = acc.as_object_mut().expect("merge target must be an object");
+    for (key, val) in incoming {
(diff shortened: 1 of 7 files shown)


aaronsb:
## Review: `ways settings compile` (merge-law focus)

Solid MVP — the strategy split is legible, the four `.expect()`s are genuinely unreachable, and CLI ergonomics (multi-scope bail, empty-store bail, `--out` vs stdout) are clean. All 10 tests pass. The concerns are in the **merge law itself** and its agreement with the linter, plus provenance under shape-changes.

### HIGH — the concat set diverges from documented CC, and the gap is silent

**H1 · `is_concat` is missing documented-merge arrays.** Per CC docs (settings/permissions.md), these arrays *merge across layers*: `permissions.allow`, `permissions.deny`, `permissions.ask`, **`permissions.additionalDirectories`**, and **`hooks`** (merge + dedupe). `is_concat` (compile.rs:119) covers the first three + `deniedMcpServers` but **omits `permissions.additionalDirectories`** — which is documented — while *including* `deniedMcpServers`, whose concat semantics CC does **not** clearly document. Concrete drop: `10-a.md` `permissions.additionalDirectories: ["/a"]` + `20-b.md` `["/b"]` bakes to `["/b"]`, losing `/a`; real CC unions them. `hooks` is a harder case (nested arrays under event keys) and can stay a documented MVP gap — but `additionalDirectories` is a one-line fix and should land now.

**H2 · lint's array-exemption hides exactly these drops.** The duplicate-scalar check skips all non-scalars (lint.rs:135) on the premise *"arrays concatenate under Claude Code"* (test `duplicate_ignores_objects_and_arrays`, lint.rs:373). But compile concatenates only 4 arrays and **overrides every other array** (test `plain_arrays_override_not_concat`). So a non-concat array set twice in one scope (`availableModels`, `allowedMcpServers`, `permissions.additionalDirectories`) is silently dropped by compile **and** silently passed by lint. The two halves of the feature disagree on which arrays concatenate. Either lint must warn on duplicate non-concat arrays, or both must share one predicate (see L3).

**H3 · `allowedMcpServers` vs `deniedMcpServers` asymmetry.** schema.rs:103-107 (and its comment) name **both** MCP lists as concatenating; lint.rs:366 protects `deniedMcpServers` as concat. compile's `is_concat` lists only `deniedMcpServers` — so `allowedMcpServers` gets overridden. The codebase models the pair symmetrically everywhere except the one place that merges. Pick one.

### MEDIUM — provenance goes stale on shape-changes

**M1 · Orphaned / mis-credited provenance when a path changes shape.** `merge_object` records prov per leaf but never prunes:
- *object→scalar*: `permissions:{allow:[...]}` then `permissions:"off"` → the scalar overrides (prov `permissions`=[20-b]), but `prov["permissions.allow"]`=[10-a] **remains**, describing a leaf the baked object no longer has.
- *scalar→array on a concat path*: `permissions.allow:"x"` then `permissions.allow:["y"]` → the concat branch replaces the non-array slot (compile.rs:172) yet still `push`es onto the old prov vec → `[10-a, 20-b]`, crediting 10-a whose scalar was discarded.

Reachable whenever the schema is absent (`active()==None` → no type gate) and for nested paths even *with* a schema (lint type-checks only top-level keys). Since provenance is a stated deliverable (git-blame / reconciler merge-base), silent wrongness matters. Fix: on a shape-change override clear prov under that prefix; on concat-replace reset the vec instead of pushing.

### LOW

- **L1** Concat provenance credits a fragment that added only duplicates (net-zero contribution still recorded as a contributor).
- **L2** compile re-implements the "any `Severity::Error`" predicate (compile.rs:43) that lint already exposes (lint.rs:223) — share a helper so gate and reporter can't drift.
- **L3** `is_concat` and lint's array-exemption encode the *same* merge law in two files with no shared source — the structural root of H2/H3. One `concat_paths()` used by both would make disagreement impossible.

### Verified sound
Filename-order == precedence holds (load_dir byte-wise `file_name` sort, zero-pad caveat documented). The tricky no-panic cases — concat path whose value isn't an array, and object↔scalar shifts — are all guarded (`&& val.is_array()`, `if !entry.is_object()`). Lint gate threads `schema_doc::active()` and degrades (not blocks) on a missing schema; the dirty-fixture refusal fires on a scope error that needs no schema, so the gate is robust to schema absence.

**Assessment:** merge engine and provenance are correct for the pure override / pure concat paths the tests cover; the gaps are (a) an incomplete + slightly-wrong concat set vs documented CC, (b) a lint/compile contract mismatch that makes array drops silent, and (c) provenance staleness on shape-changes. H1+H2 are the ones I'd fix before merge — they're the conformance-critical core.

aaronsb:
Review addressed in c2fcafb: (H1) permissions.additionalDirectories now concatenates (+ test); (H2/H3/L3) single shared compile::is_concat_path() is the concat law — lint's duplicate check consults it and now warns on a repeated non-concat array (lossy under compile), closing the divergence; MCP allow/deny made symmetric; the bug-encoding test split into concat-vs-override; (M1) provenance pruned on shape-change (+ test); (L1) concat credits only contributing fragments; (L2) shared has_errors predicate. 47 settings unit + 13 integration green, clippy clean.

PR #243, opened 2026-07-01
feat(settings): ways settings project — install compiled config into live settings.json (ADR-147)

aaronsb:
## Summary
The pipeline's final stage: project a compiled store into the live `settings.json`.

- `ways settings project <store> [--scope] [--dry-run]`: compiles (lint-gated, via the extracted `compile_store`), then three-way-merges each scope's baked object into the live settings — user → `~/.claude/settings.json`, project → `$CLAUDE_PROJECT_DIR/.claude/settings.json`, managed → printed blob (**never** auto-writes a system path).
- **Coexists with the reconciler** (`settings_merge` co-owns `{hooks, ways-perms}`): project owns the *fragment store's* keys, keeps its own last-applied base (`$XDG_STATE/agent-ways/settings-fragments-<scope>.json`) to clean up keys it stops setting, and preserves everything it doesn't own — the `kubectl apply` field-ownership model with two disjoint controllers.
- **MVP scope:** top-level override + cleanup; `hooks`/`permissions` are skipped with a warning (the reconciler co-owns them; recursive permissions-coexistence deferred). Writes keep a `.bak`; `--dry-run` previews.

This completes author → lint → compile → **project**. The config-authoring skill (next) will drive these primitives.

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml` — 238 unit (4 project: override+preserve, cleanup of dropped keys, skip-reconciler-owned, dry-run) + project integration (project-scope apply + dry-run), clippy clean.
- Pre-existing `session_sim` failures are environment-only, unchanged.

### DIFF
--- a/tools/ways-cli/src/cmd/settings/compile.rs
+++ b/tools/ways-cli/src/cmd/settings/compile.rs
@@ -25,34 +25,30 @@ use std::path::Path;
 /// `dotted path → fragment file name(s)` that produced the effective value.
 pub type Provenance = BTreeMap<String, Vec<String>>;
 
-/// Compile a store: lint-gate, then merge each scope into a baked settings
-/// object + provenance. Returns `true` if the lint gate failed (caller exits
-/// non-zero). With `out`, writes `<scope>.settings.json` + `provenance.json`;
-/// otherwise prints a single scope's baked blob to stdout.
-pub fn run(store: &Path, scope_filter: Option<Scope>, out: Option<&Path>) -> Result<bool> {
+/// The result of compiling a store: either the lint gate refused, or a baked
+/// object + provenance per scope present.
+pub enum Outcome {
+    /// Lint errors blocked the bake (the error findings, for reporting).
+    Refused(Vec<lint::Finding>),
+    /// One `(scope, baked settings, provenance)` per scope that had fragments.
+    Baked(Vec<(Scope, Value, Provenance)>),
+}
+
+/// Lint-gate a store, then merge each scope into a baked object + provenance.
+/// Pure (no output) — shared by `compile` and `project`.
+pub fn compile_store(store: &Path, scope_filter: Option<Scope>) -> Result<Outcome> {
     let frags = load_dir(store)?;
 
-    // Lint gate — refuse to bake a store with errors (warnings are fine).
     let findings = lint::check(&frags, schema_doc::active());
     if lint::has_errors(&findings) {
-        let errors: Vec<_> = findings
-            .iter()
+        let errors = findings
+            .into_iter()
             .filter(|f| f.severity == lint::Severity::Error)
             .collect();
-        eprintln!(
-            "compile refused: {} lint error(s) — fix first (`ways settings lint {}`):",
-            errors.len(),
-            store.display()
-        );
-        for e in &errors {
-            eprintln!("  {}: {}", e.file, e.message);
-        }
-        return Ok(true);
+        return Ok(Outcome::Refused(errors));
     }
 
-    // Merge each scope present (or the requested one), in filename order.
-    let mut baked: BTreeMap<&str, Value> = BTreeMap::new();
-    let mut provenance: BTreeMap<&str, Provenance> = BTreeMap::new();
+    let mut scopes = Vec::new();
     for scope in [Scope::User, Scope::Project, Scope::Managed] {
         if scope_filter.is_some_and(|f| f != scope) {
             continue;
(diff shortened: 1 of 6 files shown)


aaronsb:
## Review: `ways settings project` (ADR-147 final stage)

Reviewed as a second live-`settings.json` writer coexisting with the reconciler (`settings_merge`). The disjoint-field / `kubectl apply` model is sound and the top-level three-way merge is **logically correct** for the override + cleanup cases. My concerns are about **crash-safety, an unreachable cleanup path, and the missing self-audit** — all of which matter because this writes the user's live config. Ranked by severity.

### HIGH — Non-atomic write of live `settings.json`
`project_write` (project.rs:172+) writes the target with a plain `std::fs::write(target, …)`. The reconciler deliberately uses temp-file + `rename` (`write_json_atomic`, settings_merge.rs:367) precisely so "a crash never leaves a half-written file." This new writer regresses that guarantee on the *same* `~/.claude/settings.json`: an interrupt/power-loss between truncate and full write leaves a truncated or corrupt live config. A `.bak` is written first so it's recoverable by hand, but the two writers should share the same atomic discipline.
**Fix:** reuse a temp+rename write (or lift `write_json_atomic` into a shared helper).

### HIGH — Full-removal cleanup is unreachable through the CLI (test masks it)
The module doc promises project can "remove a key it stops setting." But `compile_store` (compile.rs:57-59) **skips any scope with zero fragments** (`if scoped.is_empty() { continue; }`), and `run` only iterates the scopes it returns. So if the operator deletes *all* fragments for a scope, that scope is never projected → the keys recorded in `settings-fragments-<scope>.json` are **never GC'd** from `settings.json`. The base file still lists them but is never revisited; the operator cannot un-set, via the tool, what the tool installed.
The `removes_a_key_we_stopped_setting` test passes only because it calls `project_write` directly with `json!({})`, **bypassing `compile_store`** — so it exercises a code path the real pipeline can't reach. This is the classic "green test, dead path."
**Fix:** drive cleanup from the union of (scopes-with-fragments ∪ scopes-with-a-base-file), or have `run` reconcile every scope that has a persisted base even when no fragment remains. And add a test that goes through `compile_store`/the binary.

### MEDIUM — No self-audit; `project` writes `~/.claude` with no invariant check
The reconciler's central safety property is the post-write self-audit (`stripped_user_view` before == after, revert-on-mismatch, settings_merge.rs:331-344): a botched merge becomes a loud failure, never silent corruption. `project` has **no equivalent**. The merge here is simpler (top-level whole-value), so the blast radius is smaller, but this is the writer with *no* net. Given HIGH #2 and the deferred permissions recursion, an "unmanaged top-level key count/identity unchanged except our owned set" assertion would be cheap insurance on the user's live file.

### MEDIUM — `read_json` collapses every read error to "empty object"
project.rs:57-62: any `Err` from `read_to_string` (permission denied, transient I/O, is-a-dir) — not just not-found — yields `{}`. Since `{}` then means "no live keys," the merge writes a file containing **only project's keys**, dropping the user's real content. It's backed up to `.bak` first so it's recoverable, but an existing-but-momentarily-unreadable file should not be treated as absent. (A *parse* error correctly propagates and aborts — good; only the I/O-error path is over-eager.) The reconciler shares this shape but its self-audit partly covers it; here nothing does. Distinguish `NotFound` from other errors.

### MEDIUM — Coexistence is asserted but never tested end-to-end
The top-level disjointness is correct: reconciler owns `{hooks, permissions}` and passes everything else through; `project` skips exactly those two and preserves them via the `live` clone (the `continue` never removes them from `result`). Interleaved `reconcile → project → reconcile` is safe by construction. **But no test exercises it.** `skips_reconciler_owned_keys` is a single `project` run only. A test that runs the real reconciler, then `project_write`, then asserts hooks/perms survived (and vice-versa) would lock the contract that the whole design rests on.

### LOW — Shared `.bak` slot between the two writers
Both `project` (project.rs:162) and the reconciler (settings_merge.rs:317) back up to the identical `~/.claude/settings.json.bak`. Sequential CLI invocations, so no mid-flight clobber, and each writer only needs its backup during its own run — but the operator only ever has one `.bak`, reflecting whichever writer ran last. Fine for now; worth a distinct suffix (e.g. `settings.json.project.bak`) if the backup is ever meant as a durable rollback point. Also `fs::copy(...).ok()` silently proceeds if the backup fails, then writes unprotected.

### LOW / acknowledged — `permissions` skipped wholesale is coarse but *safe*
Skipping the entire `permissions` object (not just the reconciler's 6 owned strings) means fragments can't manage `permissions.deny` or non-ways allow entries — a real limitation the PR already flags as MVP. Importantly it is **not a correctness trap**: by never writing `permissions`, `project` structurally cannot clobber the reconciler's nested strings, so the coarse top-level `reconciler_owned` check is sufficient. No need to also protect the nested strings *while* the whole key is skipped. Just don't narrow the skip later without adding the nested three-way merge at the same time.

### Nits
- `applied`/base is written after the target with best-effort ordering (not atomic, not crash-consistent). If the base write fails after the target write, base goes stale; mostly self-correcting on next run, but worth `write_json_atomic` for the base too.
- Owned-value GC edge: if `project` set `K=30`, the operator independently keeps `30`, then the fragment is dropped, `live==base` → `K` is removed. Correct owned-field semantics, but silent; the "keep (changed since we set it)" branch only protects *divergent* values.

### What's good
- The override / keep-on-user-change / remove-on-unchanged cleanup logic is correct, and `applied` records exactly the non-reconciler-owned keys written, so the next base stays consistent with the file.
- First-projection (no base) and the disjoint-writer preservation of unmanaged keys are both right.
- Extracting `compile_store` + `report_refusal` as a pure seam is clean and the right shape for sharing with `compile`.
- No reachable `unwrap`/`expect`/panic in the non-test code.

**Recommendation:** address HIGH #1 (atomic write) and HIGH #2 (unreachable cleanup + the masking test) before merge — both are on the user's live config path. The MEDIUM items (self-audit, read-error handling, coexistence test) are strongly advised given this is the writer without a net.

---

aaronsb:
Review addressed in 309eb11: (H1) atomic write via the reconciler's write_json_atomic (now pub(crate)) — no more truncation risk; (H2) cleanup path made reachable — run() processes scopes-with-fragments ∪ scopes-with-a-base-file so an emptied scope GCs its keys, + end-to-end GC integration test that drives the previously-dead path through the binary; (M3) self-audit refuses to write if it would touch an unowned key; (M4) read_json distinguishes NotFound from real I/O errors; (M5) coexistence unit test; (L6) distinct .ways-project.bak backup slot. 7 project tests green, clippy clean.

PR #245, opened 2026-07-01
feat(skills): ways-settings — interview-driven Claude Code config authoring (ADR-149)

aaronsb:
## Summary
The capstone of the ADR-147/149 arc: a skill that is the conversational **conductor** for the `ways settings` CLI. It interviews the operator to author `settings.json` **fragments**, then lints, compiles, and projects them into `~/.claude`.

- **Synthesize, don't rebuild** (ADR-149): composes four sources — Claude Code's own config knowledge + the settings schema + the operator's history + the **`/insights` report** (read agent-natively, not parsed). The interview's "why" becomes the fragment's rationale body.
- **Sub-functions** map onto the primitives: author (`new` + fill), check (`lint`), rebuild (`compile`), project (`project`), pull-schema (`schema --refresh`), suggest (`/insights` + telemetry, the high-value move).
- **Skills Way compliance**: tight description with trigger words + a "not for" clause; location-independent (resolves `$STORE` up front, no cwd assumption); leans on real `ways` tooling; **dry-run-then-confirm before writing the live config**.
- `/insights` boundary recorded: it mines the *settings-relevant* sections (friction → permissions, features → toggles) and leaves "Suggested CLAUDE.md Additions" to memory (a different surface).

## Test plan
- Frontmatter parses; `name` matches the directory.
- Every documented flag (`--scope`, `--dir`, `--out`, `--dry-run`, `--refresh`, `--json`) verified against the built binary.
- Manual validation after merge+restart: confirm the skill fires on "help me configure Claude Code" and not on near-miss requests (per the Skills Way).

### DIFF
--- a/skills/ways-settings/SKILL.md
+++ b/skills/ways-settings/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: ways-settings
-description: Interview-driven authoring of Claude Code's settings.json as composable, lintable fragments — scaffold from the settings schema, lint, compile, and project into ~/.claude. Synthesizes Claude Code's own config knowledge and its /insights usage report with the operator's history. Use when the operator wants to set up, review, or change their Claude Code settings or configuration. Not for editing individual ways (that is the ways skill), CLAUDE.md memory, or updating the agent-ways install (ways-update).
-allowed-tools: Bash, Read, Edit, Write
+description: Manage Claude Code's settings.json as a composable, versioned store of interviewed config fragments — scaffold from the schema, lint, compile, and project into ~/.claude, with rationale and provenance. Synthesizes Claude Code's own config knowledge and its /insights usage report with the operator's history. Use when the operator wants to build up, review, or reason about their configuration as a documented fragment store. Not for a one-off settings.json change (Claude Code's built-in update-config handles those), editing individual ways (the ways skill), CLAUDE.md memory, or updating the agent-ways install (ways-update).
+allowed-tools: Bash, Read, Edit
 ---
 
 # ways-settings: interview the operator to author Claude Code config
@@ -11,6 +11,10 @@ is deterministic mechanism — `lint`, `new`, `schema`, `compile`, `project`. Th
 skill is the intelligence on top: it *interviews* the operator, authors config
 **fragments** (markdown files with a `settings:` YAML block + a rationale body),
 then lints, compiles, and projects them into `~/.claude/settings.json`.
+
+For a single ad-hoc setting change, Claude Code's built-in `update-config` is
+lighter — reach for *this* skill when the operator wants their config **managed**:
+authored with rationale, linted, versioned as a store, and re-projectable.
 
 ## The one rule: synthesize with Claude Code, don't rebuild it
 
@@ -87,6 +91,8 @@ ways settings project "$STORE"                     # apply (keeps a .ways-projec
 - `project` coexists with the reconciler: it owns the fragment keys and preserves
   `hooks`/`permissions` (the reconciler's) — it skips those with a warning.
 - Managed scope is never auto-written; it prints a blob to paste into the console.
+- Project scope writes `$CLAUDE_PROJECT_DIR/.claude/settings.json` (or cwd) — the one
+  cwd-dependent path. Confirm the operator is in the intended project before applying.
 
 ### pull-schema
 ```bash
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: skills/ways-settings/SKILL.md

Reviewed against the Skills Way (`hooks/ways/meta/skills/skills.md`) and ADR-149, with every `ways settings` invocation and behavioral claim checked against `tools/ways-cli/src/`. This is close to ready — the substance is strong. Findings below, ranked.

### Verified correct (worth stating)
- **Command accuracy — all match `main.rs`:** `new <key> --scope --dir`, `lint <store> [--json]`, `compile <store> --scope/--out`, `project <store> --dry-run`, `schema --refresh`. Positional-vs-flag usage is right (store is positional for lint/compile/project; `new` takes `--dir`).
- **Behavioral claims match the source:** the `.json.ways-project.bak` backup (`project.rs:215`), hooks/permissions **skipped-with-a-warning** as reconciler-owned (`project.rs:13-14,26`), managed never auto-written, and the `NN-<key>.md` fragment name (`scaffold.rs:56`) are all accurate.
- **`project` safety bar is met:** dry-run + explicit operator confirmation is stated repeatedly and bolded; no path projects without a preview.
- **ADR-149 fidelity is excellent:** \"read /insights, don't parse it,\" the \"Suggested CLAUDE.md Additions is memory, not settings — different surface\" boundary, and the \"skills can't invoke slash commands → ask the operator to run /insights\" constraint are all carried faithfully.
- **Location-independence:** `$STORE` resolved up front, `command -v ways` preflight, schema-availability note. Naming/family (`ways-settings`, matches dir) and defer-don't-duplicate are honored.

### Medium — trigger overlaps the built-in `update-config` skill
The `description` closes with *\"Use when the operator wants to set up, review, or change their Claude Code settings or configuration.\"* That is nearly the same surface as the always-present `update-config` skill (\"any changes to settings.json/settings.local.json files\" — permissions, env vars, hooks). Both live in personal/global scope, so at runtime they compete for requests like *\"allow npm in my settings\"* or *\"set DEBUG=true\"*. The \"not for\" clause excludes ways / CLAUDE.md memory / ways-update but does **not** distinguish from one-off settings.json edits. This is exactly the loose-global-trigger hijack the Skills Way warns about, and ADR-149's own Open Question flags the phrasing as unresolved. Suggest leaning the trigger on this skill's *distinctive* value — the fragment store + interview + /insights synthesis — rather than generic \"change their settings,\" and adding a boundary line vs. direct one-off settings.json edits.

### Low — \"or configuration\" is broader than settings.json
Even setting `update-config` aside, \"...settings **or configuration**\" reaches past settings.json toward MCP config, keybindings, direnv, etc. Tightening to settings.json specifically reduces near-miss fires.

### Low — `allowed-tools` grants `Write`, which the skill's own rule discourages
The doc is emphatic that fragments come from the CLI (\"let the CLI create it… never hand-invent a key\") and uses `Edit` to fill placeholders. No documented step exercises `Write`; granting it widens the surface toward the exact hand-authoring bypass the skill warns against. Consider `Bash, Read, Edit` only — or document the write path if store-bootstrapping is intended.

### Low — `suggest` sits slightly ahead of ADR-149's MVP line
ADR-149 marks `suggest` (read /insights + mine telemetry) as **v2**, with MVP being interview→author→lint→compile→project on schema + CC self-knowledge. The skill presents `suggest` as a live sub-function reading /insights now, scoping only telemetry mining to v2. Defensible (the ADR's own MVP wording on /insights is fuzzy, and reading the report is low-cost/high-value), but a one-line note aligning skill and ADR would remove the ambiguity.

### Nit — project-scope projection is cwd-dependent
The skill is otherwise rigorously location-independent, but `project --scope project` writes `$CLAUDE_PROJECT_DIR/.claude` — inherently tied to where the session is rooted. A one-liner (\"project scope targets the current project — confirm you're in the right one before applying\") would close the loop for a global skill.

None of these block the interview→…→project MVP. The Medium trigger item is the one worth resolving before this lands in every project's skill list.

---

aaronsb:
Review addressed in 8ad9c65: differentiated the trigger from Claude Code's built-in update-config — the description now leans on the distinctive value (a managed, versioned store of interviewed fragments with rationale + provenance) and explicitly excludes one-off settings.json changes, in both the description and body. Also: dropped 'or configuration' (scoped to settings.json), removed the unused Write tool (-> Bash, Read, Edit), aligned suggest as v2 per ADR-149, and added a project-scope cwd caution. Thanks for verifying every command against source.

PR #246, opened 2026-07-01
feat(cli): ways update — self-update from the binary (pre-built first, rename-revert safe)

aaronsb:
## Summary
A `ways update` subcommand that wraps the mainline update flow — so the lifecycle is `ways migrate` / `ways settings` / **`ways update`** rather than "find the app dir, cd, make update, ways reconcile".

Two properties the naive `make update` lacks (both requested):
- **Pre-built first.** `make update` force-*builds* (breaks without a toolchain). This mirrors the *install* flow: download-first, build-fallback for `ways` + `way-embed`. `attend`/`attend-chat` have no pre-built binaries yet, so they refresh **only when a toolchain is present**, else keep their current version (with a clear note).
- **Rename-then-revert.** Each component binary is *renamed aside* (not removed) so the download/build re-acquires it; on failure the old binary is moved back — a failed update **restores the prior state** instead of stranding a binary-less install. Self-replacement is safe (the running process keeps its inode; the new binary lands for next invocation).

Flow: `scripts/update.sh` (pull) → refresh `ways` (fatal) → `way-embed` (optional) → `attend`/`attend-chat` (toolchain-gated) → corpus regen → `ways reconcile`. `--dry-run` prints the plan. Reuses the reconciler's legacy-clone guard.

## Test plan
- `cargo test --manifest-path tools/ways-cli/Cargo.toml` — 241 unit (incl. checkout-guard + rename-revert-on-failure), clippy clean.
- `ways update --dry-run` verified live against the real app dir (guards pass, toolchain detected, plan correct).
- Known gap (tracked): `attend`/`attend-chat` need pre-built binaries published to be fully build-tool-free — until then they're toolchain-gated.

### DIFF
--- a/tools/ways-cli/src/cmd/mod.rs
+++ b/tools/ways-cli/src/cmd/mod.rs
@@ -34,3 +34,4 @@ pub mod tree;
 pub mod tune;
 pub mod tune_curves;
 pub mod tune_precision;
+pub mod update;
--- a/tools/ways-cli/src/cmd/reconcile.rs
+++ b/tools/ways-cli/src/cmd/reconcile.rs
@@ -209,7 +209,7 @@ fn make_symlink(src: &Path, dst: &Path, is_dir: bool) -> Result<()> {
 
 /// True if `dest` is a legacy in-place agent-ways clone (its own git repo that
 /// also ships the app source) rather than a thin projection.
-fn is_legacy_in_place(dest: &Path) -> bool {
+pub(crate) fn is_legacy_in_place(dest: &Path) -> bool {
     // A projection has symlinked/looked-up subtrees but no app source of its
     // own; a clone has .git AND the app's source dirs (tools/, docs/).
     dest.join(".git").exists() && dest.join("tools").is_dir() && dest.join("docs").is_dir()
--- a/tools/ways-cli/src/cmd/update.rs
+++ b/tools/ways-cli/src/cmd/update.rs
@@ -19,8 +19,11 @@
 //!   binary is *renamed* aside (not removed) to defeat the "already installed"
 //!   early-return; if re-acquiring it fails, the old binary is moved back. A
 //!   failed update restores the prior state rather than stranding a binary-less
-//!   setup. (Self-replacement is safe: the running process keeps its own inode;
-//!   the new binary is in place for the next invocation.)
+//!   setup. (Self-replacement is safe on Unix: the running process keeps its own
+//!   inode, so replacing `bin/ways` under it is fine and the new binary lands for
+//!   the next invocation. On Windows a running `.exe` can't be replaced; there the
+//!   command fails early and safe — it shells `bash`/`make`, absent on a bare
+//!   Windows install — rather than corrupting anything.)
 
 use crate::paths;
 use anyhow::{bail, Context, Result};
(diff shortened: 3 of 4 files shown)


aaronsb:
## Review: `ways update`

Traced the rename-revert, the Makefile targets it drives (`ways`, `way-embed setup-binary`, `attend`), `scripts/update.sh`, and the two download scripts. The `ways` path is sound; the safety net is real. Findings ranked by severity.

### HIGH — way-embed rename-aside targets the wrong path; download-refresh never fires
`refresh_component(app, "way-embed", ["-C","tools/way-embed","setup-binary"], app)` renames **`app/bin/way-embed`** aside and later asserts `app/bin/way-embed` exists. But `setup-binary` → `download-binary.sh` installs to **`$XDG_CACHE/agent-ways/user/way-embed`** (`OUTPUT_FILE`), and its "already installed" early-return guards *that cache path*, which the rename never touches.

Failing sequence (download path, the common case):
1. rename `app/bin/way-embed` → `.pre-update` (had=true)
2. `download-binary.sh` sees a working `$XDG_CACHE/.../way-embed` → **early-returns, does nothing**
3. `bin.exists()` (`app/bin/way-embed`) is now false → `built=false` → revert restores old binary, returns `Err`
4. caller prints `⚠ way-embed not refreshed`

Net: way-embed is **never refreshed via the pre-built path**, and every run warns spuriously even though way-embed is healthy. It only actually updates when the *build* fallback runs (`$(MAKE) $(BIN)` → `app/bin/way-embed`). This directly contradicts the PR's "download-first for ways + way-embed" claim. `ways` works only because `download-ways.sh` `OUTPUT_FILE == app/bin/ways` — same path as the rename target. way-embed's canonical download location diverges, so the rename-aside trick doesn't apply to it.

### MEDIUM — pull (step 1) mutates shared source *before* the fatal ways gate
Order is guards → `scripts/update.sh` (pull) → refresh `ways` (fatal). Hooks/ways/skills are live symlinks into the app checkout, so after the pull `~/.claude` already resolves to the **new** markdown while the binary is still old and `corpus`/`reconcile` haven't run. If the `ways` refresh then fails and aborts, the install is left in a mixed state: new source text, stale corpus embeddings, old binary, un-reconciled projection (newly-added/removed ways not wired). "Restores the prior state" holds for the *binary* but not for the *source/projection*. Consider running `reconcile` (and corpus) best-effort even on an aborting refresh, or refreshing the binary before/independently of the pull.

### MEDIUM — interrupted refresh orphans the only good copy
If the process is killed between the rename-aside and `make` completing, `app/bin/<x>` is gone and the only copy sits in `.pre-update`. On the next run `had = bin.exists()` is **false**, so the revert branch never restores it: the code runs `make` with had=false and, if that also fails, leaves the slot empty with a live binary stranded in `.pre-update`. There's no startup reconciliation of a leftover `.pre-update`. A one-line "if `.pre-update` exists and `bin` doesn't, adopt it" at entry closes this.

### LOW/MEDIUM — `make install` (PATH-symlink refresh) is skipped
`make update` normally runs `update-binaries` + `install`; this command runs neither `install` nor its `$XDG_BIN/{ways,attend}` / `$CLAUDE_BIN/way-embed` wiring. It works only because those symlinks point at stable `app/bin/*` paths that pre-exist. A missing/stale PATH symlink won't self-repair here. Worth confirming `reconcile` re-materializes the binary projection roots; if it doesn't, a broken PATH link survives the update.

### LOW — Windows self-replacement claim overstated
The module doc says self-replacement is safe generally. On Windows `remove_file(&backup)` on the running `.exe` fails → stale `ways.exe.pre-update` (error is `let _ =`-ignored, so cosmetic), and re-run `rename` over an existing backup fails (Unix replaces atomically; Windows doesn't). In practice step 1 shells `bash scripts/update.sh` + `make`, unlikely to exist on Windows, so it fails early and safe — but the doc claim should be scoped to Unix.

### NITS
- Success-path `let _ = remove_file(&backup)` silently leaves a stale `.pre-update` if removal fails.
- `refresh_component`'s error "`make ...` did not produce a working binary — reverted" is opaque when the real cause is "no `gh` and no `cargo`" (both `download-ways.sh` and the cargo branch bail). A hint would help.
- `is_legacy_in_place` on the projection can false-positive if a user git-tracks `~/.claude` (has `.git`) with symlinked `tools`/`docs` (both `is_dir()` follow the link) → refuses to update. Niche.

### What's solid
- `ways` rename-aside correctly defeats both early-returns (Makefile `ways:` and `download-ways.sh`) because the target path matches.
- Fatal/optional split is right; a reverting `ways` refresh aborts before corpus/reconcile with the old binary intact.
- Uses the freshly-refreshed `app/bin/ways` for corpus + reconcile, guarded by an existence check — no reachable panic in `run`/`refresh_component`.
- On Unix the inode reasoning is correct (source build: `bin/ways` is a symlink, running inode is the cargo target; download: unlinking a running file is safe).
- Good test coverage for the revert-on-failure invariant and the checkout guard.

The HIGH is the one to fix before merge — the pre-built path is the headline feature and it's inert for way-embed.

aaronsb:
Review addressed in 125efae: (HIGH) way-embed now refreshes via its own cache-aware target `make -C tools/way-embed rebuild-binary` instead of the app/bin rename dance that targeted the wrong path; (MEDIUM) a failed ways refresh reverts + continues to reconcile so the pulled source is never left un-projected; (MEDIUM) orphaned .pre-update from an interrupted run is recovered at entry + regression test; (LOW) self-replacement doc scoped to Unix. 3 update unit tests, 242 total green, clippy clean. Great catch on the way-embed cache-path divergence.

PR #247, opened 2026-07-01
fix(reconcile): stale base no longer false-reverts on hooks change

aaronsb:
## Summary

- Fixes the `ways reconcile` failure surfaced during `ways update`: **"settings merge changed unmanaged fields — reverted from settings.json.bak"** whenever an update ships changed hook definitions.
- Root cause: the reconciler's last-applied base (`$XDG_STATE/agent-ways/settings-applied.json`) is only rewritten after a *non-no-op* apply. A long run of converged reconciles leaves it stale (observed in the wild as `hooks: {}`). The first hooks-changing update then can't attribute the prior (old-command) ways-hook entries in the live file to us — they match neither the base nor the new `ours` — so the self-audit treats them as vanished user content and reverts. No data loss (clean revert), but reconcile is blocked on exactly the updates that ship new hooks.
- Fix: **structural hook ownership.** `entry_is_ours` recognizes an entry as ours when every command targets the agent-ways projection (`.claude/hooks/` or `.claude/bin/ways`), independent of the persisted base. Wired into both the merge's user-entry filter and the self-audit strip. Byte-matching against the base still runs first; this is the base-independent backstop.

### Trade-off (called out for review)
A user hook hand-placed under `.claude/hooks/` would be classified as ours. This is contrary to the ways model (users extend via `ways/` fragments, not hand-edited `settings.json` hooks) and is documented in the code. Assumes the ADR-142 projection root (`~/.claude`).

## Test plan
- New regression `update_that_changes_our_hooks_with_a_lost_base_converges`: lost base (`hooks: {}`) + old command ≠ new command → converges (old removed, new present, user hook preserved, no abort). Fails pre-fix (old entry lingers → 3 entries → revert).
- Existing `reapply_is_idempotent_when_base_hooks_were_lost`, `preserves_user_authored_hooks`, `removes_our_deprecated_hooks_but_keeps_user`, `user_view_invariant_holds_across_merge`, migration seed test all still green.
- `cargo test -p ways`: 243 unit tests pass. (Pre-existing `session_sim` integration failures on clean main are environmental — marker/corpus setup — and unrelated to this change.)

### DIFF
--- a/tools/ways-cli/src/cmd/settings_merge.rs
+++ b/tools/ways-cli/src/cmd/settings_merge.rs
@@ -284,13 +284,41 @@ fn quote_first_token(cmd: &str) -> String {
     }
 }
 
-/// True if a hook *command* invokes the agent-ways projection — a `.claude/hooks/`
-/// script or the `.claude/bin/ways` CLI. Matches the literal `${HOME}`-prefixed
-/// form we ship and the expanded/quoted forms a prior merge may have written
-/// (substring match, so quoting and `${HOME}` vs `/home/...` both pass). Assumes
-/// the ADR-142 projection root (`~/.claude`).
+/// True if a hook *command*'s executable is an agent-ways projection artifact —
+/// a `.claude/hooks/…` script or a `.claude/bin/…` CLI (`ways`, `way-embed`,
+/// `attend`). Only the **first (exe) token** is examined (quote-aware), so a
+/// user hook that merely passes a projection path as an *argument* — e.g.
+/// `tail -f ${HOME}/.claude/hooks/ways/x.log` — is NOT mistaken for ours. That
+/// distinction is load-bearing: misclassifying a user hook would silently drop
+/// it (the self-audit strips it symmetrically, so no revert catches it).
+///
+/// Path separators are normalized so a Windows `\`-form path matches. Matches
+/// the literal `${HOME}`-prefixed form we ship and the expanded/quoted forms a
+/// prior merge wrote.
+///
+/// Assumes the ADR-142 projection root (`~/.claude`) — the hooks we ship are
+/// written with a literal `${HOME}/.claude/...` prefix, so a relocated config
+/// dir (`CLAUDE_CONFIG_DIR`) is out of scope; if that ever changes, the
+/// shipped-hooks guard test (`shipped_hooks_are_all_recognized_as_ours`) fails
+/// loudly rather than the reconciler silently regressing.
 fn command_is_ours(cmd: &str) -> bool {
-    cmd.contains(".claude/hooks/") || cmd.contains(".claude/bin/ways")
+    let exe = exe_token(cmd).replace('\\', "/");
+    exe.contains(".claude/hooks/") || exe.contains(".claude/bin/")
+}
+
+/// The executable token of a hook command: the leading `"`-quoted span when the
+/// command is quoted (the form our installer writes, e.g.
+/// `"${HOME}/.claude/bin/ways" corpus`), else the run up to the first
+/// whitespace. Mirrors the tokenization `quote_first_token` assumes.
+fn exe_token(cmd: &str) -> &str {
+    if let Some(rest) = cmd.strip_prefix('"') {
+        match rest.find('"') {
+            Some(i) => &rest[..i],
+            None => rest,
+        }
+    } else {
+        cmd.split_whitespace().next().unwrap_or(cmd)
+    }
 }
 
 /// True if a hook *entry* is one agent-ways ships: it runs at least one command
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: structural hook ownership

Reviewed `tools/ways-cli/src/cmd/settings_merge.rs`. The fix is sound in its core intent and the regression test reproduces the field failure precisely. The mechanism (structural, base-independent ownership applied symmetrically in both the merge drop and the self-audit strip) is the right shape. Findings below, ranked by consequence. Nothing here is a correctness regression against `main`; the top two are about widening/narrowing the ownership predicate.

### 1. Substring-over-whole-command widens the false positive to *silent* user-hook deletion (settings_merge.rs:292)

`command_is_ours` matches `.claude/hooks/` / `.claude/bin/ways` anywhere in the command **string**, not against the executable token. So a genuine user hook that merely *references* the projection as an argument is classified as ours:

- `tail -f ${HOME}/.claude/hooks/ways/foo.log`
- `mytool --config ~/.claude/hooks/whatever`

(confirmed: `tail -f ${HOME}/.claude/hooks/ways/foo.log` -> `entry_is_ours == true`.)

Because `entry_is_ours` is applied **symmetrically** — merge drop (`:120`) and `stripped_user_view` strip (`:197`) — the self-audit's `user_before` and `user_after` both omit the entry, so `user_before == user_after` (`:378`) and the write is **not** reverted. The user's hook is deleted with no error and no `.bak` restore. That is strictly worse than the documented trade-off, which only covers a hook *placed under* `.claude/hooks/` (an executable path), not an *argument* reference.

Suggestion: match the **executable token** only, not the whole string. The notion already exists in `quote_first_token` — take the first whitespace token (dequoted) and test whether *that* targets the projection. This aligns the code with the documented intent and removes the argument-reference class entirely.

### 2. Signature assumes POSIX separators and the literal `.claude` root — Windows and `CLAUDE_CONFIG_DIR` are false negatives that resurrect this exact bug (settings_merge.rs:292-293)

The substrings only match forward-slash, `.claude`-rooted forms. Two live-ish escape hatches:

- **Windows backslash paths**: `C:\Users\Jo\.claude\hooks\ways\check.sh` -> `entry_is_ours == false`. On a stale base that is precisely the false-revert this PR fixes, back for Windows users.
- **Relocated config dir**: Claude Code honors `CLAUDE_CONFIG_DIR`; `${XDG_CONFIG_HOME}/claude/hooks/ways/check.sh` -> `false` (no `.claude/` component).

The doc comment covers `${HOME}` vs expanded and quoting, and honestly names the `~/.claude` assumption — good — but does not name the separator or relocation failure modes. Consider normalizing separators before matching, and/or deriving the actual projection root rather than hardcoding `.claude`. At minimum, extend the doc comment so the Windows/relocation gap is a known, deliberate limitation rather than a surprise.

### 3. Latent coupling: a future ways hook that shells a projection binary other than `ways` is not recognized (settings_merge.rs:293)

`.claude/bin/ways` is not a substring of `.claude/bin/way-embed` or `.claude/bin/attend` (confirmed both -> `false`). Those binaries exist and are referenced in-tree (`tools/way-embed/*`). No **current** shipped hook invokes them — I checked `settings.json`; every hook command today lands under `.claude/hooks/`, `.claude/hooks/ways/`, or `.claude/bin/ways`, so there is no live miss — but the day someone adds a hook calling `${HOME}/.claude/bin/attend ...`, structural ownership silently fails and the stale-base false-revert returns for that hook.

Suggestion: broaden to `.claude/bin/` (the whole projected bin dir is ours) instead of the single `ways` name, and add a guard test asserting every hook entry in the shipped `settings.json` satisfies `entry_is_ours` — so this coupling breaks at CI, not in the field.

### 4. Interaction with `union_owned` / base byte-matching: sound, now partly redundant (no action)

Traced the merge filter (`:114-122`) and the self-audit (`:375-388`). `entry_is_ours` is applied identically on both sides of the invariant, so it cannot by itself introduce a NEW asymmetric (spurious) revert. It does now overlap `union_owned` (`:237-258`): both aim to strip an under-recorded ours-entry for hooks. No correctness issue; the two agree by construction. A one-line note that `union_owned` is now largely subsumed for *hooks* (perms still need it) would help the next reader.

### 5. Test adequacy

`update_that_changes_our_hooks_with_a_lost_base_converges` is well-targeted (old != new command, stale base, asserts no-duplication + user preservation). Gaps worth closing:

- **Pin the accepted false positive** (Finding 1): a user hook whose command references `.claude/hooks/` is dropped — assert the current behavior so the trade-off is intentional and any future tightening shows as a visible diff.
- **`.all()` semantics** (`:314`): a mixed entry (one ours command + one genuine user command in the same `hooks` array) must be **kept**. No test guards this today.
- **Direct unit tests** of `command_is_ours` / `entry_is_ours` — a small table test would document intent and catch the exe-vs-argument (Finding 1) and separator (Finding 2) cases cheaply; they are currently only exercised through `apply_to_files`.
- **Shipped-hooks guard** (Finding 3).

### What is good

- Doc comments are honest and name the base-staleness root cause and the `~/.claude` assumption.
- The fix keeps byte-matching first and adds the structural predicate as a backstop rather than replacing the mechanism.
- `entry_is_ours` is conservative on empty entries (`!cmds.is_empty()`, `.all()` requires every command) — the right default for a predicate that can delete user content.
- The regression test asserts convergence *and* non-duplication *and* user preservation, not just non-abort.

None of these block on correctness against current shipped hooks. Findings 1 and 2 are the ones I would want addressed (or explicitly deferred with a note) before this ships, since both can silently harm real user settings — 1 by deleting a referencing hook, 2 by re-breaking reconcile on Windows / relocated config dirs.

aaronsb:
Addressed the review in ffb7fba:

**1. Silent user-hook deletion (critical) — fixed.** `command_is_ours` now examines only the first (exe) token, quote-aware (new `exe_token` helper). A user hook that references a projection path as an *argument* (`tail -f ${HOME}/.claude/hooks/ways/x.log`) is no longer classified as ours. Pinned by `command_is_ours_examines_the_exe_token_only` and `merge_preserves_user_hook_referencing_our_path_as_argument`.

**2. Windows/relocation false-negative — fixed/documented.** Separators normalized (`\`→`/`) so Windows-form paths match (tested). The `~/.claude` (ADR-142) projection assumption — `CLAUDE_CONFIG_DIR` out of scope — is now documented *and* guarded by `shipped_hooks_are_all_recognized_as_ours`, which fails loudly if a shipped hook ever stops being recognized.

**3. `.claude/bin/ways` → `.claude/bin/` — broadened.** Sibling CLIs (`way-embed`, `attend`) are now recognized if a future hook shells them.

**4/5. Tests.** Added direct predicate units, the `.all()` mixed-entry semantics test (mixed entry preserved), the false-positive-preserved merge test, and the shipped-hooks guard. Suite: 247 pass (was 243).

Not changed: the base byte-matching/`union_owned` interaction (confirmed sound/symmetric by the review). Merging.

PR #250, opened 2026-07-02
feat(release): cut-release/publish-release helper (ADR-150 pt.2)

aaronsb:
## Summary

ADR-150 part 2 — the release-cadence helper, reworked to be **PR-first** (main is branch-protected, so a direct-to-main release commit can't work).

Two steps:
1. **`make cut-release COMPONENT=ways LEVEL=patch`** → creates a `release/ways-vX.Y.Z` branch with the version bump + a **verified** `Cargo.lock`, pushes it, opens a PR. No tag yet — the bump is reviewed/merged like anything.
2. (merge that PR, `git checkout main && git pull`) **`make publish-release COMPONENT=ways [PUSH=1]`** → tags `ways-vX.Y.Z` on the merged main. Without `PUSH=1` it creates the tag locally and prints the push command; `PUSH=1` pushes it → CI (`build-ways.yml`) builds every platform + `gh release create`.

## Addressed from review
- **Branch protection / direct-to-main** (moderate): reworked to the two-step PR-first flow above. No direct-to-main commit; the outward step is only the tag push.
- **Stale Cargo.lock behind `|| true`** (moderate): after `cargo update`, the script **verifies** the lock records the new version (`lock_version`) and **dies** if not — a tagged release can't carry a stale lock. `tag` also re-checks lock↔manifest agree (catches an unmerged/inconsistent bump).
- Tag guard now checks **local and origin** for an existing tag.

## Test plan
- `bash -n` clean; guards verified: refuses off-main, unknown component, bad mode/level.
- `awk` bump confirmed to touch only the package `version` line; version math verified (patch/minor/major).
- Will be exercised for real by cutting `ways-v1.0.1` (cut-release → merge → publish-release).
- way-embed out of scope (its own Makefile versioning).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ help:
 	@echo "  make test-multilingual  Verify multilingual way matching (18 languages)"
 	@echo "  make docs         Regenerate docs/cli/attend.md from the clap definition"
 	@echo "  make release      Build release binary for current platform"
+	@echo "  make cut-release  Bump+tag a component release (COMPONENT=ways LEVEL=patch [PUSH=1])"
 	@echo "  make uninstall    Remove ways from PATH"
 	@echo "  make clean        Remove build artifacts"
 	@echo "  make purge-attend-state  Wipe all attend runtime cache (peers, signals,"
@@ -348,6 +349,13 @@ release: ways-rebuild
 		cd dist && sha256sum ways-$$PLATFORM > ways-$$PLATFORM.sha256; \
 		echo "dist/ways-$$PLATFORM ($$(ls -lh ways-$$PLATFORM | awk '{print $$5}'))"; \
 		cat ways-$$PLATFORM.sha256
+
+# Cut a release for a Cargo-versioned component (ADR-150): bump version + tag
+# locally. Publishing is a separate, explicit push (or PUSH=1). Examples:
+#   make cut-release COMPONENT=ways LEVEL=patch          # local: bump 1.0.0 -> 1.0.1, tag
+#   make cut-release COMPONENT=ways LEVEL=patch PUSH=1   # also push main + tag (publishes)
+cut-release:
+	@bash scripts/release.sh "$(COMPONENT)" "$(LEVEL)" $(if $(PUSH),--push,)
 
 # --- Supporting ---
 
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,8 @@ help:
 	@echo "  make test-multilingual  Verify multilingual way matching (18 languages)"
 	@echo "  make docs         Regenerate docs/cli/attend.md from the clap definition"
 	@echo "  make release      Build release binary for current platform"
-	@echo "  make cut-release  Bump+tag a component release (COMPONENT=ways LEVEL=patch [PUSH=1])"
+	@echo "  make cut-release  Open a version-bump PR for a component (COMPONENT=ways LEVEL=patch)"
+	@echo "  make publish-release  After the bump PR merges: tag + publish (COMPONENT=ways [PUSH=1])"
 	@echo "  make uninstall    Remove ways from PATH"
 	@echo "  make clean        Remove build artifacts"
 	@echo "  make purge-attend-state  Wipe all attend runtime cache (peers, signals,"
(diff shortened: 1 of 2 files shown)


aaronsb:
## Review — release.sh safety & correctness

Reviewed with a focus on the outward git operations. I verified the version-bump and lock-sync behavior empirically in a throwaway `git worktree` (no changes to any tracked file). Overall the safety design holds up: **nothing pushes or publishes without `--push`**, and the four guards are correct. One actionable finding (the `|| true` lock masking), plus a couple of ordering/convention notes.

### Verified good

- **No unexpected push/publish.** Push happens only inside `if [[ "$PUSH" == "--push" ]]` (`scripts/release.sh:78-81`). Everything above line 78 is local. `PUSH="${3:-}"` (line 21) is the only way to set it, and the Makefile only passes `--push` when `PUSH` is set. Guards are all present and correct: on-main (`:36`), clean tree (`:38`), in-sync with origin (`:39-40`, and `git fetch origin main` does update the `origin/main` tracking ref on modern git, so the compare is fresh), tag-doesn't-exist (`:56`). `set -e` means a failed `fetch` or a rejected `git push origin main` aborts before anything further.
- **awk bump is correctly scoped.** Confirmed empirically: bumping `ways` 1.0.0→1.0.1 changed **only** line 3 (`^version = "..."`), leaving `clap = { version = "4" ... }` and every other dependency `version` field untouched. The `^version *= *"` anchor plus the `!done` one-shot guard (`:66-69`) means only the first column-0 package `version` is rewritten. `sub(/"[^"]+"/, ...)` replaces the first quoted token, which is the version. Solid, and the temp-file + `mv` sidesteps the BSD/GNU `sed -i` split — good portability call.
- **Version math.** `^([0-9]+)\.([0-9]+)\.([0-9]+)$` (`:45`) is strict; zeroing is correct (minor zeros patch; major zeros minor+patch, `:48-51`). No off-by-one. Non-numeric / pre-release versions (e.g. `0.1.0-rc1`) fail the regex and `die` — a safe refusal, worth being aware of if a component ever adopts a pre-release scheme.
- **Lock sync, happy path.** Confirmed: after the awk bump, `cargo update -p ways --precise 1.0.1 --offline` updated the `ways` entry in `tools/Cargo.lock` from 1.0.0→1.0.1 cleanly. The `--push` split, the printed publish commands, and the abort recipe all read correctly.

### Finding 1 (moderate) — `|| true` can commit a stale Cargo.lock

`scripts/release.sh:70-72`:
```sh
cargo update -p "$PKG" --precise "$NEXT" ... --offline >/dev/null 2>&1 \
  || cargo update -p "$PKG" ... --offline >/dev/null 2>&1 || true
git add "$MANIFEST" tools/Cargo.lock
git commit -q -m "release(${COMPONENT}): bump to ${NEXT}"
```
If **both** cargo invocations fail — no cargo on PATH, a cold `--offline` cache missing a needed dep, or any resolver error — `|| true` swallows it and the script commits + tags with `Cargo.toml` at the new version but `Cargo.lock` still pointing at the old one. Nothing between here and the tag verifies the lock actually moved.

Blast radius, given current CI: `build-ways.yml:57` runs `cargo build` **without `--locked`**, so CI silently re-resolves and the built binary is still correct. So this isn't release-breaking today — but the **tagged commit carries an internally-inconsistent lockfile**, which is exactly the kind of silent drift this helper exists to prevent, and it'd bite anyone doing `--locked`/`--frozen` later.

Suggestions (any one closes it): after the cargo step, assert the lock moved and `die` otherwise, e.g. verify `tools/Cargo.lock` contains the new version under the package stanza; and/or drop the blanket `|| true` and handle the genuinely-lockless checkout as its own explicit case; and/or add `--locked` to the build workflows so drift can't ship unnoticed.

### Finding 2 (minor) — tag-exists guard is local-only; main is pushed before the tag

- `:56` checks `refs/tags/${TAG}` locally only, and `:39` fetches the branch, not tags. A tag already existing on origin (but not fetched) passes the guard. With `--push`, `git push origin main` (`:80`) runs and succeeds **first**, then the tag push (`:81`) fails and `set -e` aborts — leaving main published without its tag.
- Independently: `build-ways.yml` triggers on `push: branches:[main]` with a path filter that includes `tools/ways-cli/**`, `tools/Cargo.toml`, `tools/Cargo.lock` (`:5-11`). The release commit touches exactly those, so **the main push alone triggers a full platform build** before the tag is pushed — then the tag push triggers it a second time. The Release job is correctly gated `if: startsWith(github.ref, 'refs/tags/ways-v')` (`:117`), so no duplicate Release, but it's a redundant all-platform build, and the irreversible action (main push) goes out ahead of the tag the PR frames as "the publish."

Consider fetching tags in the sync step (so a remote tag collision is caught pre-commit), and/or pushing the tag with (or before) main — or just documenting that the main push is itself the point of no return.

### Finding 3 (convention) — direct-to-main release commit

The bump commits straight to `main` and pushes `main` with no PR. For a deterministic, fully-guarded mechanical version bump that's a defensible release convention. Two things to note: (a) if `main` has branch protection requiring PRs, `git push origin main` (`:80`) is rejected — but the local commit+tag are already made (`:74-75`), leaving the branch diverged and recoverable only via the printed abort recipe; (b) it deliberately steps around this project's own PR discipline. Worth an explicit line in ADR-150 that release commits push directly to `main` (and that main must permit it), so the exception is on the record rather than implicit.

### Notes
- `set -u` handling is clean — `${3:-}` (`:21`) is the only unset-risk positional and it's guarded; `$# -ge 2` gates the rest. Quoting is consistent throughout.
- Scoped `git add "$MANIFEST" tools/Cargo.lock` (`:73`) plus the clean-tree guard means the commit can't sweep in unrelated changes. Good.

No security concerns (no secrets, no eval of external input; component/level are allowlisted via `case`). Finding 1 is the one I'd resolve before cutting `ways-v1.0.1` for real; 2 and 3 are judgment calls worth a sentence in ADR-150.

PR #253, opened 2026-07-02
feat(settings): ways settings show — inventory + per-key detail (ADR-147)

aaronsb:
## Summary

Adds `ways settings show`, the read-only human view of the settings fragment store — the UAT gap you flagged: `lint` is pass/fail, `compile` is raw JSON, `project` is a diff, but nothing just *shows* you the store.

- **`ways settings show`** — inventory: fragments grouped by scope, one line per top-level key (`key · compact value · one-line why`), aligned; per-scope footer (`→ target · N keys differ from live`) + a lint verdict line.
- **`ways settings show <key>`** — manpage-shaped: **NAME** (scope · type), **VALUE** (+ contributing fragment via provenance), **SCHEMA** (official description, wrapped), **WHY** (rationale body, leading heading stripped), **STATUS** (live-matches / pending-change / pending-add / reconciler-owned / managed-blob).

Composes what the other verbs already own (fragment loader, `schema_doc`, `compile::merge_scope` + provenance, live `settings.json`) — no re-derivation. Read-only; writes nothing.

## Test plan
- 9 unit tests on the pure helpers: why-extraction (heading/blank skip + fallback), value compaction, truncate, leading-heading strip, provenance sources (scalar + nested), wrap, status branches, lint summary. **259 pass** (+9); clippy clean.
- Smoke-tested both views against a throwaway 2-fragment store (see below).

```
USER (2 fragments)
    cleanupPeriodDays  30    Keep a month of session history — long enough to…
    model              opus  Default to the strongest model; this is the driver's…
  → ~/.claude/settings.json  ·  1 key differs from live
lint: 1 warning
```

No new ADR — additive, read-only, under ADR-147's CLI shape.

### DIFF
--- a/tools/ways-cli/src/cmd/settings/mod.rs
+++ b/tools/ways-cli/src/cmd/settings/mod.rs
@@ -19,6 +19,7 @@ pub mod project;
 pub mod scaffold;
 pub mod schema;
 pub mod schema_doc;
+pub mod show;
 pub mod source;
 
 /// `ways settings schema` — report the settings schema and its (configurable)
--- a/tools/ways-cli/src/cmd/settings/project.rs
+++ b/tools/ways-cli/src/cmd/settings/project.rs
@@ -22,8 +22,10 @@ use serde_json::{Map, Value};
 use std::collections::{BTreeSet, HashMap};
 use std::path::{Path, PathBuf};
 
-/// Keys the reconciler co-owns — `project` must not touch them (MVP).
-fn reconciler_owned(key: &str) -> bool {
+/// Keys the reconciler co-owns — `project` must not touch them (MVP). Single
+/// source of this ownership fact: `show` consults it too (for a key's live
+/// status), so the two components cannot diverge.
+pub(crate) fn reconciler_owned(key: &str) -> bool {
     matches!(key, "hooks" | "permissions")
 }
 
--- a/tools/ways-cli/src/cmd/settings/show.rs
+++ b/tools/ways-cli/src/cmd/settings/show.rs
@@ -1,0 +1,470 @@
+//! `ways settings show` — a human-readable view of the fragment store (ADR-147).
+//!
+//! Read-only, writes nothing. Two modes:
+//!
+//! - **inventory** (`ways settings show`) — every fragment grouped by scope, one
+//!   line per top-level key (key · value · one-line rationale), with a per-scope
+//!   footer: where it projects, the lint verdict, and how many keys differ from
+//!   the live `settings.json`.
+//! - **detail** (`ways settings show <key>`) — a manpage-shaped page for one key:
+//!   NAME (scope · type), VALUE (+ source fragment), SCHEMA (the official
+//!   description), WHY (the fragment's rationale body), STATUS (live/pending).
+//!
+//! It *composes* the pieces the other verbs already own — the loader
+//! ([`super::fragment`]), the schema ([`super::schema_doc`]), the compile merge +
+//! provenance ([`super::compile`]), and the live `settings.json` — rather than
+//! re-deriving any of them.
+
+use super::compile::{self, Provenance};
+use super::fragment::{load_dir, Fragment, Scope};
+use super::schema_doc;
+use anyhow::Result;
+use serde_json::{Map, Value};
+use std::path::Path;
+
+/// Keys the reconciler co-owns — `project` skips them, so their live status is
+/// "managed by the reconciler", not pending. Mirrors `project::reconciler_owned`.
+fn reconciler_owned(key: &str) -> bool {
+    matches!(key, "hooks" | "permissions")
+}
+
+pub fn run(store: &Path, key: Option<String>) -> Result<()> {
+    if !store.is_dir() {
+        println!("No settings store at {} yet.", store.display());
+        println!("Author one with: ways settings new <key> --dir {}", store.display());
+        return Ok(());
+    }
+    let frags = load_dir(store)?;
+    match key {
+        Some(k) => show_key(store, &frags, &k),
+        None => show_inventory(store, &frags),
+    }
+}
+
+// ── inventory ──────────────────────────────────────────────────
+
+fn show_inventory(store: &Path, frags: &[Fragment]) -> Result<()> {
+    println!("{}", store.display());
+    if frags.is_empty() {
+        println!("\n  (empty — no fragments yet; `ways settings new <key>` to start)");
+        return Ok(());
+    }
+
+    let schema = schema_doc::active();
+    let findings = super::lint::check(frags, schema);
+    let lint_verdict = lint_summary(&findings);
+
+    for scope in [Scope::User, Scope::Project, Scope::Managed] {
+        let scoped: Vec<&Fragment> = frags.iter().filter(|f| f.scope == scope).collect();
+        if scoped.is_empty() {
+            continue;
+        }
+        println!("\n{} ({} fragment{})", scope.as_str().to_uppercase(), scoped.len(), plural(scoped.len()));
+
+        // Widest top-level key, for column alignment.
+        let width = scoped
+            .iter()
+            .flat_map(|f| top_keys(f))
+            .map(|k| k.len())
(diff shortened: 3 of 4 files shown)


aaronsb:
## Review: `ways settings show` (ADR-147)

Read-only presentation verb, ~483 additive lines in one new file + thin wiring. Nothing writes; degradation on absent store/schema is handled. All findings below are non-blocking (read-only, warn-but-allow), but **F1** is the one I'd want addressed before merge because it can misrepresent what `project` would do.

### F1 — `pending_count` claims a lint gate it never runs (correctness + doc drift)

**Location**: `show.rs` `pending_count` (doc comment: *"None if the scope doesn't compile cleanly (lint gate) — the footer stays quiet"*).

**Problem**: The function calls `compile::merge_scope(scoped)` directly and returns `Some(n)`. The only `None` path is `baked.as_object()?`, and `merge_scope` always yields a `Value::Object`, so it never fires. There is no `lint::has_errors` gate here — unlike `compile::compile_store`, which refuses (`Outcome::Refused`) when the store has lint errors, and `project`, which builds on it.

**Why it matters**: For a store with lint *errors*, `project`/`compile` refuse to touch anything, yet the per-scope footer still prints `· N keys differ from live` / `· in sync with live` — implying a projection outcome that would not happen. The inventory even prints a separate `lint: N error(s) — compile/project are gated` line, so the two lines contradict each other. `live_status` has the same property (it reports `pending — project would change it` regardless of lint state).

**Suggestion**: Either gate the footer through `compile::compile_store(...)`/`lint::has_errors` (returning `None` on error, as the comment promises), or drop the false claim from the comment and suppress the footer's differ/in-sync clause when the lint verdict is `error`.

### F2 — `live_status` / `pending_count` read the live file internally → untestable core (testability / DIP)

**Location**: `show.rs` `live_status`, `pending_count`, which both call `live_settings(scope)` internally (FS + `CLAUDE_PROJECT_DIR` env).

**Problem**: The interesting branches — matches / pending-change / pending-add — can only be reached by mutating process-global state, so the unit test explicitly only covers the Managed + reconciler-owned branches. `project.rs` solved exactly this by threading `base_file`/`target` into `project_write` *"so tests supply their own base path rather than mutating the process-global \$XDG_STATE_HOME."* That seam wasn't applied here.

**Suggestion**: Split a pure core that takes the live object/value as a parameter (mirror `project_write`), and let `run` do the FS read. Then the three status branches and `pending_count`'s diff math become unit-testable.

### F3 — `pending_count` can't see removals (divergence from `project`)

**Location**: `show.rs` `pending_count` vs `project.rs` `project_write` (keys iterated as `ours.keys().chain(base_obj.keys())`).

**Problem**: `project` counts a key that's in its persisted base but no longer in the fragments as a change (`remove {key}`). `show` has no base file and only diffs baked-vs-live, so after you delete a fragment, `show` can report `in sync with live` while `project` would still remove the stale key. The footer's number is therefore a lower bound, not the projection delta.

**Suggestion**: Read the scope's `base_path` (as `project` does) to include removals, or reword the footer to `N keys to add/update` so it doesn't claim full parity with what `project` would do.

### F4 — `reconciler_owned` duplicated verbatim from `project.rs` (drift risk)

**Location**: `show.rs` `reconciler_owned` — identical `matches!(key, \"hooks\" | \"permissions\")` to `project.rs`.

**Problem**: The codebase already treats this class of shared law as single-sourced: `compile::is_concat_path` is explicitly documented as *"the single source of the concat law … the two components cannot diverge."* Copying `reconciler_owned` into `show` is the opposite pattern — if `project`'s owned set changes, `show`'s STATUS and footer silently misreport.

**Suggestion**: Promote one definition (e.g. into `compile` or a small shared module) and have both `project` and `show` import it, matching the `is_concat_path` precedent. The in-code comment already admits it's a mirror — make it a reference instead.

### F5 — the detail "overridden" note is wrong for object/concat keys

**Location**: `show.rs` `show_key`, the `setters.len() > 1` branch → prints `set by N fragments — last in filename order wins`.

**Problem**: That statement is only true for scalar-override keys. Per `compile.rs`, objects deep-merge and `permissions.allow/deny/ask/additionalDirectories` concat+dedupe. For `permissions` set across two fragments, `top_keys` yields `permissions` for both → `setters.len() == 2` → the view claims last-wins, contradicting the documented merge law (both contribute; lists union).

**Suggestion**: Gate the note on the compiled value being a scalar, or reword for objects/concat-lists (e.g. `merged from N fragments`). The provenance you already print (`from: …`) is the accurate signal.

### F6 — `wrap` measures width in bytes while `truncate` measures chars (consistency)

**Location**: `show.rs` `wrap` uses `cur.len()` / `word.len()` (byte lengths); `truncate` uses `s.chars().count()` / `chars().take(...)`.

**Assessment**: No panic — `wrap` only pushes whole words and never slices mid-char, and a single over-width word lands alone (no infinite loop). But multibyte schema descriptions wrap short of `width`, inconsistent with the char-based `truncate`. Since the review flagged multibyte specifically: I confirmed **no byte-boundary slicing anywhere** — `truncate` is char-safe. `wrap` is just measuring in the wrong unit. Use `.chars().count()` for parity.

### F7 — nested-only lookup message is misleading (nit)

`show permissions.allow` reports `No fragment sets 'permissions.allow' … isn't in the current schema`, even though it's effectively set (nested) and `permissions` is in the schema. `show` only addresses top-level keys. A one-line hint ("only top-level keys are addressable; try `permissions`") would avoid the false-negative read.

### F8 — whitespace-only schema description prints an empty SCHEMA block (nit)

`Some(i) if !i.description.is_empty()` passes on a whitespace-only string; `wrap` then yields an empty vec, so the `SCHEMA` header prints with no body. Harmless and unlikely from the vendored schema; `.trim().is_empty()` would close it.

### What's solid

- **No byte-boundary slicing** — `truncate` and the two heading helpers are char-based; safe on multibyte/unicode, no panic on empty/whitespace input (verified `one_line_why("")`, empty body, heading-only body → clean fallbacks).
- **Malformed/empty store**: non-dir and empty-store handled explicitly; a malformed fragment propagates via `load_dir` exactly as `compile`/`lint` do — consistent degradation, not a new panic path. Multi-scope keys render one page per scope. Nested keys (`permissions`) resolve via `top_keys` + provenance without crashing.
- **`key_sources`** correctly mirrors the dotted `Provenance` and uses a proper `key.` boundary, so `model` won't falsely match `modelX`. Scalar + nested cases are tested.
- **Consistency with siblings**: `--path` flag name and the `config_root().join(\"settings\")` default match `compile`/`project`/`lint`; schema-absent path returns the documented degraded messaging.
- **Helper test coverage** is good for the pure functions (9 tests). The gap is the impure core (F2) and the untested `pending_count`/lint-gate claim (F1).

No prompt-injection or embedded-instruction content in the diff or PR description.

---
*AI-assisted review via Claude Code (code-reviewer). Read-only — no files changed.*

aaronsb:
Addressed all eight findings in the latest commit.

- **F1** (the one to fix) — the footer no longer prints a pending count for a store with lint **errors**; it says `project gated until lint errors are fixed`, consistent with the `lint:` line. Verified: a type-error store now shows the gate. (Warnings still show the count — they don't block project.)
- **F2** — `live_status` and `pending_of` are pure (live object injected by the caller); all match/pending/add + change/removal branches are now unit-tested, mirroring `project.rs`'s `base_file` DI.
- **F3** — `pending_count` reads this scope's projector base and counts base-tracked keys `project` would **remove** (live still equals what we wrote), not just value diffs.
- **F4** — `reconciler_owned` promoted to `project::reconciler_owned` (pub(crate)); the duplicate is gone. Single source, per the `is_concat_path` precedent.
- **F5** — multi-fragment note now states the real merge law (objects deep-merge, permission lists union, scalars last-wins).
- **F6** — `wrap` is char-based (matches `truncate`).
- **F7** — a dotted lookup points at the top-level key (`permissions.allow` → try `permissions`).
- **F8** — whitespace-only description treated as absent.

260 tests pass; clippy clean. Merging.

PR #260, opened 2026-07-02
refactor(tools): extract ways-core library crate (ADR-151 §1)

aaronsb:
## Summary

Implements **ADR-151 §1** — extracts the shared engine out of the bin-only `ways` crate into a new **`ways-core`** library crate. Both `ways` and the forthcoming `ways-audit` binary will depend on one library rather than the engine being trapped inside an executable (the sensor-trait / agent-fmt house pattern).

This is the first of the ADR-151 PR series:
- **PR A (this):** extract `ways-core` — pure engine cluster + tests
- PR B: `ways-audit` sibling binary; migrate the manifest builder + firing-log reader into `ways-core`; core `ways` sheds `governance`
- PR C: finding pipeline, claim-authoring assist, `satisfied_when` determination field
- PR D: `ways-audit-v*` release machinery

## What moved

The engine is a **closed cluster** with no upward edges into `cmd/`: `util`, `paths`, `frontmatter`, `scanner`, `config`, `agents` (+ the `languages.json` asset). Verified there is no genuine engine→cmd compile dependency — only a stale rustdoc intra-doc link in `config.rs`, retargeted here.

The `ways` binary re-exports the modules at its crate root:
```rust
pub use ways_core::{agents, config, frontmatter, paths, scanner, util};
```
so every `crate::util::…` / `crate::paths::…` path across the ~40 command modules keeps resolving **unchanged** — zero churn in `cmd/`. Files moved with `git mv` to preserve history.

## Why now (independent of compliance work)

The library boundary is exactly what makes the previously-untested engine unit-testable. `scanner.rs` (way discovery) had **zero** tests; this PR adds 6 covering identity derivation, `.check.`/non-md skipping, description-gating, sort order, and empty/missing roots.

## Test plan

- `ways-core`: **75 tests green** (69 that moved with the modules + 6 new scanner tests), `cargo clippy` clean.
- `ways` binary: **191 unit + 7 integration green**, clippy clean, `ways --version` and `ways governance report` run.
- Full workspace builds.
- The 9 `session_sim` integration failures are **pre-existing and environment-dependent** (need a live ways projection); verified byte-identical on clean `main`, so unrelated to this change.
- `ways governance` still runs from the core binary, untouched.


aaronsb:
## Review: extract `ways-core` (ADR-151 §1)

Verified against a local checkout of the branch head (`d21a026`): full workspace builds clean, `ways-core` = 75 tests green (incl. the 6 new scanner tests), `ways` binary compiles against the re-export shim. This is a clean, well-scoped extraction. Details below, organized by the risk areas that matter for a crate split.

### 1. Behavioral equivalence of the re-export shim — verified sound
The shim in `main.rs` (`pub use ways_core::{agents, config, frontmatter, paths, scanner, util};`) preserves every `crate::…` path. I enumerated all 9 intra-engine `crate::` references in the moved modules and every one resolves to a sibling that is declared in `ways-core/src/lib.rs`:
- `paths.rs:26` → `crate::util`
- `frontmatter.rs:59` → `crate::config`
- `util.rs:155` → `crate::agents`
- `agents/mod.rs:26` → `crate::config`
- `agents/claude_code.rs:9` → `crate::util`
- `config.rs:157,291,322` → `crate::paths`; `config.rs:330` → `crate::util`

On the binary side, `crate::frontmatter` / `crate::scanner` / etc. are consumed across `cmd/` (graph, corpus, tune, language, lint, `session/engagement.rs`) and all resolve through the re-export — confirmed by the clean build, not just by eye.

### 2. Boundary correctness — the closed-cluster claim holds
Grepped the entire moved set for upward edges into `cmd`/`session`/`main` (`crate::cmd`, `crate::session`, `super::cmd`, `super::session`, `crate::main`): **zero matches**. The PR's claim that only a stale rustdoc intra-doc link needed retargeting is accurate — that's the `config.rs` `[crate::cmd::settings::source]` → prose change in the diff. No missed engine→cmd compile dependency. This is a genuinely closed cluster.

### 3. `languages.json` asset move — path resolves
`agents/mod.rs:47` is `include_str!(\"../../languages.json\")`. From `ways-core/src/agents/mod.rs`, `../../` is the crate root `ways-core/`, and the asset was `git mv`d to `ways-core/languages.json`. Resolves correctly; the clean `ways-core` build (which would fail at compile time on a bad `include_str!`) confirms it.

### 4. Visibility changes — correct and appropriately minimal
- `frontmatter`/`scanner` private `mod` → `pub` in `lib.rs`: **required** to cross the crate boundary for the re-export; there's no narrower option. On what is otherwise a bin-only consumer this widening has no external-API cost.
- `disabled_ways`: the field is `pub(crate)` in `config.rs:46` with a `pub fn disabled_ways(&self) -> &[String]` accessor (`config.rs:132`). This is the *better* design, not just a mechanical fix — keeping the field crate-private preserves the documented \"only the project overlay may populate this\" invariant while still exposing reads. The `disable.rs:56` test swap (`.disabled_ways` → `.disabled_ways()`) is the only call site that needed it. I checked the rest of the binary for direct field access: the only remaining `.disabled_ways` hit is a doc comment at `session.rs:557`, not code. Good.

### 5. New scanner tests — meaningful, not superficial
They exercise real discovery behavior against a real on-disk tree: identity derivation (`softwaredev/code/quality` → domain/id split), top-level id→domain fallback, description-gating (both \"frontmatter closes without description\" and \"no frontmatter\"), `.check.`/non-`.md` skipping, sort order, and empty/missing root. This is the right coverage for the previously-untested unit.

Two minor, non-blocking notes:
- Cleanup runs *after* the asserts, so a failing test leaks its temp dir. It's self-healing (`scratch()` does `remove_dir_all` on the next run), so this is cosmetic — a `Drop` guard would be tidier but isn't worth churn here.
- `empty_or_missing_root_yields_no_ways` cleverly reuses one path for both the empty and the removed/missing case; the comment makes the intent clear. Fine as-is.

### 6. `ways-core/Cargo.toml` — deps are exact, nothing over-scoped
I cross-checked every declared dependency against actual engine usage: `anyhow` (4), `serde` (12), `serde_json` (13), `serde_yaml` (11), `walkdir` (1), and `sensor-trait` (1, via `use sensor_trait::Curve` in `frontmatter.rs:2`). All six are used; none is missing; the heavier bin-only deps (`rust-stemmers`, `crossterm`, `clap`, `agent-fmt`) correctly stayed in `ways-cli`. The `[lib]` target is right.

One thing to confirm, not necessarily change: `serde_json` carries `features = [\"preserve_order\"]`, inherited from the binary. It's justified if any engine-emitted JSON is order-sensitive (manifest); worth a mental note that the two crates should keep that feature consistent as the manifest builder migrates in PR B.

### One forward-looking observation (not blocking this PR)
The extraction gives `scanner` a genuinely clean, pure-function test seam. But `frontmatter.rs:59` and `agents/mod.rs:26` reach into `config::global()`, which is a process-wide `LazyLock<Config>` singleton (`config.rs:16`). That global moved wholesale into `ways-core` — so `ways-audit` (PR B) will inherit and share its initialization semantics, and the parts of the engine that read it remain harder to unit-test in isolation than the scanner is. Nothing to do here; flagging it because \"the library boundary makes the engine testable\" is true for the pure paths and only partially true for the config-coupled ones. Worth keeping in view as the claim model and firing-log reader land in the shared crate.

**Assessment:** No blocking issues. The behavioral-equivalence, boundary, asset-path, and dependency claims in the description all check out under direct inspection and a clean build. Approve-worthy as the PR-A foundation of the series. (Per instructions I did not evaluate the pre-existing `session_sim` env-dependent failures.)

PR #261, opened 2026-07-02
feat(tools): ways-audit sibling binary; shed compliance from core ways (ADR-151 §2+§4)

aaronsb:
## Summary

Implements **ADR-151 §2 + §4** (the release wiring folded in per the 3-PR plan). The compliance operator surface moves out of the core `ways` runtime into a new, deliberately-invoked **`ways-audit`** binary — the same relationship `/ship` or `/wrap` have to the tool they act on.

Second of the ADR-151 series (PR-A `ways-core` extraction merged as #260). Character: **mechanical relocation + rename + packaging** — no new capability. Session-derived *findings* and the `satisfied_when` assessability field are the net-new next layer (**PR-C**), deliberately not here.

## What moved

**`ways-core`** (the shared engine) gains the compliance engine per §1:
- `provenance` — the claim (`provenance.yaml`) sidecar model + in-memory manifest builder (from `cmd/provenance.rs`), `Value`-based at parity
- `firing` — the way-firing-event log reader (`load_events` / `count_fires`)
- `util::days_to_ymd` — the pure calendar helper both `session.rs` and the audit date math need (drops the old `session::days_to_ymd_pub` wrapper)

All three ship with unit tests.

**`ways-audit`** (new binary → depends on `ways-core`) owns the operator surface: `report / trace / control / policy / gaps / stale / active / matrix / lint`, migrated from `ways governance` and reframed to the ADR-200 compliance vocabulary (**claim**, not provenance).

Core **`ways`** sheds the `governance` and `provenance` subcommands and goes back to being the runtime. The structural sidecar lint stays in `ways lint` (a file-format check, not the compliance surface).

## Distribution (§4)

- `scripts/release.sh` — `ways-audit` registered as a releasable component (`ways-audit-v*` tags)
- `.github/workflows/build-ways-audit.yml` — its own 4-platform build + release CI job
- `make ways-audit` — on-demand build/link target, kept **out** of default `setup`/`install` so the core install stays lean (honoring the "optional, deliberately-invoked" framing)
- `build-ways.yml` now also triggers on `tools/ways-core/**` (a correctness fix — `ways` depends on it)

## Docs

Swept 10 docs/guidance/skill files off the removed commands onto `ways-audit`; the "not yet shipped" framing now correctly applies only to *findings*, not to `ways-audit`. ADRs left as-decided (historical records).

## Known / out of scope

The claim lint resolves policy URIs under `~/.claude`, but `governance/` isn't projected there post-ADR-142 — a **pre-existing** latent lint-path bug, preserved exactly (a mechanical move shouldn't change behavior). Flagged for a separate fix.

## Test plan

- `ways-core` **80** + `ways-audit` **3** + `ways` **191 unit / 7 integration** tests green
- `cargo clippy -D warnings` clean on both new crates
- `doc lint` 0/0; `ways lint` clean on the changed way
- `ways-audit` exercised end-to-end against the real 109-way corpus (report/matrix/lint render with the new claim vocab)
- release wiring verified: `make -n ways-audit`, `release.sh` component check, workflow YAML validated
- pre-existing `session_sim` env-failures unchanged (verified on `main` in PR-A)


aaronsb:
## Code Review — ADR-151 §2+§4 (ways-audit sibling + shed from core)

Reviewed as a **mechanical move + rename + packaging** change (no new capability). Verdict: **behavior-preserving, no blocking issues.** Findings below are low-severity; the parity claims all hold up under inspection.

### Behavioral parity — PASS

- **Import retargeting is correct.** `audit.rs` splits the old `super::helpers::{count_fires, cutoff_date, find_stale_ways, load_events}` into `crate::helpers::{cutoff_date, find_stale_ways}` + `ways_core::firing::{count_fires, load_events}`; `report.rs` → `crate::helpers`; `trace.rs` → `ways_core::firing::load_events`. All resolve.
- **The vocab rename (`provenance`→`claim`) only touches `println!` display strings.** No `json!` output keys changed in `report/trace/query/audit/matrix/lint`. The JSON contract consumed by `--json` / `jq` is intact.
- **`days_to_ymd` relocation is byte-for-byte identical.** Diffed `tools/ways-core/src/util.rs:185-201` against the original `session.rs` body — no transcription drift in the Hinnant calendar algorithm. `session.rs:522` `chrono_utc_now` now calls `crate::util::days_to_ymd`, which resolves through `pub use ways_core::{…util}` at `main.rs:8`. The `days_to_ymd_pub` wrapper is gone with no remaining callers.
- **`ways-core` boundary is clean.** `provenance.rs` reaches only `crate::util::{has_frontmatter, home_dir}`; `firing.rs` reaches only `crate::paths::events_log`. Neither touches `cmd`/`session` (grep-confirmed). The manifest builder matches the old `cmd/provenance.rs` — only the CLI `run` printer was dropped.
- **Shed from core is clean.** No dangling references to `cmd::provenance`, `cmd::governance`, `days_to_ymd_pub`, `GovernanceCommand`, `Commands::Governance/Provenance` anywhere in `ways-cli` (grep-confirmed). `cmd/mod.rs` module list and `main.rs` dispatch both consistent.

### Release wiring — PASS

`build-ways-audit.yml` faithfully mirrors `build-ways.yml`: `-p ways-audit`, `ways-audit-v*` tags, correct binary name / artifact paths / checksums, `test` job covers `-p ways-audit -p ways-core`, and it correctly omits the `session_sim` job (no such tests here). `build-ways.yml` correctly gains the `tools/ways-core/**` trigger. `make ways-audit` is sound and deliberately kept out of `setup`/`install`. `release.sh` registers `ways-audit` with the right manifest and tag prefix.

### Docs — PASS

`docs/governance.md` and `docs/hooks-and-ways/provenance.md` now present `ways-audit` as shipped while keeping *findings* / `satisfied_when` framed as not-yet-built (`governance.md:296-299`, `:365-366`; `provenance.md:409-411`). No over-claim that findings exist. Zero lingering `ways governance` / `ways provenance` command strings outside the ADRs (grep-confirmed) — the 10-file sweep is complete.

### Pre-existing lint-path bug — CONFIRMED pre-existing

`tools/ways-audit/src/lint.rs:72` resolves policy URIs as `format!("{home}/.claude/{uri}")`. Lines 60-80 are **byte-identical** to `origin/main:tools/ways-cli/src/cmd/governance/lint.rs` — this PR did not introduce or alter it. Correctly preserved as-is and flagged for a separate fix, per the mechanical-move principle. Agreed it should not be fixed here.

---

### Low-severity notes (non-blocking)

1. **`generator` JSON value changed** — `tools/ways-core/src/provenance.rs:65` flips `\"generator\": \"ways provenance\"` → `\"ways-audit\"`. This is the one genuine JSON-*value* change in an otherwise parity-preserving move. Verified it has no in-repo consumers (set, never read), and the new value is arguably more correct now that `ways-audit` owns the surface — just flagging it since the PR frames itself as behavior-preserving. Fine to keep.

2. **Stale comment in `release.sh`** — the `PKG=\"$COMPONENT\"` comment still reads \"package name == component name for all three\" (now four components). Cosmetic.

3. **`ways-audit` with no subcommand errors (pre-existing design).** clap requires a subcommand; `Report` is documented \"(default)\" but not wired as a clap default, so bare `ways-audit` errors rather than running the coverage report. `commands/governance.md` says \"if no arguments, run the default coverage report\" — the agent must pass `report` explicitly. This mirrors the old `ways governance` structure (same required-subcommand shape), so it's not a regression. If a true default is wanted, `command: Option<Command>` defaulting to `Report` would do it — but that's a capability tweak, out of scope for this move.

Tests reported green (ways-core 80 / ways-audit 3 / ways 191+7), clippy clean. I did not re-verify the build locally; the structural review is what's above.

PR #262, opened 2026-07-02
feat(audit): assemble classifier-ready finding dataset (ADR-201, ADR-151 §3)

aaronsb:
## Summary

The **net-new assessability layer** — the third and final ADR-151 PR (PR-A #260, PR-B #261 merged). It closes the question ADR-200 §1 left open: *who fills a finding's determination, and when.*

Introduces **ADR-201** (new, Accepted): `ways-audit` **assembles** findings; it **never determines** them. A finding is one row of a classifier-ready dataset — the `satisfied_when` criterion + firing evidence (*features*) beside an **empty** `determination` (*label*) and empty provenance slots. The classifier that labels it is a **separate, undesigned, out-of-scope system** (human, deterministic check, or model). The tool's guarantee is purely structural: a clean, uniform, unambiguous-to-label record. **No code path in `ways-audit` writes a determination** — the honesty boundary that keeps this from manufacturing the evidence the claim/finding split exists to protect.

## `ways-core`

- **`provenance`** — typed `Claim`/`Policy`/`Control` schema (ADR-151 §3). `Control` carries the new **`satisfied_when`** determination criterion. `from_provenance_value` tolerates the legacy bare-string control form.
- **`finding`** — `Finding`/`FiringEvidence`/`Tier` + `assemble()`, a **pure function** (`manifest + events + timestamp → rows`). Every `Option` serializes explicitly so the dataset has a **uniform column set**. Process vs. outcome tier follows criterion presence — never a determination.
- `util::now_utc` (chrono-free RFC3339); `paths::findings_ledger` (`$XDG_STATE`).

## `ways-audit`

- **`assemble [--way <id>] [--write]`** — build the dataset, print it (`--json` = the machine dataset), optionally append to the ledger.
- **`findings`** — read the ledger back.

## Docs

`satisfied_when` graduates from ADR-200 §1's "proposed, not yet in schema" note into the actual schema (`governance.md`, `provenance.md`, `ways-cli.md`). Findings reframed from "not yet built" → "assembled; classification out of scope." ADR-200/151 gain reciprocal ADR-201 graph edges.

## Honest scope

- The tool ships a **dataset, not an assessment engine**. Classification is a separate effort on a separate schedule — deliberately not designed here (ADR-201 §3).
- All 51 rows assembled from the real corpus are **process-tier**, because no sidecar carries `satisfied_when` yet. That's correct: existing claims aren't outcome-assessable until authors seed criteria (task #9).
- A dataset with a perpetually-empty label column is **inert until something classifies it** — this PR does not deliver that something, by design.

## Test plan

- `ways-core` **84** (+5) + `ways-audit` **3** tests green; the assembler logic (in `ways-core::finding`) has 4 tests, incl. an assertion that assembly **never** produces a determination.
- `cargo clippy -D warnings` clean on both crates; `doc lint` 0/0; `adr lint` clean (pre-existing `Rejected`-status warning only).
- `ways-audit assemble` exercised end-to-end against the real 109-way corpus (51 rows, uniform null-label schema); ledger `--write` → `findings` round-trips.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -71,14 +71,15 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-148](./system/ADR-148-framework-surface-ships-operator-content-dev-harness-in-project-scope.md) | framework surface ships operator content; dev harness in project scope | Draft |
 | [ADR-149](./system/ADR-149-operator-config-interview-skill.md) | operator config interview skill | Accepted |
 | [ADR-150](./system/ADR-150-version-truth-and-downgrade-safe-self-update.md) | Version-truth and downgrade-safe self-update | Accepted |
-| [ADR-151](./system/ADR-151-extract-ways-core-crate-and-ways-audit-sibling-binary.md) | Extract ways-core crate and ways-audit sibling binary | Draft |
+| [ADR-151](./system/ADR-151-extract-ways-core-crate-and-ways-audit-sibling-binary.md) | Extract ways-core crate and ways-audit sibling binary | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
 
 | ADR | Title | Status |
 |-----|-------|--------|
-| [ADR-200](./governance/ADR-200-compliance-claims-and-session-derived-findings.md) | Compliance claims and session-derived findings | Draft |
+| [ADR-200](./governance/ADR-200-compliance-claims-and-session-derived-findings.md) | Compliance claims and session-derived findings | Accepted |
+| [ADR-201](./governance/ADR-201-findings-assembled-as-classifier-ready-assessment-records.md) | Findings assembled as classifier-ready assessment records | Accepted |
 
 ## Documentation
 _Documentation structure, tooling, coherence_
--- a/docs/architecture/governance/ADR-200-compliance-claims-and-session-derived-findings.md
+++ b/docs/architecture/governance/ADR-200-compliance-claims-and-session-derived-findings.md
@@ -10,6 +10,7 @@ related:
   - ADR-110
   - ADR-111
   - ADR-151
+  - ADR-201
 supersedes:
   - ADR-005
 ---
--- a/docs/architecture/governance/ADR-201-findings-assembled-as-classifier-ready-assessment-records.md
+++ b/docs/architecture/governance/ADR-201-findings-assembled-as-classifier-ready-assessment-records.md
@@ -1,0 +1,137 @@
+---
+status: Accepted
+date: 2026-07-02
+deciders:
+  - aaronsb
+  - claude
+related:
+  - ADR-110
+  - ADR-151
+  - ADR-200
+---
+
+# ADR-201: Findings assembled as classifier-ready assessment records
+
+## Context
+
+ADR-200 separated a **compliance claim** (a control-*design* assertion; SOC 2 Type I;
+OSCAL Component Definition) from an **assessment finding** (session-derived evidence
+with a determination of *satisfied* / *other than satisfied*; NIST SP 800-53A; OSCAL
+Assessment Results). It also required, in §1, that every claim carry a **determination
+criterion** — the observable session behavior that would let an assessor decide the
+finding. ADR-151 packages the tool that produces findings: `ways-audit`.
+
+ADR-200 left one thing unresolved: **who fills the determination, and when.** Two
+tempting answers are both wrong.
+
+- **The tool determines inline.** If `ways-audit` reads a claim and its firing evidence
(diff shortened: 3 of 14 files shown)


aaronsb:
## Review — assemble/findings assessability layer (ADR-201)

Reviewed against the load-bearing commitment: **assemble builds findings, never determines them.** That boundary holds in code, not just prose. Details below, ranked by severity. Nothing here is blocking.

### What I verified (the honesty boundary is real)

- **No path writes a determination.** `finding.rs:656-668` (`assemble`) constructs every `Finding` with `determination/assessed_by/assessed_at/basis = None`, and nothing downstream mutates them. `findings.rs` only *reads* `determination` (`list` at :424, `render_table` at :453) — `append_to_ledger` (:464) serializes the row as-built. The guard test at `finding.rs:704-713` asserts assembly never labels. This is the ADR's core claim and the implementation genuinely enforces it structurally.
- **Dataset uniformity is real.** `Finding` (`finding.rs:568-588`) deliberately omits `skip_serializing_if` on `criterion/determination/assessed_by/assessed_at/basis`, and `FiringEvidence` (:555-561) does the same on `first_seen/last_seen`. So every serialized row carries an identical column set with `determination: null` present, not omitted — the PR's uniform-column claim is accurate.
- **`findings_ledger` location is correct** — `paths.rs:766-768` uses `state_root()` = `\$XDG_STATE/agent-ways/findings.jsonl`, not the `~/.claude` projection. Matches ADR-201 and the doc-comment rationale.
- **`from_provenance_value` parses both control forms** — `provenance.rs:851-873` handles the structured `{id, justifications, satisfied_when}` and the legacy bare string (`- OWASP A01`). And the manifest builder preserves the raw sidecar shape (`provenance.rs:227-236` stores the full parsed YAML), so `satisfied_when` survives from sidecar → manifest → assembler. Integration path is sound.
- **`now_utc` is correct** — `util.rs:185-194`; `days_to_ymd` is unit-tested for epoch/leap/known dates (:219-225) and the time-of-day split is standard. No issue.
- **Firing aggregation is correct** — `index_firing` (`finding.rs:599-623`): `total` counts all `way_fired` (3 in the test), `sessions` dedups in encounter order (`[s1, s2]`), first/last_seen pick min/max. Tier = outcome iff `satisfied_when` present (:651-655), matching ADR-200 §3. Tests at :726-734 confirm.

### Finding 1 (medium) — ADR-201 §2 table names fields the schema doesn't have

`ADR-201...md:100-102` lists two feature fields:
- `evidence.firing` — counts, sessions, timestamps
- `evidence.transcript_refs` — session + pointer

Neither exists in the serialized record. The actual `FiringEvidence` (`finding.rs:555-561`) is flat — `total / sessions / first_seen / last_seen` — and `sessions` doubles as the transcript pointer (per the doc comment at :554). A reader mapping the ADR table to the `--json` output finds no `evidence.firing` and no `evidence.transcript_refs`. Since the ADR ships *in this PR*, it should describe the shape the code emits. Suggest aligning the table to the real fields (or noting `sessions` serves as `transcript_refs`). The *concept* is honored — this is naming drift, not a broken guarantee.

### Finding 2 (low) — re-running `assemble --write` appends duplicate rows

`append_to_ledger` (`findings.rs:464-477`) blind-appends with no dedup/upsert on `(way, control)`. Two `assemble --write` runs produce duplicate rows differing only in `assembled_at`. That's defensible as append-only *audit-trail* semantics, but it's in mild tension with the ADR's *dataset* framing — a classifier training on the ledger would see the same `(way, control)` features repeated. Worth either a dedup-on-read step, an upsert, or an explicit doc note that the ledger is append-only and consumers dedup by latest `assembled_at`.

### Finding 3 (low/nit) — timestamp ordering relies on lexicographic compare

`index_firing` first/last_seen use string `<`/`>` on `ts` (`finding.rs:614-620`). Correct for the fixed-width `...Z` UTC form `now_utc` emits (the sole writer today), but it would misorder if the firing log ever carried offset or variable-precision timestamps. A one-line comment stating the assumption would harden it.

### Nit — structured control missing `id` is silently dropped

`provenance.rs:860` (`o.get(\"id\")?`) drops a structured control with no `id` via `filter_map`. Fine given `lint` should catch malformed sidecars, but a silent drop in the assembler is worth being aware of.

### Assessment

The PR does the one thing that mattered most: it makes the assemble-never-determine boundary a property of the code (empty `Option`s, no writer, guard test), not a promise in the docs. Docs consistently present the classifier as out-of-scope and `satisfied_when` as optional (governance.md:198, provenance.md:299, ways-cli.md:344). `assemble` being a pure function with an injected timestamp is a nice testability call. The only substantive follow-up is Finding 1 — an ADR table that names fields the schema doesn't emit — plus the append/dataset-duplication consideration to think through before a classifier ever consumes the ledger.

PR #264, opened 2026-07-02
feat(settings): secret-path permissions.deny baseline (ADR-152, task #1)

aaronsb:
## Summary

Implements **ADR-152** (task #1). agent-ways already owns `permissions.allow` in the `settings.json` three-way merge; this adds the **deny-side sibling** — a curated secret-path baseline that protects the agent's file tools from credential material **by default, in every reconciled session**.

`WAYS_DENY`:
```
Read(~/.ssh/**)  Edit(~/.ssh/**)  Write(~/.ssh/**)
Read(~/.aws/**)  Read(~/.gnupg/**)  Read(~/.config/gcloud/**)
Read(~/.kube/config)  Read(~/.netrc)
Read(./.env)  Read(./.env.local)
```
`.env.example` / `.env.sample` are intentionally **not** denied.

## Design

- **Reuses the proven merge machinery** — set-union into `permissions.deny`, tracked in the `Owned` base so the three-way cleanup applies to deny exactly as to allow.
- **Secure by default, explicit opt-out** — `secret_path_deny: false` in config suppresses the baseline. Because a Claude Code deny **cannot be re-opened by a user `allow`** (deny precedes allow), the escape hatch lives in config, not settings-editing. Opting out flows through the ordinary deprecated-cleanup, restoring the user's own deny.
- **Self-audit safe** — `stripped_user_view` + `union_owned` extended to the deny slice, so the post-write invariant holds our deny and never spuriously reverts (proven by test).

## Honest scope

This gates the `Read`/`Edit`/`Write` **tools**, not the shell — `Bash(cat ~/.ssh/id_ed25519)` reads through a different door, and command-string denial is unreliable. It's a **floor against the common accident** (the agent's own file tools wandering into a secret), not exfiltration-proofing. Overselling it would be the exact overclaim ADR-200 exists to avoid.

## Docs

- config.yaml template documents the opt-out.
- **install-guide corrected**: it claimed "only the hooks block is merged into settings.json" — but the merge has long owned `permissions.allow` and now `permissions.deny` too. Ground-truth fix.

## Test plan

- 6 new deny merge tests: added-when-enabled, absent-when-opted-out, preserves-user-deny, opt-out-removes-previously-owned, idempotent, and the **self-audit invariant** holds with deny.
- config default-true / opt-out test.
- `ways-core` **85** + `ways` **197** green; `doc lint` 0/0; new code clippy-clean.
- The two commits keep ADR-first history (ADR-152 Accepted → implementation).

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -72,6 +72,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-149](./system/ADR-149-operator-config-interview-skill.md) | operator config interview skill | Accepted |
 | [ADR-150](./system/ADR-150-version-truth-and-downgrade-safe-self-update.md) | Version-truth and downgrade-safe self-update | Accepted |
 | [ADR-151](./system/ADR-151-extract-ways-core-crate-and-ways-audit-sibling-binary.md) | Extract ways-core crate and ways-audit sibling binary | Accepted |
+| [ADR-152](./system/ADR-152-framework-default-secret-path-deny-baseline.md) | Framework-default secret-path deny baseline | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-152-framework-default-secret-path-deny-baseline.md
+++ b/docs/architecture/system/ADR-152-framework-default-secret-path-deny-baseline.md
@@ -1,0 +1,127 @@
+---
+status: Accepted
+date: 2026-07-02
+deciders:
+  - aaronsb
+  - claude
+related:
+  - ADR-142
+  - ADR-147
+supersedes: []
+---
+
+# ADR-152: Framework-default secret-path deny baseline
+
+## Context
+
+agent-ways already owns a slice of Claude Code's `settings.json`. During
+`ways reconcile` the settings three-way merge (ADR-142, ADR-147) set-unions a
+fixed list of permission strings — `WAYS_PERMS` — into `permissions.allow`
+(`Bash(ways:*)`, `Edit(~/.claude/**)`, …) and tracks them in a base so entries it
+stops owning are cleaned up. The user's own `permissions` entries are held
+invariant beside ours.
+
+`permissions.deny` is currently empty. Nothing stops the agent's **file tools**
+(`Read`, `Edit`, `Write`) from reaching credential material — `~/.ssh` private
+keys, `~/.aws/credentials`, GPG keyrings, a project `.env`. A single mistaken
+tool call can read a private key into the transcript, where it is then persisted
+and, if the transcript is ever shared, leaked. The framework projects itself into
+every session; it should also project a floor of protection for the paths that
+are almost never a legitimate read target for an agent.
+
+Claude Code's permission model makes this expressible: `permissions.deny` is a
+list of `Tool(pattern)` rules (`Read(./.env)`, `Read(~/.ssh/**)`,
+`Read(//abs/path/**)`), and **deny takes precedence over allow** — a deny is a
+hard block, not a prompt. That precedence is the whole point (a broad
+`Read(~/**)` allow cannot re-open a denied secret path) and also the constraint
+this ADR must respect: a forced deny the user cannot override via `allow` needs a
+deliberate, documented escape hatch, or it becomes a cage.
+
+## Decision
+
+### 1. Ship a curated secret-path deny baseline, reconciler-owned
+
+Add a `WAYS_DENY` list, the deny-side sibling of `WAYS_PERMS`, set-unioned into
+`permissions.deny` by the same merge and tracked in the same base. The baseline
+is intentionally **narrow and high-confidence** — paths an agent has essentially
+no legitimate reason to read or write:
+
+```
+Read(~/.ssh/**)      Edit(~/.ssh/**)      Write(~/.ssh/**)
+Read(~/.aws/**)      Read(~/.gnupg/**)    Read(~/.config/gcloud/**)
+Read(~/.kube/config) Read(~/.netrc)
+Read(./.env)         Read(./.env.local)
+```
+
+`~/.ssh` denies read **and** write (the agent should neither exfiltrate a key nor
+tamper with `authorized_keys`/`config`). The credential stores deny read.
+Project env files deny the secret ones (`.env`, `.env.local`) but **not**
+`.env.example` / `.env.sample`, which are meant to be read. The list is a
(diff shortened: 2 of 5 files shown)


aaronsb:
## Code Review — ADR-152 secret-path `permissions.deny` baseline

Reviewed for the highest-risk property first (self-audit revert/strip), then opt-out, base-seeding, pattern validity, scope honesty, and idempotency. The merge machinery reuse is clean and the test set is genuinely good — six deny tests including the self-audit invariant, plus the config default/opt-out. One real correctness gap, then some minor notes and what checks out.

---

### Finding 1 — HIGH (correctness / silent user-data loss): first-reconcile base seed records `WAYS_DENY` unconditionally, defeating the opt-out and the self-audit

**Location**: `tools/ways-cli/src/cmd/settings_merge.rs`, `apply_to_files` first-reconcile seed (the new `deny: WAYS_DENY.iter().map(...).collect()` in the `else` arm, ~L446).

**Problem**: When there is no base file (fresh install *or* pre-1.0 migration), the base is seeded with `deny: WAYS_DENY` **regardless of `secret_path_deny`**. But unlike `perms`, agent-ways has *never* written `permissions.deny` before this PR — so there is no legacy deny ownership to migrate. Seeding ownership of those strings means: if a user *independently* had one of the `WAYS_DENY` patterns in their own `permissions.deny`, the seed claims agent-ways owns it.

The concrete harm is in the **opt-out + first-reconcile** path — exactly the honest-escape-hatch flow the ADR emphasizes. Trace with `secret_path_deny: false`, no base file, user's live deny = `[\"Read(~/.ssh/**)\", \"Bash(rm:*)\"]`:

- seed → `base.deny = WAYS_DENY` (includes `Read(~/.ssh/**)`)
- `merge(..., deny_secrets=false)`: `ours_deny = []`; `deprecated_deny = base.deny.filter(!ours) = WAYS_DENY`; so the user's own `Read(~/.ssh/**)` is filtered **out** of `new_deny`. Result deny = `[\"Bash(rm:*)\"]`; `merged.base.deny = []`.
- Self-audit: `user_before = stripped_user_view(live, union_owned(base, merged.base))` strips `union(WAYS_DENY, []) = WAYS_DENY` from live → `[\"Bash(rm:*)\"]`. `user_after = stripped_user_view(after, merged.base=[])` → `[\"Bash(rm:*)\"]`. **They match — the audit passes** while the user's own `Read(~/.ssh/**)` deny was silently deleted.

**Why it matters**: This is the one thing the PR's own thesis promises can't happen — a user's own entry stripped, uncaught by the self-audit. It's a security-relevant setting (a deny the user added), removed silently, on the opt-out path. Narrow trigger (first reconcile + opt-out + a personally-authored entry that textually matches a baseline pattern), but it's real data loss and the audit is blind to it precisely *because* the seed pre-declares false ownership.

**Suggestion**: Seed `deny` empty — there is no legacy deny ownership to migrate, so nothing is lost by not seeding, and the enabled path is unchanged (I traced both: enabled first-reconcile produces identical output whether the seed is `WAYS_DENY` or `[]`, because `ours_deny` re-adds the baseline and folds any user duplicate into ours):

```rust
// in the no-base-file seed:
deny: Vec::new(),   // agent-ways never previously owned deny; nothing to migrate-as-ours
```

A conditional (`if secret_path_deny { WAYS_DENY } else { [] }`) also closes the opt-out hole, but `[]` is simpler and strictly correct given there's no prior deny ownership. Worth a regression test: opt-out + no base file + a user deny entry matching a baseline pattern must survive.

---

### Finding 2 — MINOR (scope accuracy of the patterns, not a bug)

**Location**: `WAYS_DENY` (~L48) and ADR-152 Decision §1.

Two intentional-but-worth-documenting narrownesses:

1. `Read(./.env)` / `Read(./.env.local)` are root-anchored (the `./` prefix). A secret env in a subdirectory (`service/.env`) is **not** covered. Fine as a conservative floor, but the ADR's \"a project `.env`\" phrasing reads slightly broader than the pattern delivers.
2. `.env` / `.env.local` deny **Read only**, not `Edit`/`Write` — unlike `~/.ssh` which denies all three. So the agent's file tools can still *overwrite/corrupt* a project `.env`. Given the ADR's framing is exfiltration/secret-read, this is defensible, but the asymmetry with `~/.ssh` isn't called out.

Neither is an over-match: I confirmed `Read(./.env)` is an exact path and does **not** match `.env.example` / `.env.sample` (gitignore-style exact vs prefix) — the ADR's stated intent holds. These are just notes on where the floor stops; consider a one-line mention in the ADR Consequences so scope stays honest.

---

### Finding 3 — TRIVIAL (cosmetic)

`apply_to_files` success message still reports only `\"hooks + {} ways permissions\"` (allow count). It doesn't mention the deny baseline was applied. Not blocking; a user opting-in/out gets no signal in the summary line.

---

### What checks out (verified, not assumed)

- **Opt-out with an existing base file** is correct and well-tested (`opt_out_removes_previously_owned_deny_keeps_user`): `ours_deny=[]` → the prior baseline flows through `deprecated_deny` and is removed, user's own deny survives. The gap in Finding 1 is *only* the no-base-file seed.
- **`union_owned` and `stripped_user_view`** were extended to `deny` symmetrically with `perms` — the loop over `[(\"allow\", perms), (\"deny\", deny)]` and the `union_vec` helper are clean refactors, and the before/after strip is symmetric for all combinations *except* the seed case above.
- **Idempotency and three-way cleanup**: re-running is stable (traced + `deny_is_idempotent`); removing an entry from `WAYS_DENY` later cleans it from users via `deprecated_deny`/base. Correct.
- **`new_deny.is_empty()` → remove the `deny` key**: avoids writing an empty `deny: []`, and the self-audit handles the presence/absence symmetrically. Good.
- **Honest scope**: the Bash-not-covered limitation is stated plainly in the ADR §3, the PR body, and the `WAYS_DENY` doc-comment — no overclaim. The install-guide correction (\"only the hooks block is merged\" → owned slices incl. allow+deny) is an accurate ground-truth fix.
- **Config plumbing**: `secret_path_deny` defaults `true`, opt-out parsed via `apply_yaml`, tested. `apply_to_files` threads `config::global().secret_path_deny` into `merge`; all test call sites updated to the new arity.

---

**Assessment**: Design is sound and the machinery reuse is the right call. Finding 1 is the one I'd want resolved before merge — it's a silent strip of a user's own security setting on the documented opt-out path, and it slips past the self-audit that the PR relies on as its safety net. The fix is small (seed `deny` empty) and the enabled path is provably unaffected. Findings 2–3 are documentation/polish.

*Reviewer-only; I did not modify code. The pre-existing `compile.rs:334` clippy warning is out of scope here.*

PR #265, opened 2026-07-02
feat(dist): download-first attend/attend-chat pre-built binaries (task #4)

aaronsb:
## Summary

`attend` already publishes per-platform binaries (`build-attend.yml`, `attend-v*` releases), but nothing fetched them — `ways update` built them (or skipped without a toolchain). `attend-chat` had no CI at all. This wires both to the same **download-first** path `ways` and way-embed use.

## Changes

- **`tools/attend/download-attend.sh`, `tools/attend-chat/download-attend-chat.sh`** — mirror `download-ways.sh`: platform detect → GitHub Release download → checksum verify → build-from-source fallback.
- **`.github/workflows/build-attend-chat.yml`** — mirrors `build-attend.yml` so `attend-chat-v*` tags publish 4-platform binaries + checksums + a Release (`attend-chat` is already a `release.sh` component).
- **Makefile** `attend`/`attend-chat` targets — download-first, then build-fallback (matching `ways`); liveness probe switched to `--version`.
- **`ways update`** — attend/attend-chat refresh is no longer toolchain-gated; their `make` targets download the pre-built binary, so `ways update` refreshes them **without cargo** (only the build fallback needs a toolchain).

## Test plan

- Both download scripts pass `bash -n`; key substitutions verified (`BIN_NAME`, tag-prefix grep, output paths).
- `build-attend-chat.yml` is valid YAML (`-p attend-chat`, `attend-chat-v*` tag gate, correct artifact names).
- `ways` builds clean.

## Follow-up (release)

`attend-chat`'s first release (`attend-chat-v*`) still needs cutting for its binaries to exist; until then its download path falls back to build. `attend` binaries already exist (`attend-v0.7.1`), so its download path works today.

### DIFF
--- a/.github/workflows/build-attend-chat.yml
+++ b/.github/workflows/build-attend-chat.yml
@@ -1,0 +1,137 @@
+name: Build attend-chat
+
+on:
+  push:
+    branches: [main]
+    tags: ['attend-chat-v*']
+    paths:
+      - 'tools/attend-chat/**'
+      - 'tools/agent-fmt/**'
+      - 'tools/sensor-*/**'
+      - 'tools/Cargo.toml'
+      - 'tools/Cargo.lock'
+      - '.github/workflows/build-attend-chat.yml'
+  pull_request:
+    paths:
+      - 'tools/attend-chat/**'
+      - 'tools/agent-fmt/**'
+      - 'tools/sensor-*/**'
+      - 'tools/Cargo.toml'
+      - 'tools/Cargo.lock'
+      - '.github/workflows/build-attend-chat.yml'
+  workflow_dispatch:
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            platform: linux-x86_64
+            os: ubuntu-latest
+          - target: aarch64-unknown-linux-gnu
+            platform: linux-aarch64
+            os: ubuntu-latest
+          - target: x86_64-apple-darwin
+            platform: darwin-x86_64
+            os: macos-latest
+          - target: aarch64-apple-darwin
+            platform: darwin-arm64
+            os: macos-latest
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Install cargo-zigbuild (Linux cross-compile)
+        if: runner.os == 'Linux' && matrix.target == 'aarch64-unknown-linux-gnu'
+        run: |
+          pip3 install ziglang
+          cargo install cargo-zigbuild
+
+      - name: Build (native)
+        if: "!(runner.os == 'Linux' && matrix.target == 'aarch64-unknown-linux-gnu')"
+        run: cargo build --release --manifest-path tools/Cargo.toml -p attend-chat --target ${{ matrix.target }}
+
+      - name: Build (zigbuild cross-compile)
+        if: runner.os == 'Linux' && matrix.target == 'aarch64-unknown-linux-gnu'
+        run: cargo zigbuild --release --manifest-path tools/Cargo.toml -p attend-chat --target ${{ matrix.target }}
+
+      - name: Package binary
+        run: |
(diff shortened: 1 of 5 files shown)


aaronsb:
## Review — download-first attend/attend-chat (task #4)

Reviewed the two new download scripts against the `download-ways.sh` template line-by-line, `build-attend-chat.yml` against `build-attend.yml`, the Makefile targets, and the `ways update` toolchain-gate removal. **No blockers** — nothing downloads the wrong binary, mis-tags a release, or breaks the build. All ways→attend / ways→attend-chat substitutions are correct. Notes below are polish only.

### What this accomplishes
Wires both `attend` and `attend-chat` to the same download-first → build-fallback path `ways` already uses, and drops the toolchain gate in `ways update` so refresh works without cargo (for components that have a release).

### Verified correct (the collision-risk items)
- **Release-tag grep** — `download-attend.sh` uses `grep '^attend-v'`, which does **not** match `attend-chat-v*` (anchored `^`, byte 8 is `c` vs `v`); `download-attend-chat.sh` uses `grep '^attend-chat-v'`. Neither cross-matches the other. This was the highest-risk substitution and it's right.
- **Asset existence check** — `grep -q \"^\${BIN_NAME}\$\"` is anchored both ends, so `attend-linux-x86_64` won't match `attend-chat-linux-x86_64`.
- **BIN_NAME / OUTPUT_FILE / env var** — `attend-\${PLATFORM}` → `bin/attend` / `ATTEND_RELEASE`, and `attend-chat-\${PLATFORM}` → `bin/attend-chat` / `ATTEND_CHAT_RELEASE`. All correct.
- **REPO_ROOT depth** — `\$SCRIPT_DIR/../..` from `tools/attend/` and `tools/attend-chat/` both resolve to repo root (same two-up as `tools/ways-cli/`).
- **Build-fallback target** — `make attend` / `make attend-chat` respectively.
- **`build-attend-chat.yml`** — `-p attend-chat` (build/zigbuild/test), `tags: ['attend-chat-v*']`, release gate `refs/tags/attend-chat-v`, artifacts `attend-chat-<platform>`, `sha256sum attend-chat-*`, title `attend-chat \${TAG#attend-chat-}`. No non-chat `attend` leftover.
- **Makefile** — download-first → cargo → error ordering matches the `ways` target; `--version` liveness probe; if/elif chain, no loop; failure paths clean up and fall through. Correct.
- **`ways update`** — dry-run print and execution block both updated consistently; `has_toolchain` remains used by the `refresh_ways` downgrade-guard / `guard_action` path, so it is **not** now-dead.

### Minor (non-blocking)

**1. Leftover `download-ways.sh` in the header comment — both new scripts**
`tools/attend/download-attend.sh:9` and `tools/attend-chat/download-attend-chat.sh:9` still read `#   download-ways.sh [--release TAG] [output-dir]` in the usage banner (and the `# Find the latest ways release` comment near line 71 is also copied verbatim). Comment-only; runtime `--help` uses `\$0` so user-facing help is correct. Worth a sweep for accuracy.

**2. `grep \"attend-\"` in the available-binaries listing — `download-attend.sh`**
The error-path listing uses unanchored `grep \"attend-\"`, which would also surface `attend-chat-*` names. Harmless in practice (attend and attend-chat ship in separate releases, so an `attend-v*` release only contains `attend-*` assets), but it's broader than the template's intent. Consider `grep \"^attend-\"` for symmetry. Purely cosmetic (informational output on the no-binary path).

**3. `ways update` doc comment slightly ahead of reality for attend-chat**
`update.rs` now states attend/attend-chat \"refresh without a toolchain; only the build fallback needs cargo.\" True for `attend` (has `attend-v0.7.1`), but **attend-chat has no release yet**, so today its download path exits non-zero and falls back to the cargo build — which does need a toolchain. The PR description discloses this honestly (\"until then its download path falls back to build\"); the tension is only that the in-code comment reads as a blanket claim. A one-line caveat would keep the code comment as honest as the PR body. Not a correctness issue.

### Honesty check
The PR does not overclaim: `attend` binaries exist and the download path works today; attend-chat's first `attend-chat-v*` release still needs cutting and is explicitly called out as build-fallback until then. `build-attend-chat.yml` is what makes that release possible.

**Assessment:** Sound, faithful mirror of the proven template. Ship-ready once the three cosmetic notes are considered; none block merge.

PR #267, opened 2026-07-03
feat(dist): make ways-audit a first-class suite member in install + update

aaronsb:
## Summary

Closes the gap the release exposed: `ways-audit`'s **publish** side shipped (CI + `ways-audit-v1.0.0` is live), but nothing **fetched** it — `make ways-audit` was build-only, `install`/`setup` skipped it, and `ways update` never refreshed it. A published-but-unfetchable binary, and any opted-in copy would silently go stale.

Per the principle: **`ways update` manages the entire agent-ways collection — no partial tool collections, no per-tool lifecycle.** So `ways-audit` joins the suite through the **same machinery** as `ways`/`attend`, not a bespoke path.

## Changes

- **`tools/ways-audit/download-ways-audit.sh`** — mirrors `download-ways.sh`. **Verified end-to-end against the live `ways-audit-v1.0.0` release**: platform-detects, downloads, checksum-verifies (`a8140c69…`), runs (`ways-audit 1.0.0`).
- **Makefile** — `make ways-audit` is now download-first → build-fallback (matching `ways`; produces `bin/`, `install` links it); new `ways-audit-rebuild`; **`setup` builds it, `install` links it, `update-binaries` rebuilds it** (uninstall already removed it).
- **`ways update`** — `ways-audit` refreshes through the **existing** `refresh_component` loop (`[ways-audit, attend, attend-chat]`) via its download-first `make` target. No new lifecycle manager, no toolchain gate.
- **ADR-151 §4** — records the principle: `ways-audit` is a first-class suite member, not an opt-in add-on; "deliberately-invoked" (§2) is about *usage* (you run it when you want it, like `/ship`), not *installation*. Corrects PR-B's lean-install call that stranded the binary.
- **install-guide** — suite binary list + the "what gets downloaded" table reconciled (it was stale for `attend` too, from #265).

## Design note

This is deliberately **not** a separate suite lifecycle manager — every binary flows through the one existing update path (`refresh_component` → its download-first `make` target). Adding `ways-audit` is one more entry in the same list, in the same places `attend` already appears.

## Test plan

- Download script: fetches + checksum-verifies + runs the real `ways-audit-v1.0.0` binary.
- `ways` builds; `make -n ways-audit` parses; `setup` depends on `ways-audit`.
- `adr lint` / `doc lint` 0/0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 # Update:        make update
 
 .DEFAULT_GOAL := help
-.PHONY: setup install uninstall update update-binaries sync-to-home sync-to-home-link sync-to-home-test clean help deps ways ways-rebuild ways-audit attend attend-rebuild attend-chat attend-chat-rebuild hooks-install way-embed-rebuild lint test test-unit test-sim test-lang test-locales test-multilingual release purge-attend-state
+.PHONY: setup install uninstall update update-binaries sync-to-home sync-to-home-link sync-to-home-test clean help deps ways ways-rebuild ways-audit ways-audit-rebuild attend attend-rebuild attend-chat attend-chat-rebuild hooks-install way-embed-rebuild lint test test-unit test-sim test-lang test-locales test-multilingual release purge-attend-state
 
 ifeq ($(OS),Windows_NT)
     SHELL := C:/Program Files/Git/usr/bin/bash.exe
@@ -45,7 +45,7 @@ help:
 	@echo "  make sync-to-home-link  [legacy] symlink variant of sync-to-home"
 	@echo "  make ways         Get ways binary (download or build from source)"
 	@echo "  make ways-rebuild Force rebuild ways from source"
-	@echo "  make ways-audit   Build + link the optional compliance binary (ADR-151)"
+	@echo "  make ways-audit   Get ways-audit compliance binary (download or build)"
 	@echo "  make attend       Build attend binary"
 	@echo "  make attend-rebuild Force rebuild attend from source"
 	@echo "  make lint         Run clippy on Rust workspace (warnings = errors)"
@@ -95,7 +95,7 @@ deps:
 	@echo "Toolchain ready. Now run: make setup"
 
 # Build ways CLI + set up embedding engine + generate initial corpus.
-setup: ways attend attend-chat
+setup: ways ways-audit attend attend-chat
 	@echo "Setting up embedding engine..."
 	@# Optional accelerator — if its build deps are missing, warn and continue so
 	@# the rest of install (incl. the PATH symlinks) still completes. Without it,
@@ -117,6 +117,7 @@ setup: ways attend attend-chat
 install: hooks-executable setup hooks-install
 	@mkdir -p "$(XDG_BIN)"
 	@$(LINK) "$(CURDIR)/$(WAYS_BIN)" "$(XDG_BIN)/ways"
+	@$(LINK) "$(CURDIR)/$(WAYS_AUDIT_BIN)" "$(XDG_BIN)/ways-audit"
 	@$(LINK) "$(CURDIR)/$(ATTEND_BIN)" "$(XDG_BIN)/attend"
 	@$(LINK) "$(CURDIR)/$(ATTEND_CHAT_BIN)" "$(XDG_BIN)/attend-chat"
 	@mkdir -p "$(CLAUDE_BIN)"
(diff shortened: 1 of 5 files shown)


aaronsb:
## Review: ways-audit as first-class suite member

Reviewed against the stated principle (one existing mechanism, no parallel lifecycle, slot in exactly like `attend`). This holds up well. No blocking issues; the wiring is complete and the highest-risk item (tag cross-contamination) is provably clean. Findings ranked by severity.

### PASS (headline) — no tag cross-contamination
Verified empirically against the live release list (`ways-v1.0.3` and `ways-audit-v1.0.0` both exist right now):
- `download-ways.sh:73` `grep '^ways-v'` -> matches **only** `ways-v1.0.3`. `ways-audit-v1.0.0` begins `ways-a...`, so the `^ways-v` anchor cannot match it.
- `download-ways-audit.sh:63` `grep '^ways-audit-v'` -> matches **only** `ways-audit-v1.0.0`, never `ways-v*`.

Neither script can select the other's release tags in either direction. This was the load-bearing check and it is correct.

### PASS — Makefile lifecycle wiring is consistent everywhere a sibling appears
`ways-audit` / `ways-audit-rebuild` now appear in every place `ways`/`ways-rebuild` do:
- `.PHONY` (Makefile:9) — both targets ✓
- `setup:` prereqs (Makefile:98) ✓
- `install` link block (Makefile:120) ✓
- `install` echo list (Makefile:135) ✓
- `update-binaries:` (Makefile:184) — `ways-audit-rebuild` ✓
- `uninstall:` (Makefile:154-155) — pre-existing ✓
- `help` (Makefile:48) — `ways-audit` main target ✓

Restructured `ways-audit` target (Makefile:223-239) correctly mirrors `ways` (Makefile:188-204): existing-check -> download (`download-ways-audit.sh`) -> cargo build-fallback, produces `bin/ways-audit`, and does **not** self-link to XDG — `install` does the linking (Makefile:120). The old target's XDG self-link and its `mkdir -p bin "$(XDG_BIN)"` are correctly gone. New `ways-audit-rebuild` (Makefile:242-251) matches `ways-rebuild` exactly (cargo guard -> check-rust -> build -> link bin only).

### PASS — download-ways-audit.sh diff vs download-ways.sh
All substitutions complete and correct: `BIN_NAME=ways-audit-${PLATFORM}` (:23), `OUTPUT_FILE=.../ways-audit` (:49), tag grep (:63), platform-asset existence check `grep -q "^${BIN_NAME}$"` (:80, exact-anchored), checksum grep (:107), `make ways-audit` fallback strings. No leftover `ways`-that-fetches-the-wrong-binary. Note the available-binaries listing at :86 uses `grep "^ways-audit-"` (anchored) — tighter than download-ways.sh's unanchored `grep "ways-"`; a small improvement, harmless.

### PASS — `ways update` uses the existing mechanism, no new manager
`refresh_component` loop extended to `["ways-audit", "attend", "attend-chat"]` (update.rs:104). `refresh_component` (update.rs:146-183) resolves `bin/ways-audit`, renames it aside to defeat the download script's own already-installed early-return, shells `make ways-audit` (download-first), and reverts on failure. Same path as `attend`, no per-tool special-casing, no parallel lifecycle. Dry-run line (update.rs:66) and module doc (update.rs:12-18, 105-110) updated consistently.

### PASS — ADR-151 §4 amendment is accurate, no overclaim
The added paragraph claims `make install`/`make setup` build+link `ways-audit` (true: setup:98, install:120) and that `ways update` refreshes it via the same `refresh_component` path (true: update.rs:104). The reframing of "deliberately-invoked" as usage-not-install is coherent with the code. No claim exceeds what the diff implements.

---

### Low (informational, not blocking) — silent-downgrade window from omitting the guard
`refresh_component` has no downgrade guard; only `refresh_ways` does (update.rs:202). This is **consistent with `attend`** and matches the design intent, so it is acceptable. Flagging the behavioral consequence so it's a conscious acceptance rather than an oversight: source HEAD for `ways-audit` may currently be ahead of `ways-audit-v1.0.0`. If someone has a from-source `ways-audit` newer than the latest release, `ways update` will rename it aside, download v1.0.0, see `--version` succeed, and keep it — a silent downgrade. `attend` carries the identical property today. If uniformity with `attend` is the deliberate call (it appears to be), no change needed; worth a one-line note in the ADR only if you want the asymmetry-with-`ways` documented.

### Trivial (optional) — `ways-audit-rebuild` absent from `help`
`help` lists `ways-rebuild` and `attend-rebuild` but not `ways-audit-rebuild`. This is consistent with `attend-chat-rebuild`, which is also unlisted, so `help` is already selective — no action required unless you want every `*-rebuild` enumerated.

### Assessment
Solid, faithful to the "one mechanism" principle. The change adds `ways-audit` as one more entry in the same lists `attend` already occupies, the download script is a clean mirror with correctly-anchored tag selection, and the update path reuses `refresh_component` with no bespoke logic. The only substantive thing to consciously sign off on is the deliberate absence of the downgrade guard (matching `attend`). Recommend merge once that tradeoff is acknowledged.

---
*AI-assisted review via Claude Code (code-reviewer). Reviewer only — no files modified.*

PR #269, opened 2026-07-03
fix(dist): ways update links newly-added suite binaries onto PATH

aaronsb:
## The bug

After `ways update`, `ways-audit` was **not on PATH**. Root cause: `$XDG_BIN/<bin>` PATH symlinks were created **only by `make install`**. `ways update` refreshes `bin/` but relied on those symlinks pre-existing — true for `ways`/`attend` (linked at the original install), but **not** for a binary *added to the suite later*. So `ways-audit` (once a 1.0.4 update builds it) lands in `bin/` with no PATH link; `reconcile` only handles the `~/.claude` projection, not `$XDG_BIN`. My #267 removal of the `make ways-audit` self-link sealed the gap.

## The fix

A single idempotent **`make relink`** target that symlinks every *present* suite binary onto PATH:
- `ways`, `ways-audit`, `attend`, `attend-chat` → `$XDG_BIN`
- `way-embed` → `$CLAUDE_BIN`

Existence-guarded (safe before all binaries are built) and idempotent (`ln -sf`). **`install`** now uses it (DRY), and **`ways update`** runs it after the refresh loop — so a binary *newly added to the suite* self-heals its PATH link on the next update, no fresh `make install` needed.

This is the **general** fix, not a ways-audit-specific patch: the next binary added to the suite is handled the same way, through the one mechanism.

## Test plan

- `make relink` against a scratch `$XDG_BIN`: links only existing binaries (`ways`/`ways-audit` present → linked; `attend*` absent → skipped), idempotent on re-run, does not touch the real `~/.local/bin`.
- `ways` builds; `install`/`relink` targets parse.
- Confirmed the live repro on a real install and manually unstuck it (built + linked `ways-audit`; `ways-audit 1.0.0` now resolves on PATH).

## One-release lag

The fix lives in the `ways` binary (`update.rs`), so it's live from the next release; an older install self-heals once updated to it. `make install` is the manual recovery meanwhile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 # Update:        make update
 
 .DEFAULT_GOAL := help
-.PHONY: setup install uninstall update update-binaries sync-to-home sync-to-home-link sync-to-home-test clean help deps ways ways-rebuild ways-audit ways-audit-rebuild attend attend-rebuild attend-chat attend-chat-rebuild hooks-install way-embed-rebuild lint test test-unit test-sim test-lang test-locales test-multilingual release purge-attend-state
+.PHONY: setup install relink uninstall update update-binaries sync-to-home sync-to-home-link sync-to-home-test clean help deps ways ways-rebuild ways-audit ways-audit-rebuild attend attend-rebuild attend-chat attend-chat-rebuild hooks-install way-embed-rebuild lint test test-unit test-sim test-lang test-locales test-multilingual release purge-attend-state
 
 ifeq ($(OS),Windows_NT)
     SHELL := C:/Program Files/Git/usr/bin/bash.exe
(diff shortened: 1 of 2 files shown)


aaronsb:
## Review: relink target + ways-update integration

Reviewed the `install` refactor, the new `relink` target, and the `update.rs` wiring against the four focus areas. **No blocking issues** — the fix links what it claims, preserves all prior `install` linking behavior, and handles failure non-fatally. Details below, ranked by severity.

### Assessment: behavior preservation (the core question)

`relink` (Makefile:122-133) links exactly the same set the old inline `install` did:
- The 4 XDG binaries — `ways`, `ways-audit`, `attend`, `attend-chat` → `$(XDG_BIN)` (loop, Makefile:124-128).
- `way-embed` → `$(CLAUDE_BIN)` with the `-ef` same-file guard preserved (Makefile:131). When the repo IS `~/.claude`, `$(CURDIR)/bin/way-embed -ef $(CLAUDE_BIN)/way-embed` is true → `:` no-op, so no "same file" error. Correct.

No binary that `install` used to link is now missed. During `make install`, `setup` (Makefile:98) builds the 4 XDG binaries before `relink` runs, so every existence guard is satisfied — no regression from the guards.

Windows names check out: the `*_BIN` vars are `bin/<name>` with **no** `$(EXE)` (the build copies `tools/target/release/<name>$(EXE)` into `bin/<name>`, Makefile:26-30, 205). `relink`'s `bin/$$b` and `$(WAY_EMBED_BIN)` match those names on both platforms — no `.exe` mismatch inside the Makefile.

Integration order is correct: refresh loop (update.rs:113-117) → `make relink` (update.rs:124-130) → corpus + reconcile (update.rs:138-143). Binaries are refreshed, then linked, then projected. The relink call is non-fatal (`if let Err(e)` → warn, no propagation, update.rs:125-130) and runs in `&app`, which `is_app_checkout` (update.rs:50) has already validated as the checkout holding the Makefile. `.PHONY` includes `relink` (Makefile:9) and the dry-run is renumbered consistently (update.rs:67-68).

### Finding 1 — Low: `ways update` relink doesn't surface the "XDG_BIN not on PATH" note
**Location**: update.rs:124-130 vs Makefile:145-151

`install` prints a note when `$(XDG_BIN)` isn't on `$PATH`. The `relink` step in `ways update` links correctly but prints nothing, so a user whose `$XDG_BIN` isn't on PATH gets a silently-linked-but-unfound binary. In practice this is low impact — the user ran `ways`, so `$XDG_BIN` (where `install` linked `ways`) is almost certainly already on PATH. Optional: have `update` echo the same PATH hint, or leave as-is since it mirrors the pre-existing assumption.

### Finding 2 — Low / informational: way-embed link is now existence-guarded (strict improvement)
**Location**: Makefile:129-133

Old `install` linked `way-embed` unconditionally and always `mkdir -p $(CLAUDE_BIN)`. Since `way-embed` is optional (Makefile:103 tolerates a failed build), the old path would create a **dangling** `$(CLAUDE_BIN)/way-embed` symlink when the engine didn't build. `relink` now guards on `[ -e bin/way-embed ]` and only makes the dir when it links — strictly safer, no dangling link. Not a regression; just confirming the changed behavior is intended.

### Finding 3 — Nit: mixed literal vs. variable bin names
**Location**: Makefile:124-126

The XDG loop hardcodes literal names (`for b in ways ways-audit attend attend-chat` / `bin/$$b`) while `way-embed` uses `$(WAY_EMBED_BIN)`. Equivalent today since every var is `bin/<name>`, but a future rename of one of the `*_BIN` vars wouldn't propagate into the loop. Minor consistency point, not a defect.

### Note (pre-existing, out of scope)
`update.rs` names bin files via `exe()` → `bin/ways.exe` on Windows (update.rs:59, 161), while the Makefile uses `bin/ways` (no `.exe`). This mismatch predates this PR and the PR notes Windows update fails early anyway; `relink` itself is internally consistent with the Makefile. Flagging only for awareness.

### Edge cases checked
- **Bare install / no Makefile**: `is_app_checkout` (update.rs:50) bails earlier with a clear message, so `make relink` never runs without a Makefile — no spurious per-update warning.
- **Re-pointing a correct symlink**: `ln -sf` always points `$XDG_BIN/<b>` → `$(CURDIR)/bin/<b>`, i.e. the app dir make runs in. Same target `install` would set; running update in a different clone repoints there, identical to running `install` there. Expected, not wrong.
- **Failed component refresh**: if a binary can't be acquired, `bin/<b>` is absent and `relink` skips it (correct — nothing to link), with the refresh failure already warned upstream.

Solid, well-scoped fix with matching comments and a clear test plan. The general "self-heal any newly-added suite binary" framing holds — the mechanism isn't ways-audit-specific.

PR #271, opened 2026-07-03
docs(adr): draft session-introspection design (ADR-153/154 + plan)

aaronsb:
## Summary

**Design drafts — not for immediate merge.** Research-grounded ADRs + implementation plan for the session-introspection tooling (rethink fixes, `ways think` live monitor, non-interactive dump, why-fired drill-down). Status **Draft**; to review/refine next session before implementation.

- **ADR-153 (Draft)** — the introspection *substrate*: one typed model joining events log + transcripts + frontmatter. Captures the correctness prerequisite (the events-log split-brain that causes "rethink misses post-1.0.0 sessions") and the **honest join grain** (fired-way↔criteria solid; fired-way↔turn heuristic until fire-time enrichment; semantic stays way-level cosine, never a fabricated term). Shares the transcript-evidence substrate with ADR-201 findings.
- **ADR-154 (Draft)** — three front-ends over one model: keep crossterm + a ~300-line micro-compositor (not ratatui — with a documented escape hatch), poll+stat-gate refresh (not notify), project-scoping fix + `--list --json`.
- **Design note** — 6 independently-shippable increments (correctness fixes first), guardrails, open questions.

## Provenance

Grounded in a 3-probe research pass (rethink internals, correlation data model, TUI trade study) with `file:line` citations throughout. `adr lint` / `doc lint` clean.

## Next

Review the drafts → settle the open questions (command naming, ways-core placement, the ratatui escape-hatch criterion) → implement increment 1 (the rethink correctness fixes, independently shippable).

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -73,6 +73,8 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-150](./system/ADR-150-version-truth-and-downgrade-safe-self-update.md) | Version-truth and downgrade-safe self-update | Accepted |
 | [ADR-151](./system/ADR-151-extract-ways-core-crate-and-ways-audit-sibling-binary.md) | Extract ways-core crate and ways-audit sibling binary | Accepted |
 | [ADR-152](./system/ADR-152-framework-default-secret-path-deny-baseline.md) | Framework-default secret-path deny baseline | Accepted |
+| [ADR-153](./system/ADR-153-session-introspection-substrate-correlating-fired-ways-to-turns.md) | Session-introspection substrate — correlating fired ways to turns | Draft |
+| [ADR-154](./system/ADR-154-rethink-think-and-non-interactive-introspection-one-model-three-front-ends.md) | rethink, think, and non-interactive introspection — one model, three front-ends | Draft |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -73,8 +73,8 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-150](./system/ADR-150-version-truth-and-downgrade-safe-self-update.md) | Version-truth and downgrade-safe self-update | Accepted |
 | [ADR-151](./system/ADR-151-extract-ways-core-crate-and-ways-audit-sibling-binary.md) | Extract ways-core crate and ways-audit sibling binary | Accepted |
 | [ADR-152](./system/ADR-152-framework-default-secret-path-deny-baseline.md) | Framework-default secret-path deny baseline | Accepted |
-| [ADR-153](./system/ADR-153-session-introspection-substrate-correlating-fired-ways-to-turns.md) | Session-introspection substrate — correlating fired ways to turns | Draft |
-| [ADR-154](./system/ADR-154-rethink-think-and-non-interactive-introspection-one-model-three-front-ends.md) | rethink, think, and non-interactive introspection — one model, three front-ends | Draft |
+| [ADR-153](./system/ADR-153-session-introspection-substrate-correlating-fired-ways-to-turns.md) | Session-introspection substrate — correlating fired ways to turns | Accepted |
+| [ADR-154](./system/ADR-154-rethink-think-and-non-interactive-introspection-one-model-three-front-ends.md) | `ways introspect` — one model, three front-ends | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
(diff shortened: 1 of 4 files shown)


aaronsb:
## Design review — ground-truth verification (ADR-153/154 + plan)

Reviewed as a design-docs PR: the priority was verifying the specific, checkable claims the ADRs make about the *existing* codebase, since a design built on a wrong claim is the real risk. **Every major factual claim checks out against the source.** Findings below are minor internal-consistency nits only — no confabulated references, no factual errors, no substantive overclaims.

### Ground-truth claims — all VERIFIED

1. **Split-brain events log (ADR-153 §Context/§1):** Confirmed exactly as described.
   - `hooks/ways/clear-markers.sh:28` (session_start) and `hooks/ways/inject-subagent.sh:151` (subagent way_fired) both hardcode `${HOME}/.claude/stats/events.jsonl`.
   - `paths::events_log()` (`tools/ways-core/src/paths.rs:220`, via `resolve_events` 234-244) prefers `$XDG_STATE/agent-ways/events.jsonl`, falling back to legacy only if the preferred file is absent.
   - The Rust writer `session::log_event` (`tools/ways-cli/src/session.rs:429`) *also* routes through `events_log()`, so post-migration the shell-written session_start lands in legacy while Rust readers (`rethink.rs:115`) read XDG — the split is real and precisely characterized.
   - `ways response-topics-path` is a genuine registered subcommand (`main.rs:300`/`770`, consulted by `check-prompt.sh:29` / `check-response.sh:25`) — the cited precedent is accurate.

2. **Fire sites + no uuid / no matched term (ADR-153 §3, plan incr. 3):** Confirmed.
   - `cmd/show/mod.rs:152-186` and `cmd/scan/state.rs:100-108` emit way_fired; `cmd/scan/mod.rs:108` is the `PromptMatch::Fired` arm dispatching into show. Line refs are accurate.
   - The way_fired field set (show/mod.rs:153-184) carries no transcript `uuid` and no matched term/span; semantic fires carry only way-level `fire_score`. The "why is unanswerable from today's log" premise holds.

3. **Rethink scoping bug + no `--list --json` (ADR-154 §Context, plan incr. 1):** Confirmed.
   - `rethink.rs:124-131` sets `project_scope = None` when detection fails; `list_sessions` at `665-668` filters only `if let Some(pf)` and uses substring `project.contains(pf)` — so a `None` scope silently globalizes and the compare is a loose substring. `rethink_dump.rs:94-102` mirrors this.
   - `main.rs:669-674`: `--json` dispatches to `rethink_dump::run_json` (single-session dump) and ignores `list`; the `list` path (`list_sessions`) is text-only. There is genuinely no machine-listable enumeration.

4. **TUI trade study (ADR-154 §2/§3):** Confirmed.
   - `ways-cli/Cargo.toml`: crossterm 0.29, no ratatui. `rethink.rs:8` imports crossterm only; rendering is hand-rolled ANSI strings.
   - `reconstruct_frames` (`rethink.rs:404`) returns `Vec<Frame>` (pure data) and feeds both the TUI (`rethink.rs:160`) and the JSON dump (`rethink_dump.rs:133`) — the "already half-exists" split is real.
   - `attend-chat/Cargo.toml`: iocraft 0.8, smol 2, async-channel 2 — exactly as claimed; ways-cli carries none of these.

### Citations — all exist and support their use
ADR-111 (unified CLI single-surface), ADR-134 (fire/near-miss telemetry), ADR-142 (XDG distribution — the migration that created the split), ADR-200 (compliance epistemic-honesty framing), ADR-201 (findings read `firing events + transcript pointers`, `ADR-201:47-48`) all exist and roughly support the claimed use. INDEX.md updated correctly (rows 76-77). No fabricated ADR/precedent found.

### Findings (minor)

**F1 — Low: stray `think` naming contradicts the decided surface.** ADR-154 decides one `ways introspect <replay|live|dump>` surface and explicitly *rejects* the `think`/`rethink` verb pair as too cute. But two spots still name the live front-end `think`:
- ADR-153:19 — \"post-hoc \`rethink\`, live \`think\`, a non-interactive dump\"
- implementation-plan:6 — \"**Motivates:** \`ways rethink\` fixes, \`ways think\`, non-interactive introspection…\"

Given the project convention that backtick-quoted strings read as literal doctrine, these read as if `think`/`ways think` were a decided command name, which ADR-154 says it is not. Worth reconciling to `ways introspect live`. (ADR-154's filename slug also contains `rethink-think`, but the doc title and body are consistent, so that's cosmetic.)

**F2 — Very low / optional: way_fired field enumeration is non-exhaustive.** ADR-153 lists way_fired as carrying `way, domain, trigger, scope, project, session, token_position` (+ semantic `fire_score`). The code (show/mod.rs:174-184) also conditionally emits `parent`, `tree_depth`, `epoch_distance`, `team`. Not an error — the load-bearing claims (no uuid, no matched term) are correct and the list isn't asserted exhaustive — noted only for completeness; it doesn't affect the design.

### Assessment
The ground truth is sound. The two correctness bugs the design is built on (split-brain log, silent globalization) are real and accurately located; the enrichment premise (no uuid/term persisted) is correct; the TUI trade study reflects the actual dependency graph; every cited ADR exists and says roughly what's claimed. Only F1 (naming reconciliation) is worth acting on before merge, and it's a doc-consistency fix, not a design flaw.

PR #272, opened 2026-07-03
fix(introspect): increment 1 — rethink correctness (events-log, scoping, --list --json)

aaronsb:
Increment 1 of the session-introspection work (ADR-153/154, PR #271). These are the correctness fixes that make `ways rethink` right today — independently shippable, no dependency on the later introspection increments.

## What was broken

1. **`rethink` missed post-1.0.0 sessions.** The shell hooks (`clear-markers.sh`, `inject-subagent.sh`) hardcoded the legacy `~/.claude/stats/events.jsonl` while the Rust writer and all readers resolve `paths::events_log()`, which prefers `$XDG_STATE` post-ADR-142. On migrated installs every `session_start` the hooks wrote was orphaned in the legacy file — invisible to rethink.
2. **`rethink` silently globalized.** When current-project detection returned `None` (manual run, no `CLAUDE_PROJECT_DIR`), it replayed across *all* projects instead of failing. It also filtered with a loose `project.contains(scope)` substring, so sibling paths (`/a/proj` vs `/a/proj-2`) bled together.
3. **No machine-listable session enumeration.** `--json` ignored `--list` and dumped a session; an agent had no way to enumerate sessions before picking one.

## Fixes

- **Single-writer events log + transitional union-read** (ADR-153 §1). New `ways events-log-path` subcommand (mirrors `sessions-root`) is the one canonical path every writer resolves; new `hooks/ways/events-log.sh` helper resolves it via the binary (pattern of `response-topics-path`) with a legacy fallback, and both hooks write `$EVENTS_LOG`. `paths::events_log_sources()` + `firing::load_events_text()`/`load_events()` union the `$XDG_STATE` and legacy logs so already-orphaned sessions surface without a backfill. The migrator *moves* `stats/` (never copies), so the files never overlap — no dedup needed.
- **Project scoping** (ADR-154 §4). Default current project; add `--all`; **fail loud** (naming the missing marker + remedies) instead of globalizing; compare normalized absolute paths for equality, not substring. Shared `resolve_project_scope` + `project_matches`.
- **`ways rethink --list --json`** enumerates in-scope sessions (newest first; `scope` + `count`) as structured data.

## Deployment / manifest

`events-log.sh` is git-tracked, so it ships via the existing `hooks/ways` tree symlink on `ways update` (no `settings.json` registration — it's a sourced helper like `sessions-root.sh`). The `events-log-path` subcommand is compiled into the binary that `ways update` refreshes; during any window where pulled hooks meet an old binary, the helper falls back to the legacy path (no breakage — verified against the currently-installed v1.0.5).

## Test plan

- New unit tests: `resolve_event_sources` union resolution; `project_matches` exact-not-substring; scope resolution (`--all`→all, explicit wins).
- `cargo test -p ways -p ways-core`: 198 + 7 pass; the 9 `session_sim` failures are pre-existing on `main` (environment-dependent), unchanged by this PR.
- `cargo clippy -p ways -p ways-core`: clean under `-D warnings`.
- Manually verified all four surfaces: scoped `--list --json` (now shows the live post-1.0.0 session), `--all` (1322 across projects), and both fail-loud paths (text + JSON error).

### DIFF
--- a/hooks/ways/clear-markers.sh
+++ b/hooks/ways/clear-markers.sh
@@ -6,6 +6,7 @@
 # Clears this session's state directory only — other sessions stay intact
 
 source "$(dirname "$0")/sessions-root.sh"
+source "$(dirname "$0")/events-log.sh"
 
 INPUT=$(cat)
 SESSION_ID=$(echo "$INPUT" | jq -r '.session_id // empty' 2>/dev/null)
@@ -18,11 +19,11 @@ else
   rm -rf "${SESSIONS_ROOT}" 2>/dev/null
 fi
 
-# Log session event
-mkdir -p "${HOME}/.claude/stats" 2>/dev/null
+# Log session event (canonical events-log path resolved via events-log.sh)
+mkdir -p "$(dirname "$EVENTS_LOG")" 2>/dev/null
 jq -nc --arg ts "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
   --arg event "session_start" \
   --arg project "${CLAUDE_PROJECT_DIR:-$PWD}" \
   --arg session "${SESSION_ID:-unknown}" \
   '{ts:$ts,event:$event,project:$project,session:$session}' \
-  >> "${HOME}/.claude/stats/events.jsonl" 2>/dev/null
+  >> "$EVENTS_LOG" 2>/dev/null
--- a/hooks/ways/events-log.sh
+++ b/hooks/ways/events-log.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+# Canonical telemetry events-log path — shared by every hook that appends events.
+#
+# MUST resolve to the same file as ways-core `paths::events_log()`, which the
+# Rust writer (`session::log_event`) and every reader (`firing::load_events`)
+# use. That path has XDG-vs-legacy fallback logic (ADR-142/ADR-153) that is
+# awkward to reproduce faithfully in bash, so instead we ask the binary via
+# `ways events-log-path` — the same delegation pattern as `ways
+# response-topics-path`. Hardcoding `~/.claude/stats/events.jsonl` here is what
+# orphaned post-migration `session_start` lines from `rethink` (ADR-153 §1).
+#
+# If the binary is unavailable we fall back to the legacy path so telemetry still
+# lands somewhere; the reader's union recovers it later.
+#
+# Usage: source this file, then use $EVENTS_LOG
+#   source "$(dirname "$0")/events-log.sh"
+
+_ways_bin="${HOME}/.claude/bin/ways"
+EVENTS_LOG=""
+if [[ -x "$_ways_bin" ]]; then
+  EVENTS_LOG=$("$_ways_bin" events-log-path 2>/dev/null)
+fi
+EVENTS_LOG="${EVENTS_LOG:-${HOME}/.claude/stats/events.jsonl}"
+unset _ways_bin
--- a/hooks/ways/inject-subagent.sh
+++ b/hooks/ways/inject-subagent.sh
@@ -15,6 +15,7 @@
 # context regardless of what the parent already triggered.
 
 source "$(dirname "$0")/sessions-root.sh"
+source "$(dirname "$0")/events-log.sh"
 
 INPUT=$(cat)
 SESSION_ID=$(echo "$INPUT" | jq -r '.session_id // empty')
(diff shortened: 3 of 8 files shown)


aaronsb:
## Review — increment 1 (events-log, scoping, --list --json)

Verified each focus area against the code on `fix/introspect-correctness`, not the description. Build/`cargo test -p ways -p ways-core` unit suites pass (200 + 7); the 9 `session_sim` failures reproduce identically on `main` (marker-creation, environment-dependent) — pre-existing and unrelated. `cargo clippy -p ways -p ways-core -- -D warnings` is clean. (Note: `--all-targets` clippy trips on `settings/compile.rs:334` `unnecessary_get_then_check` — that's pre-existing test code untouched by this PR, not a finding here.)

### Finding 1 (minor, functional): `--list --json` emits duplicate sessions + inflated `count`

**Location**: `rethink.rs` `gather_sessions` (~L703) → `rethink_dump.rs` `run_list_json`.

`gather_sessions` pushes one `SessionInfo` per `session_start` line. But `clear-markers.sh` runs on **SessionStart and after compaction** (its own header says so), so any session that compacts writes `session_start` more than once with the *same* `session` id. Result: the `sessions` array contains duplicate ids, and `count` is inflated by the number of compactions. The duplicate entries even carry identical `event_count`/`way_fires` (those are aggregated by id) but differing `ts`/`duration_secs`, which is extra-confusing as structured data.

For the stated "agent enumerates then dumps by id" flow it's tolerable — `most_recent_session` still resolves a single id — but `count` misreports and an agent iterating the array sees redundant candidates. The old text `--list` had the same dupes, but exposing it as a JSON contract is new surface. Suggest de-duping by `session` id (keep earliest `ts` as the session start) before serializing in `run_list_json`, or documenting that consumers must dedup. Cheap fix: collapse in `gather_sessions` (only push if id unseen), which also cleans up the text list.

### Consideration (not blocking): equality scoping can hide sessions in symlink / subdir edge cases

**Location**: `rethink.rs` `project_matches`/`normalize_project_path`, `resolve_project_scope`.

The substring→equality change is correct for the primary path: when `CLAUDE_PROJECT_DIR` is set at both write (`clear-markers.sh` stores `${CLAUDE_PROJECT_DIR:-$PWD}`) and read (`resolve_project_scope` reads the same env first), scope == stored exactly and normalization only needs the trailing-slash trim it does. Sibling bleed (`/a/proj` vs `/a/proj-2`) is genuinely fixed.

The residual edge: on a **manual** run with `CLAUDE_PROJECT_DIR` unset, scope falls to `detect_project_dir()`, which derives from `std::env::current_dir()` (physical, symlink-resolved) — this can differ from a stored `CLAUDE_PROJECT_DIR` that was a logical/symlinked path, or from a session recorded in a subdirectory (`$PWD` != root). In those cases equality yields "No sessions found" where the old substring matched. Given ADR-154 mandates equality and the substring bug was worse, I judge this acceptable — flagging only because it's a real behavior change worth a doc line, and because that path surfaces as an empty list rather than the fail-loud error (the fail-loud only covers *undetected* project, not *detected-but-mismatched*).

### What checks out (verified, not assumed)

- **Union = safe concat, no double-count.** `migrate_exec.rs` `phase_lift_state`→`move_path` genuinely moves (`fs::rename`, else copy+`remove_any`) — never copies-and-keeps. Writes are deterministic per `resolve_events`: once `$XDG_STATE/events.jsonl` exists, every writer (Rust `log_event` + both hooks via `events-log-path`) targets it, so legacy is frozen at migration and shares no lines with preferred. `resolve_event_sources` also guards `legacy != preferred`. No path double-counts `count_fires`.
- **`load_events_text` join is merge-safe.** `join("\n")` always inserts exactly one separator between file contents, so a trailing newline in file 1 produces an inert empty line, never two JSON objects on one line. `filter_map` drops the empty line.
- **Deployment atomic.** `events-log.sh` is git-tracked; `clear-markers.sh`, `inject-subagent.sh`, and the new helper all land in the *same* commit (a71fb3d) — no pull-window where a hook sources a missing file. Ships via the `hooks/ways` tree projection; no `settings.json` registration needed (sourced helper, like `sessions-root.sh`).
- **Shell fallback correct.** Neither hook uses `set -e`/`set -u`, so empty `EVENTS_LOG` mid-resolution is harmless. An old binary predating `events-log-path` exits non-zero with empty stdout → `EVENTS_LOG` stays empty → `${EVENTS_LOG:-legacy}` fires. All expansions quoted; `_ways_bin` is underscore-scoped and unset at the end.
- **main.rs `(list, json)` routing** is correct: `(true,true)`→`run_list_json`, `(false,true)`→`run_json`, `_`→`rethink::run` (covers text `--list` and interactive). Doesn't regress the existing `--json` dump.
- **Removed import clean.** `detect_project_dir` dropped from `rethink_dump.rs`; still legitimately used in `rethink.rs:126`. No dangling refs; new `pub(crate)` surface is minimal.

**Assessment**: Sound, correctly-scoped increment. Only Finding 1 (dedup `--list --json` by session id) is worth addressing before the introspection increments build on this JSON contract; everything else is a note.

— reviewed by code-reviewer (analysis only; no files modified)

PR #273, opened 2026-07-03
feat(introspection): SessionIntrospection model in ways-core (increment 2)

aaronsb:
Increment 2 of the session-introspection work (ADR-153 §2, plan in docs/design-notes/session-introspection-implementation-plan.md). Builds the typed model; no UI (that's increments 4–6).

## What this adds

A new `ways-core::introspection` module — one typed, serde-serializable model joining the firing-event log to way frontmatter:

```
SessionIntrospection { id, project, window_k, summary, turns }
  └─ Turn { epoch, token_position, ts, transcript_uuid?, join_confidence, fired_ways }
       └─ FiredWay { way_id, trigger_channel, fire_score?, way_path, criteria, match_detail? }
```

It lives in ways-core (not the CLI) because the ADR-201 finding pipeline shares this exact correlation — building it once keeps findings and introspection from drifting (ADR-153 §4).

## Join honesty (the load-bearing discipline)

- The turn↔fire link is grouped by the **same `≤3s` timestamp-gap rule** `rethink::build_frames` uses, so the model and the replay agree on epoch boundaries.
- Every `Turn` is labelled `JoinConfidence::Heuristic`, and `transcript_uuid` / `match_detail` stay `None` — until fire-time enrichment (increment 3) records a foreign key. The types exist now so the shape is stable when enrichment lands.
- Semantic fires carry only a way-level `fire_score` — **no fabricated matched term** (one embedding per way).

## MatchCriteria

Unifies the two inconsistent frontmatter representations that exist today (the typed `Frontmatter`, which omits pattern/files/commands/trigger; and the untyped `WayCandidate` scan) into one honest surface, with a per-field-lenient parser: list values (`files: [a, b]`) join with `, `, and one bad-typed field can't nuke the rest.

## Scope boundary

This does **not** rewire `rethink`'s replay pipeline (still in ways-cli) to consume the model — that convergence is **increment 4**. Until then the clustering here deliberately mirrors `build_frames` so the two don't diverge. `build()` takes an injectable `CriteriaMap` so the join is unit-testable without touching disk; `from_session`/`default_criteria_map` wire the live event log (the increment-1 union reader) + the shipped/user way corpus.

## Test plan

- `cargo test -p ways-core`: 91 pass (5 new). New tests: lenient criteria parsing (lists / missing / bad-type / empty), frontmatter-block extraction, the clustering join (two epochs, cross-session filtering, string-encoded `fire_score`/`token_position`, unresolved-way → empty criteria not a panic), empty-events.
- `cargo clippy -p ways-core`: clean.
- `cargo build --workspace`: clean.

### DIFF
--- a/tools/ways-core/src/introspection.rs
+++ b/tools/ways-core/src/introspection.rs
@@ -20,8 +20,17 @@
 //!
 //! This increment builds the model + the join from the event log; it does not yet
 //! rewire the `rethink` replay pipeline (still in `ways-cli`) to consume it — that
-//! convergence is increment 4. Until then the epoch clustering here deliberately
-//! mirrors `rethink::build_frames`' `≤3s` rule so the two do not diverge.
+//! convergence is increment 4.
+//!
+//! **The clustering shares the `≤3s` gap *rule* with `rethink::build_frames`, but
+//! is not identical to it** — the two are reconciled at increment 4, not assumed
+//! equivalent now. Deliberate differences: this model clusters the *fire* stream
+//! only (a [`Turn`] is defined by the ways that fired, not by `session_start` /
+//! `way_redisclosed` events the replay also folds in), sorts by timestamp, numbers
+//! epochs over fire-bearing clusters only, and takes each turn's token position
+//! from the event's own recorded value rather than a transcript-timeline lookup.
+//! Increment 4 must reconcile these where the surfaces need to agree — it cannot
+//! assume a drop-in swap.
 
 use serde::Serialize;
 use serde_json::Value;
@@ -152,9 +161,10 @@ impl SessionIntrospection {
     ///
     /// The turn↔fire join is HEURISTIC — every [`Turn`] is labelled as such —
     /// until enrichment (ADR-153 §3) supplies a transcript uuid. Fires are grouped
-    /// into turns by the same `≤3s` timestamp-gap rule `rethink::build_frames`
-    /// uses, so the model and the replay agree on epoch boundaries. A way absent
-    /// from `criteria` still appears, with empty criteria — unknown, not invented.
+    /// into turns by the same `≤3s` timestamp-gap *rule* `rethink::build_frames`
+    /// uses (over the fire stream only — see the module note; the two are
+    /// reconciled at increment 4, not assumed identical). A way absent from
+    /// `criteria` still appears, with empty criteria — unknown, not invented.
     pub fn build(
         events: &[Value],
         session_id: &str,
@@ -317,16 +327,6 @@ impl MatchCriteria {
             embed_threshold: str_or_num_f64(&map["embed_threshold"]),
             scope: get("scope"),
         }
-    }
-
-    /// Parse criteria from a way file on disk (frontmatter block between the first
-    /// two `---` lines). Missing file or absent frontmatter → empty criteria.
-    pub fn from_way_file(path: &std::path::Path) -> Self {
-        std::fs::read_to_string(path)
-            .ok()
-            .and_then(|c| frontmatter_block(&c))
-            .map(|fm| Self::from_frontmatter_str(&fm))
-            .unwrap_or_default()
     }
 }
 
--- a/tools/ways-core/src/introspection.rs
+++ b/tools/ways-core/src/introspection.rs
@@ -211,10 +211,11 @@ impl SessionIntrospection {
     }
 
     /// Convenience: build for a session from the live event log (the increment-1
-    /// union reader) and the default way corpus.
+    /// union reader) and the project-aware way corpus (so a fired way resolves to
+    /// the criteria that actually fired, honoring ADR-143 override precedence).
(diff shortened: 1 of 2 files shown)


aaronsb:
## Review: increment 2 (SessionIntrospection model)

Reviewed the module in full against the real fire path, not the PR text. `cargo test -p ways-core` → 92 pass; `cargo clippy -p ways-core --all-targets -- -D warnings` → clean.

**Checks that hold (not defects, recording so they aren't re-litigated):**
- **ID join (risk 1).** `build_criteria_map` keys via `path_to_id(path.parent().strip_prefix(root))` — byte-identical to `candidates::way_id_from_path` (candidates.rs:238-241), incl. top-level `build/build.md` → `build`. The `criteria.get(&f.way)` join matches the bare, domain-prefixed id that `show/mod.rs` and `scan/state.rs` record as `way`. Verified.
- **Recognition parity (risk 2).** Both sides gate on `util::has_frontmatter` and skip `.contains(\".check.\")`; trigger-only ways (`meta/todos`, no `description:`) resolve. The `.check.` skip is safe precisely because id derives from the parent dir, so the base way file supplies the same id. Verified.
- **Field extraction (risk 4).** `session::log_event` (session.rs:436-438) writes *every* field as `Value::String`, so string-first parsing is correct; field names all match. State fires carry no `token_position` → `.unwrap_or(0)`, and `0` is only ever surfaced at turn level (max over the cluster), documented honestly. Verified.
- **MatchCriteria (risk 5).** Malformed/non-map/non-object YAML yields empty, never panics (`serde_json` `Index<&str>` returns `Null` for non-objects; bad YAML → `.ok()` → `Null`). List joins with `, `; map-valued fields → `None`; `embed_threshold` parses from string and number. Verified.
- **Epistemic honesty (risk 6).** Every `Turn` is `Heuristic`; `transcript_uuid`/`match_detail` always `None`; semantic fires carry only `fire_score`, no fabricated span. Verified.

---

### Finding 1 — Medium: clustering does NOT actually match `build_frames`, despite the module doc's claim

**Location:** introspection.rs:24 / 165-191 vs rethink.rs:460-503 (`reconstruct_frames` → `build_frames`).

**Problem.** The doc asserts the `≤3s` clustering \"deliberately mirrors `rethink::build_frames` ... so the two do not diverge.\" The *gap rule* is identical and `parse_ts_secs` is a faithful copy — but the **input to the clustering differs**, so the epoch groupings and numbers diverge:

1. **Input set.** `build` filters to `event == \"way_fired\"` **and sorts by `ts`**. `build_frames` clusters over the *entire* session event stream (`load_session_events`, rethink.rs:643-660, no event-type filter) in **file order, unsorted** (`reconstruct_frames` passes events straight through). `session_start` and every `way_redisclosed` sit in that stream and step `last_ts_secs`.

   Concrete failure: fire@t0, `way_redisclosed`@t3, fire@t5.
   - `build_frames`: t0 → t3 (`3 > 0+3`? no) → t5 (`5 > 3+3`? no) = **one epoch**.
   - `build`: fires only → t0, t5 (`5 > 0+3`? yes) = **two epochs**.

2. **Epoch numbering.** `build_frames` does `epoch += 1` for *every* cluster, including fire-less ones (a lone `session_start` cluster). `build` numbers only fire clusters. So with `session_start`@t0 and first fires@t100, the same fire is epoch 2 in rethink and epoch 1 here — even when groupings otherwise align.

3. **Turn token position.** `build` uses `max()` of recorded `token_position`; `build_frames` derives it via `find_token_position(token_timeline, cluster_ts)` (rethink.rs:511) — a different value with different meaning.

**Why it matters.** Harmless today (nothing consumes the model yet), but increment 4 makes `rethink` consume this. A silent epoch-boundary / epoch-number mismatch then reads as a replay regression. This is exactly the drift the module was written to prevent.

**Suggestion.** Either (a) soften the doc to state only the *gap rule* is shared and that input-set/epoch-numbering reconciliation is deferred to increment 4, or (b) cluster over the same event set `build_frames` sees (all session events, file order) and number epochs identically, so increment 4 is a lift-and-share rather than a behavior change. (a) is the honest minimum for this increment.

### Finding 2 — Low: `MatchCriteria::from_way_file` is unused and untested

**Location:** introspection.rs:324-330.

`build_criteria_map` reads each file once and calls `frontmatter_block` + `from_frontmatter_str` directly, so `from_way_file` is dead across the whole workspace (grep confirms no caller) and no test exercises it. It's `pub`, so clippy stays silent. Either drop it or add a test if it's intended staged API — as-is it's untested surface that will quietly rot.

---

**Assessment.** Model shape, join, recognition parity, string-first field parsing, lenient criteria parsing, and the epistemic-honesty labelling are all sound and well-tested. The one substantive issue is the overstated \"do not diverge\" clustering claim (Finding 1) — a latent contract mismatch worth resolving or re-wording before increment 4 builds on it. Nothing here blocks merge.

PR #274, opened 2026-07-03
feat(introspection): 3a — post-hoc transcript_uuid keyed joins + ADR-153 §3 correction

aaronsb:
Increment 3a of session-introspection. Investigation (see the ADR-153 §3 amendment in this PR) found the original §3 premise wrong — the `UserPromptSubmit` hook exposes no message uuid — but a **better** source: the transcript records each hook injection with a `parentUuid` chain back to its user message. So `transcript_uuid` is resolved **post-hoc from the transcript**, not at fire time.

## What this delivers

- **`ways-core::transcript`** — a read-only transcript reader that finds every `UserPromptSubmit` injection and walks its `parentUuid` chain back to the triggering `user` message (bounded against cycles). Returns `(timestamp, user-message uuid)` per prompt; absence/truncation → empty (never an error).
- **`SessionIntrospection::join_transcript`** — matches each turn to the nearest injection by session + fire timestamp (the one heuristic step: the fire happens *inside* the hook, sub-second) and flips matched turns `Heuristic`→`Keyed` with the real user-message uuid. Unmatched turns keep `Heuristic`. `from_session` wires it; summary gains `keyed_turns`.
- **ADR-153 §3 corrected** to record the post-hoc approach (+ the Alternatives entry and the implementation plan).

## Why post-hoc beats fire-time

No hot-path change, and it works for **historical sessions** whose transcript survives. The attachment→message link is a genuine keyed foreign key; only turn→injection matching is heuristic (and reliable — fires and their injection share a hook execution).

## Honesty

`Keyed` only when a real user-message uuid is found; everything else stays `Heuristic`. Semantic fires still carry only `fire_score`. Smoke on a real 269-turn session: **132 keyed (49%)** with real uuids — the other 51% are bash/file/state-triggered fires with no `UserPromptSubmit` injection to key against, correctly left heuristic rather than force-keyed.

## Test plan

- `cargo test -p ways-core`: 95 pass. New: `parentUuid`-walk (incl. non-UPS attachments + unresolvable chains + garbage lines), turn↔injection matching (keyed / out-of-tolerance / empty-transcript no-op).
- `cargo clippy -p ways-core`: clean. `cargo build --workspace`: clean.

Next: 3b (matched_span, fire-time enrichment) as a separate PR.

### DIFF
--- a/docs/architecture/system/ADR-153-session-introspection-substrate-correlating-fired-ways-to-turns.md
+++ b/docs/architecture/system/ADR-153-session-introspection-substrate-correlating-fired-ways-to-turns.md
@@ -93,21 +93,43 @@ grain: keyed where a key exists, heuristic (time/token bucket) where it does not
 and every heuristic edge is labelled as such in the model so a consumer never
 mistakes a proximity guess for a foreign key.
 
-### 3. Enrich `way_fired` at fire time to make "why" precise
+### 3. Make "why" precise — `transcript_uuid` post-hoc, `matched_span` at fire time
 
-The precise "what caused this hook to fire" is unanswerable from today's log. Add,
-at the fire sites (`cmd/show/mod.rs`, `cmd/scan/mod.rs`):
+> **Correction (implementation finding, 2026-07-03).** The original §3 assumed the
+> fire path could record a `transcript_uuid` "because the hook receives the message
+> id on stdin." It does not: the real `UserPromptSubmit` hook payload carries only
+> `prompt`, `session_id`, `cwd`, `agent_id` — no message uuid (the prompt's uuid is
+> assigned by Claude Code, after the hook). Investigation of a live transcript found
+> a *better* source, so the two halves are sourced differently.
 
-- **`transcript_uuid`** — the triggering message id (the hook receives it on
-  stdin). Converts the heuristic time-join into a foreign key.
-- **`matched_span`** — for keyword/command/file channels, the regex/glob match
-  text. The only way to show the exact clip without transcript replay.
+- **`transcript_uuid` — resolved *post-hoc* from the transcript, not at fire
+  time.** Every `UserPromptSubmit` hook injection is recorded in the session
+  transcript as an `attachment` (its `.attachment.content` holds the injected way
+  bodies), and its `parentUuid` chain walks back to the triggering `user` message
+  (verified empirically). The introspection model reads the transcript, matches each
+  turn — its fire-timestamp cluster — to the corresponding `UserPromptSubmit`
+  attachment, and follows `parentUuid` to the user-message uuid: a genuine foreign
+  key. This is strictly better than fire-time capture — no hot-path change, and it
+  works for historical sessions whose transcript survives. Only the turn→attachment
+  step is heuristic (session + sub-second timestamp — the fire happens *inside* the
+  hook); the attachment→message link is keyed. A turn whose transcript is
+  absent/unmatched stays `Heuristic`.
+
+- **`matched_span` — recorded at fire time** (`cmd/show/*`, `cmd/scan/*`). The
+  transcript's injected content is way-*anonymous* (concatenated bodies for the
+  whole prompt), so *what text matched an individual way* cannot be recovered
+  post-hoc. For the keyword/command/file channels the fire path captures the
+  regex/glob match text — additive, forward-only (old records lack it), kept cheap
+  and line-atomic on the hot path.
+
 - Semantic stays **way-level**: `fire_score` ≥ `embed_threshold` is the honest
   grain; per-vocabulary-term attribution is impossible (one embedding per way) and
   must not be faked.
 
-Enrichment is **additive and forward-only** — old records lack the fields, the
-model marks their join heuristic, and no backfill is claimed.
+Both enrichments are **additive**: the model uses each field when present and falls
+back to the heuristic time/token-bucket grain when absent. The post-hoc transcript
+join degrades to `Heuristic` when the transcript is unavailable; `matched_span`
+claims no backfill.
 
 ### 4. Share the substrate with the compliance finding pipeline (ADR-201)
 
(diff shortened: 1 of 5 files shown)


aaronsb:
## Review — increment 3a (post-hoc transcript_uuid keyed joins)

Reviewed the two Rust files in full, ground-checked assumptions against a real 19,367-line transcript (191 distinct prompts / 224 `UserPromptSubmit` injections), and ran `cargo test -p ways-core` (95 pass), `cargo clippy -p ways-core --all-targets -- -D warnings` (clean). Build/test/clippy all green.

Most of the design holds up well under the ground-truth data — see "Verified sound" below. Two findings, one of which is the epistemic-honesty question the increment is really about.

### Finding 1 (MEDIUM — honesty): rapid/colliding prompts get a confident `Keyed` uuid that may be wrong for part of the turn

**Location**: `introspection.rs` `nearest_prompt_turn` + turn clustering in `build` (`secs > last_secs + 3`); `JOIN_TOLERANCE_SECS = 8`.

**Scenario, from the real transcript**: two *distinct* user prompts whose `UserPromptSubmit` injections are recorded **0.38s apart** exist in the ground-truth session (1 such collision across 191 prompts, ~0.5%). Because turn clustering merges fires ≤3s apart, both prompts' fires collapse into a **single** introspection turn. `join_transcript` then flips that one merged turn to `Keyed` with **one** of the two user-message uuids (whichever injection wins `min_by_key`), silently dropping the association to the other prompt's fires. The attached uuid is a real key — but it is not the correct foreign key for every fire in that conflated turn.

**Why it matters**: ADR-153's core discipline is "a consumer never mistakes a proximity guess for a foreign key." The ADR text is honest (it calls turn→attachment heuristic). But the PR body's "and reliable — fires and their injection share a hook execution" overstates: the turn→injection step has a real, reachable failure mode when two prompts fall inside the clustering/tolerance window, and the *result* is still labelled `Keyed` with no ambiguity signal. That is the one case where `Keyed` can carry the wrong (adjacent) uuid.

**Suggestion (keeps the discipline exact, small change)**: make `nearest_prompt_turn` ambiguity-aware — if two or more injections with **different** `user_uuid` fall within tolerance of the turn, leave the turn `Heuristic` (or add an `ambiguous` marker) rather than confidently picking the minimum. Then `Keyed` means "unambiguous foreign key," which is what the label promises. Alternatively, if you accept the ~0.5% risk, document the collision caveat next to `JOIN_TOLERANCE_SECS` so the overclaim is at least explicit.

### Finding 2 (LOW — safe-degrading): `parse_ts_secs` month*30 collision now affects keying at month boundaries

**Location**: `introspection.rs` `parse_ts_secs` (reused by `nearest_prompt_turn`).

`days = year*365 + year/4 + month*30 + day` makes the 31st of a 31-day month collide with the 1st of the next (`Jan31 -> 61`, `Feb01 -> 61`; verified). Two consequences now that this feeds an 8s-tolerance join:
- A turn/injection pair straddling a month-boundary midnight computes `abs_diff ~= 86396s` for a real 4s gap -> **falsely stays `Heuristic`** (safe direction, just loses a key).
- Theoretically, same-clock-time activity on the 31st vs next-month-1st collides to a tiny diff -> a wrong-positive match, but that needs sub-second-aligned activity 24h apart on that exact boundary — practically unreachable.

Net: safe-degrading, not silently-wrong. The code already flags increment 4 to hoist a single real date parse; that resolves it. Noting so it's tracked.

### Verified sound (checked against real data — no defect)

- **parentUuid walk (`walk_to_user`)**: all 224 real UPS injections resolve to a user message; max chain length **6 hops** — the 64-hop bound is comfortably safe. Field paths `.type=="attachment"` + `.attachment.hookEvent=="UserPromptSubmit"` match the real transcript exactly. Walk stops at the first `type=="user"` and reaches **0** `tool_result` user messages — it lands on real prompts, not tool-result user nodes. Missing/cyclic parents return `None` safely.
- **Transcript path slug**: `replace(['/','.'], "-")` reproduces the real dir names, including the `/.dotfiles -> --dotfiles` double-dash case. On a slug miss the read errors -> empty Vec -> no-op; session ids are globally unique UUIDs so a wrong slug can never read a *different* session's file. Failure mode is safe, not silently-wrong. (Underscore/space paths, if Claude Code slugs them beyond `/`+`.`, would safely degrade to all-heuristic — pre-existing shared assumption, not introduced here.)
- **Multiple injections per prompt**: 33 prompts have 2 UPS injections; both resolve to the same `user_uuid`, so `min_by_key` returns the correct key regardless of which wins. No issue.
- **Empty/absent transcript**: genuine no-op (`join_transcript(&[])` leaves everything `Heuristic`, `keyed_turns = 0`) — covered by test.
- **Semantic fires**: untouched; still carry only `fire_score`, no fabricated term.
- **serde back-compat**: `IntrospectionSummary`/`SessionIntrospection` derive `Serialize` only (no `Deserialize`), and `keyed_turns` is additive with `Default`. Purely additive to JSON output — no consumer break.
- **Perf/memory on 19k lines**: single O(n) pass building a per-line `(parent, type)` map — bounded by transcript size, not an unbounded index. Fine at this scale. No `unwrap`/panic in new code.
- **ADR-153 §3 amendment**: accurately describes post-hoc resolution, the parentUuid walk, the heuristic turn->attachment / keyed attachment->message split, and the `Heuristic` fallback. It honestly labels the turn->attachment step heuristic (unlike the PR body's stronger "reliable" phrasing — see Finding 1).

### Assessment

Solid increment. The architecture (keyed FK from a real transcript chain, heuristic only on the sub-second turn->injection hop) is sound and the ground-truth data backs nearly every claim. The one thing worth resolving before this sets precedent is Finding 1: `Keyed` should mean *unambiguous* key, and the rapid-prompt collision is a real (if rare) case where it currently isn't. Recommend the ambiguity guard in `nearest_prompt_turn`; Finding 2 is fine to defer to increment 4.

*AI-assisted review via Claude Code — reviewer only; no files modified.*

PR #275, opened 2026-07-03
feat(introspection): 3b — matched_span fire-time enrichment (ADR-153 §3)

aaronsb:
Increment 3b — the fire-time half of ADR-153 §3, and the last piece of increment 3. Records **what text hit** a way, for the deterministic channels, so introspection can show *why* a way fired without transcript replay.

## Changes (hot path)

- New `regex_span(pattern, text)` — the first regex match, length-capped at 120 chars. `matched_span` is threaded through `PromptMatch::Fired` → `capture_show_way` → `way_scored` → the `way_fired` log block (first-fire only, mirroring `fire_score`).
- Channels: **keyword** → the regex match against the prompt; **bash** → the matched command (else the description-pattern match); **file** → the matched filepath; **semantic** → `None` (one embedding per way — no term to recover, never faked); **state / `ways show`** → `None`.
- Model: `FireRow` parses `matched_span`; a present span becomes a **Keyed** `MatchDetail` on the `FiredWay` (a fire-time record of exactly what hit). Forward-only — old records stay match-less.

## Privacy

`matched_span` is a fragment of the prompt/command/path. It stays in the **local** events log (never transmitted), is **capped** at 120 chars, and the matching pattern is **author-controlled** (a way's `pattern`), so a match is normally a bounded keyword. The cap guards against a greedy pattern dumping prompt text.

## Verification

Smoke on a real scan (`please write an adr and commit … to git`): keyword fires recorded `' adr '` and `'commit'`; all 7 semantic fires correctly `None`. So the "why" is precise for deterministic channels and honestly empty for semantic.

## Test plan

- `cargo test -p ways -p ways-core`: ways 201 unit + 7 integration pass; ways-core passes. New: keyword-records-span + semantic-records-none (`scan/mod.rs`), event `matched_span` → Keyed `MatchDetail` + semantic none (`introspection`). The 9 `session_sim` failures are pre-existing on main (environment-dependent), unchanged.
- `cargo clippy -p ways -p ways-core`: clean. `cargo build --workspace`: clean.

Completes increment 3. Next: increment 4 (`ways introspect <replay|live|dump>` surface).

### DIFF
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -105,8 +105,8 @@ pub fn prompt(query: &str, session_id: &str, project: Option<&str>) -> Result<()
             &embed_matches,
             near_miss_margin,
         ) {
-            PromptMatch::Fired { channel, score } => {
-                let out = capture_show_way(&way.id, session_id, &channel, score);
+            PromptMatch::Fired { channel, score, matched_span } => {
+                let out = capture_show_way(&way.id, session_id, &channel, score, matched_span.as_deref());
                 if !out.is_empty() {
                     context.push_str(&out);
                     context.push_str("\n\n");
@@ -244,26 +244,19 @@ pub fn command(
             continue;
         }
 
-        let mut matched = false;
+        // Commands regex first, then the description pattern — capture the span
+        // of whichever matched (ADR-153 §3).
+        let matched_span = way
+            .commands
+            .as_deref()
+            .and_then(|p| regex_span(p, cmd))
+            .or_else(|| match (description, way.pattern.as_deref()) {
+                (Some(desc), Some(pat)) => regex_span(pat, &desc.to_lowercase()),
+                _ => None,
+            });
 
-        if let Some(ref cmds_pattern) = way.commands {
-            if regex_matches(cmds_pattern, cmd) {
-                matched = true;
-            }
-        }
-
-        if !matched {
-            if let Some(desc) = description {
-                if let Some(ref pat) = way.pattern {
-                    if regex_matches(pat, &desc.to_lowercase()) {
-                        matched = true;
-                    }
-                }
-            }
-        }
-
-        if matched {
-            let out = capture_show_way(&way.id, session_id, "bash", None);
+        if let Some(span) = matched_span {
+            let out = capture_show_way(&way.id, session_id, "bash", None, Some(span.as_str()));
             if !out.is_empty() {
                 context.push_str(&out);
             }
@@ -348,8 +341,8 @@ pub fn file(filepath: &str, session_id: &str, project: Option<&str>) -> Result<(
         }
 
         if let Some(ref files_pattern) = way.files {
-            if regex_matches(files_pattern, filepath) {
-                let out = capture_show_way(&way.id, session_id, "file", None);
+            if let Some(span) = regex_span(files_pattern, filepath) {
+                let out = capture_show_way(&way.id, session_id, "file", None, Some(span.as_str()));
                 if !out.is_empty() {
                     context.push_str(&out);
                 }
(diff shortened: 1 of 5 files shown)


aaronsb:
## Code Review — increment 3b (matched_span fire-time enrichment)

Reviewed the hot fire path against the six risks. Build + targeted tests pass (`ways` 201 unit / 7 integration, `ways-core` green); `cargo clippy -p ways -p ways-core` clean. **No firing regressions, no panics, no missed callers.** Details below, then two minor notes.

### 1. Firing-behavior preservation in the bash block — VERIFIED equivalent
The `and_then(commands) → or_else(description-pattern)` rewrite preserves the original semantics exactly:
- **Precedence**: `and_then` on `way.commands` runs the commands regex first; a `Some` short-circuits `or_else`, so the description pattern is only tried on a commands miss — same as the old `if !matched` guard.
- **Fire condition**: `regex_span` returns `Some` iff `regex_matches` returned `true`, because `Regex::find(text).is_some()` and `Regex::is_match(text)` agree in the `regex` crate. So every path that fired before fires now, and none that didn't now does.
- **Invalid-regex tolerance**: `Regex::new(pat).ok()?` yields `None` on a bad pattern, falling through to the description path — identical to the old `unwrap_or(false)`. The description arm still requires *both* `description` and `pattern` present (`(Some, Some)`), matching the old nested `if let`.

### 2. regex_span panic-safety & empty-match — VERIFIED safe
- **(a) UTF-8 boundary**: `char_indices().nth(MAX_SPAN)` returns a byte offset that is always a char boundary, so `&s[..byte]` cannot split a multi-byte codepoint. No panic on multi-byte input. (Keeps 120 chars + `…`.)
- **(b) Zero-width match**: for a pattern that matches empty (`^`, `.*`, `a*`), `find` returns `Some("")` and `is_match` is `true` — so the **firing decision is unchanged** vs the old `regex_matches`. The keyword channel's condition did not subtly shift.

### 3. Redisclosure gating — VERIFIED correct
`if let (false, Some(span)) = (is_redisclosure, matched_span)` mirrors the `fire_score` block directly above it. A redisclosure records no span, which is the intended convention (the re-trigger is a different match).

### 4. Privacy — within ADR-153 framing
The bash description path captures `regex_span(pat, &desc.to_lowercase())` — a bounded match **within the tool description**, not arbitrary command output. Commands/keyword/file paths each capture a match within author-controlled-pattern bounds, capped at 120 chars, local-only. No channel captures more than the ADR's "author-controlled pattern match" framing. The 120-char cap is a reasonable greedy-pattern guard.

### 5. Model side — VERIFIED correct
`FireRow.matched_span` parses `v["matched_span"]`; `build_turn` wraps a present span as `MatchDetail{ matched_span: Some, confidence: Keyed }`. `Keyed` is the right label — it is a fire-time deterministic record of exactly what hit (distinct from the separately-computed turn-join confidence). Semantic fires carry `None` → `match_detail: None`; nothing is fabricated, and pre-enrichment records stay match-less (forward-only).

### 6. Consistency — VERIFIED
Field name `matched_span` matches between writer (`show/mod.rs`) and reader (`introspection.rs FireRow`). All callers updated: `way()`, `way_scored`, `capture_show_way`, `state.rs` (`None`), and both scan-mod call sites. All `PromptMatch::Fired` constructions/destructures compile. No other-crate or cfg-gated caller exists.

---

### Minor notes (non-blocking)

- **Empty-span telemetry noise**: an author pattern that matches empty (e.g. `.*`, `a*`) fires exactly as before, but now records `matched_span: ""`, which the model surfaces as a `Keyed` `MatchDetail` with an empty span — a "why" that says nothing. Harmless and only reachable via a mis-authored fire-on-everything pattern, but if you want the field to always carry signal, consider dropping empty matches (`.filter(|s| !s.is_empty())`) in `regex_span` or at the log-push. Firing is unaffected either way.
- **Pre-existing, not this PR**: `cargo clippy --all-targets -- -D warnings` fails on `ways-cli/src/cmd/settings/compile.rs:334` (`unnecessary_get_then_check`, new in clippy 1.92) and 9 `session_sim` scenarios fail on missing session-marker files — both are environment/toolchain artifacts unrelated to this change (the stated `clippy -p ways -p ways-core` without `--all-targets` is clean). Worth a separate cleanup, not a gate here.

**Assessment**: The high-risk items (firing preservation and the empty-match edge) resolve cleanly — the `find`/`is_match` equivalence is the load-bearing fact and it holds. Tests cover keyword-records-span, semantic-records-none, and the event→`MatchDetail` mapping. Solid to merge.

PR #276, opened 2026-07-03
feat(introspect): ways introspect dump — agent-facing MVP over the model

aaronsb:
The first slice of increment 4 (ADR-154 §1): a user/agent-facing surface over the ways-core `SessionIntrospection` model, so everything built in increments 1–3 is usable today.

## What this adds

`ways introspect dump [--session <id>] [--project <path>] [--all]` — emits a session's reconstructed introspection as JSON: turns, fired ways, their criteria, the keyed transcript join (`transcript_uuid` + `keyed_turns`), and matched spans. Scoping mirrors `rethink` (default current project, `--project`, `--all`); no `--session` dumps the most recent in scope; fail-loud as a JSON error.

New `cmd/introspect` module + CLI wiring. The heavy lifting is all in the already-merged ways-core model — this is thin glue.

## Two fixes surfaced by end-to-end testing

- **Transcript resolution by session-id scan.** The recorded `session_start` project can be a stray subagent/hook cwd that differs from the dir the transcript actually lives in, so the lossy project-slug path missed → `keyed_turns: 0`. `transcript.rs` now falls back to scanning `projects/*` for `<session>.jsonl` (session ids are globally unique → exact). This is the "deferred until needed" fallback ADR-153 §3 anticipated.
- **Project-path precedence.** `introspect::dump` prefers the caller's scope (explicit `--project` / detected current) over `find_session_project`, fixing a misleading `project` field and a wrong criteria root.

## Verification

Dump of a real 270-turn session → **131 keyed turns** with real user-message uuids, correct `project`, matched spans where recorded (older fires predate 3b → `None`, honestly forward-only).

## Test plan

- `cargo test -p ways -p ways-core`: 201 + 7 pass; ways-core passes. `cargo clippy -p ways -p ways-core`: clean. The 9 `session_sim` failures are pre-existing on main.
- The dump command is thin IO glue; its components (model, transcript, criteria) are unit-tested. Manually verified end-to-end.

Scope: `replay`/`live` modes, the `rethink` deprecated-alias, and the `build_frames` clustering reconciliation remain for the rest of increment 4.

### DIFF
--- a/tools/ways-cli/src/cmd/introspect.rs
+++ b/tools/ways-cli/src/cmd/introspect.rs
@@ -1,0 +1,63 @@
+//! `ways introspect` — the user/agent-facing surface over the ways-core
+//! `SessionIntrospection` model (ADR-154 §1). This increment ships only the
+//! non-interactive `dump` mode (the agent-facing MVP); `replay`/`live` and the
+//! why-fired drill-down follow.
+
+use anyhow::Result;
+
+use crate::cmd::{rethink, rethink_dump};
+use crate::session;
+
+/// `ways introspect dump` — emit a session's reconstructed introspection (turns,
+/// fired ways, criteria, keyed transcript join, matched spans) as JSON, so an
+/// agent can investigate *which ways fired, on which turn, and why* without a TUI.
+///
+/// Scoping mirrors `rethink`: default the current project, `--project` for a
+/// specific one, `--all` across every project (which only affects session
+/// picking). With no `--session`, the most recent session in scope is dumped.
+pub fn dump(session: Option<&str>, project: Option<&str>, all: bool) -> Result<()> {
+    let content = ways_core::firing::load_events_text();
+    if content.trim().is_empty() {
+        println!("{{\"error\":\"no events recorded yet\"}}");
+        return Ok(());
+    }
+
+    // Fail-loud scope resolution, as JSON (agent-facing).
+    let scope = match rethink::resolve_project_scope(project, all) {
+        Ok(s) => s,
+        Err(e) => {
+            println!("{{\"error\":{}}}", serde_json::to_string(&e.to_string())?);
+            return Ok(());
+        }
+    };
+
+    let session_id = match session {
+        Some(s) => s.to_string(),
+        None => match rethink_dump::most_recent_session(&content, scope.as_deref()) {
+            Some(s) => s,
+            None => {
+                println!("{{\"error\":\"no sessions found in scope\"}}");
+                return Ok(());
+            }
+        }
+    };
+
+    // Project path drives the criteria corpus and the transcript slug. Prefer the
+    // caller's scope (explicit `--project` or the detected current project) — it's
+    // authoritative — over the session's first recorded `session_start` project,
+    // which can be a stray subagent/hook cwd. The transcript reader scans by
+    // session id if the slug misses, so a wrong path here still resolves the join.
+    let project_path = scope
+        .clone()
+        .or_else(|| rethink::find_session_project(&content, &session_id))
+        .unwrap_or_default();
+    let window_k = session::detect_context_window_for(&project_path, &session_id) / 1000;
+
+    let model = ways_core::introspection::SessionIntrospection::from_session(
+        &session_id,
+        &project_path,
+        window_k,
+    );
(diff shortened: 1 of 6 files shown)


aaronsb:
## Review: introspect dump MVP

**What this changes**: adds `ways introspect dump` (thin JSON glue over the already-merged `SessionIntrospection` model) plus a transcript-resolution fallback in `ways-core::transcript`. Verified against the real code, built, clippy'd, tested, and smoke-tested the CLI.

**Assessment: no blocking defects.** The two highest-risk areas both degrade gracefully. Details below.

### Verified safe

- **Session-id scan (`transcript.rs::transcript_path`) cannot match a wrong session.** `read_dir(root).flatten().map(|e| e.path().join("<session_id>.jsonl")).find(is_file)` joins the *exact* `<session_id>.jsonl` filename onto each project dir — no substring/prefix matching, so a different session can never be selected. Session ids are UUIDs (globally unique), so "first match wins" is effectively deterministic; a cross-project collision would require the identical UUID transcript to exist under two project dirs, which Claude Code doesn't produce.
- **Slug-hit case is preserved and identical to before** — `if direct.is_file() { return Some(direct); }` short-circuits before the scan, so 3a's keyed joins are byte-for-byte unchanged.
- **Robustness**: `read_dir` on a missing/non-dir root → `.ok()?` → `None` → empty `prompt_turns`, no panic. A stray file under `projects/` → `filepath.join(...)` then `is_file()` is false → skipped harmlessly.
- **JSON surface is clean.** All five stdout paths emit valid JSON (four `{"error":...}` branches + `to_string_pretty(&model)`); the error string is serialized via `serde_json::to_string` so it's always well-formed. No `println!`/`print!` exists in the ways-core call path (introspection/transcript/firing) or in `session.rs` to pollute stdout. `--session <bogus>` returns a valid empty-ish model (`{id, project, window_k, summary, turns}`) with no panic — smoke-tested.
- **`most_recent_session`** change is signature-only (`pub(crate)`); body is untouched, behavior unchanged.
- **clap wiring**: both `IntrospectCommand` and `ScanCommand` derive `Subcommand`; `introspect dump` parses and `--help` renders. `--all` for a single-session dump only affects which session is picked, which is coherent and documented in the help text. No new `unwrap`/`panic` in touched code.

### Finding 1 — Low / latent: empty-project criteria pulls in a CWD-relative `.claude/ways`

**Location**: `introspect.rs` `dump()` (`project_path = scope.or_else(find_session_project).unwrap_or_default()`) → `introspection.rs:502 project_criteria_map`.

When scope is `None` (`--all`) **and** `find_session_project` returns `None` (session has no recorded `session_start` project), `project_path` is `""`. Then `project_criteria_map("")` computes `PathBuf::from("").join(".claude").join("ways")` = the **relative** path `.claude/ways`, resolved against the process CWD. So a `dump --all` for such a session, run from inside any repo that has a `.claude/ways`, silently folds that CWD project's ways into the criteria corpus.

**Impact**: harmless in practice (no panic; a session with no recorded project usually has no fires to enrich, and the transcript still resolves via the scan). But "empty project" isn't strictly "user+core only" as the surrounding comment implies — it's "user+core + whatever `.claude/ways` happens to sit in CWD." A latent correctness wrinkle, not a blocker.

**Suggestion** (optional): degrade strictly by only adding the project ways root when the project is non-empty, e.g. in `project_criteria_map` skip `PathBuf::from(project).join(".claude").join("ways")` when `project.is_empty()`. Leave user+core as the floor.

### Finding 2 — Note, not introduced by this PR

- `cargo clippy -p ways` emits one warning at `ways-cli/src/cmd/settings/compile.rs:334` (`unnecessary_get_then_check`), in a file untouched by this PR. It would only bite under `-D warnings`. Flagging so "clippy clean" is read as "clean on the touched files, one pre-existing warning elsewhere in the crate."
- 9 `session_sim` scenarios fail (`session_sim.rs:287`, marker-path/env harness). Confirmed pre-existing by running `scenario_1` on `main` — also fails. Not caused by #276; matches the PR description.

### Bottom line

Ship-able. The scan fallback is exact-match and safe, the empty-project path degrades without erroring, and every dump surface emits valid JSON. Finding 1 is worth a one-line guard if you want empty-project to mean strictly user+core, but it's non-blocking.

PR #277, opened 2026-07-03
feat(introspect): complete the surface (replay/list) + deprecate rethink + hoist parse_ts_secs

aaronsb:
Rounds out increment 4's non-interactive surface (ADR-154 §4). Builds on the merged dump MVP.

## Changes

- **`ways introspect <replay|list|dump>`** — `replay` (interactive TUI) and `list` (table or `--json`) join the shipped `dump`. `replay`/`list` delegate to the proven `rethink` pipeline for now; re-pointing them at the `SessionIntrospection` model is deferred until the drill-down needs it (it requires reconciling the model's fire-centric clustering with `build_frames`' full-event-stream clustering — ADR-154 flagged this as not-a-drop-in-swap). `dump` reads the model.
- **`ways rethink` deprecated** — now prints a stderr notice pointing at `introspect replay`; still fully works.
- **`parse_ts_secs` hoisted** to one ways-core `util` copy, made **calendar-correct** (cumulative month days + leap years). Removes the duplicate in `rethink.rs` + `introspection.rs` and fixes the month-boundary `abs_diff` collision noted in the 3a review (`Jan31 ≡ Feb01`). Within-second diffs — all the clustering and the transcript join actually use — are unchanged.

## Test plan

- `cargo test -p ways -p ways-core`: ways 201 + 7, ways-core 97 (new `parse_ts_secs_diffs_are_calendar_correct`). `cargo clippy -p ways -p ways-core`: clean.
- Smoke: `introspect list --json` returns the scoped session; `rethink` emits the deprecation notice on stderr and still runs.
- The 9 `session_sim` failures are pre-existing on main (environment-dependent).

Remaining for increment 4/5: re-point `replay` at the model (with the clustering reconciliation), the why-fired drill-down TUI, and `introspect live`.

### DIFF
--- a/tools/ways-cli/src/cmd/introspect.rs
+++ b/tools/ways-cli/src/cmd/introspect.rs
@@ -1,12 +1,38 @@
 //! `ways introspect` — the user/agent-facing surface over the ways-core
-//! `SessionIntrospection` model (ADR-154 §1). This increment ships only the
-//! non-interactive `dump` mode (the agent-facing MVP); `replay`/`live` and the
-//! why-fired drill-down follow.
+//! `SessionIntrospection` model (ADR-154). Modes: `replay` (interactive),
+//! `list`, `dump` (agent-facing JSON). `live` and the why-fired drill-down
+//! follow in later increments.
+//!
+//! `replay`/`list` currently delegate to the proven `rethink` pipeline
+//! (`build_frames` → TUI); re-pointing them at the `SessionIntrospection` model
+//! is deferred until the drill-down needs it, since it requires reconciling that
+//! model's fire-centric clustering with `build_frames`' full-event-stream
+//! clustering (ADR-154 §1 — not a drop-in swap). `dump` already reads the model.
 
 use anyhow::Result;
 
 use crate::cmd::{rethink, rethink_dump};
 use crate::session;
+
+/// `ways introspect replay` — interactive replay of a session's way firings.
+pub fn replay(
+    session: Option<&str>,
+    project: Option<&str>,
+    all: bool,
+    speed: Option<u64>,
+) -> Result<()> {
+    rethink::run(session, project, speed, false, all)
+}
+
+/// `ways introspect list` — enumerate candidate sessions in scope, as a table or
+/// (`--json`) machine-listable data for an agent to pick from before dumping.
+pub fn list(project: Option<&str>, all: bool, json: bool) -> Result<()> {
+    if json {
+        rethink_dump::run_list_json(project, all)
+    } else {
+        rethink::run(None, project, None, true, all)
+    }
+}
 
 /// `ways introspect dump` — emit a session's reconstructed introspection (turns,
 /// fired ways, criteria, keyed transcript join, matched spans) as JSON, so an
--- a/tools/ways-cli/src/cmd/rethink.rs
+++ b/tools/ways-cli/src/cmd/rethink.rs
@@ -19,7 +19,7 @@ use std::io::Write;
 
 use crate::cmd::render::{self, WayRow};
 use crate::session;
-use crate::util::{detect_project_dir, home_dir};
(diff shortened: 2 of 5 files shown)


aaronsb:
## Code Review — increment 4 surface + parse_ts_secs hoist

Verified against the checked-out branch. Build/test/clippy run locally. **No blocking defects found** — the two highest-value checks (difference-preservation and stderr-not-stdout) both hold. Details below.

### 1. parse_ts_secs behavior preservation — PASS
- **(1a) Within-session diffs are byte-identical to the old fn.** Every caller computes a *difference*: `rethink` clustering (`ts_secs > last_ts_secs + 3`, rethink.rs:495), cluster elapsed (`saturating_sub`, :509), session duration (:748-750), introspection turn clustering (introspection.rs:194), and the transcript join `abs_diff` (:283). For two timestamps in the same year+month, both old (`month*30 + day`) and new (`CUM[m] + day + leap_adj`) hold every term but `day` constant, so the difference reduces to `(day2-day1)*86400 + hms` identically. Within-second and within-day diffs are trivially preserved. Confirmed.
- **(1b) Monotonic across boundaries.** The old `Jan31 ≡ Feb01` collision (both `+61`) is gone; the new test asserts `Feb01 - Jan31 == 86_400`. This is a latent-bug *fix*: a session spanning a month boundary previously mis-clustered day-apart events into one epoch. No regression, since all uses are within-session diffs.
- **(1c) Leap branch correct.** `m` is 0-based; `m >= 2` = March onward, `CUM[2]=59` + 1 = 60 days before Mar 1 in a leap year. Feb itself (`m=1`) correctly excluded. Verified `2024-03-01 - 2024-02-29 == 86_400`.
- **(1d) Session duration can't go negative or wild.** `end.saturating_sub(start)` (rethink.rs:750) with a now-monotonic encoding; floors at 0.
- **(1e) No panic on garbage.** `ts.len() < 19` guard + `ts.get(a..b)` (returns `None`, not a panic, on out-of-range or non-char-boundary bytes) → `unwrap_or(0)`.
- `util::parse_ts_secs` is `pub`; both local copies removed; `use crate::util::parse_ts_secs` added in both consumers; no unused imports.

### 2. introspect replay/list delegation — PASS
`rethink::run` is `(session, project, speed, list, all)`.
- `replay` → `run(session, project, speed, false, all)` — `list=false`, session forwarded. Correct.
- `list` (text) → `run(None, project, None, true, all)` — `list=true`, no session. Correct.
- `list --json` → `rethink_dump::run_list_json(project, all)`. Signature matches. Correct.

### 3. Deprecation notice — PASS (empirically verified)
`eprintln!` placed before the `match`, so it fires for every `rethink` invocation including `--json`. Ran `ways rethink --list --json` with fds split to files: stdout is valid JSON (`jq -e` passes), `grep -c deprecated` on stdout = 0, notice appears only on stderr. JSON consumers are safe.

### 4. General — PASS (one pre-existing, non-blocking note)
- `ways-core --lib` and `ways --bins` are clean under `clippy -- -D warnings`.
- `cargo test -p ways-core`: 97 pass incl. the new `parse_ts_secs_diffs_are_calendar_correct`.
- The 9 `session_sim` failures reproduce on `main` (environment-dependent), as the PR notes.
- **Note (not introduced here):** `clippy -p ways --all-targets -- -D warnings` fails on `ways-cli/src/cmd/settings/compile.rs:334` (`unnecessary_get_then_check`, a newer-clippy lint in untouched *test* code). If CI runs clippy with `--all-targets -D warnings`, it will be red for a reason unrelated to this PR — worth a separate cleanup.

### Assessment
Clean, well-scoped increment. The parse_ts_secs consolidation is a genuine DRY improvement that also fixes a latent month-boundary clustering bug while provably preserving the only property the callers depend on (within-session diffs). The deferral rationale for not re-pointing replay/list at the SessionIntrospection model is documented in the module header and matches ADR-154. Delegation, arg order, and the stderr contract are all correct.

PR #278, opened 2026-07-03
feat(introspect): why-fired drill-down + micro-compositor (increment 5)

aaronsb:
## Summary

Increment 5 of the session-introspection arc (ADR-153/154): the **why-fired drill-down** over the `SessionIntrospection` model, plus the **micro-compositor** it's built on. Tab into `ways introspect replay` (or the `rethink` alias), focus a fired way, and read *why* it fired — its trigger, resolved `MatchCriteria`, the matched span, and the way body.

## What's here

- **The model↔replay-timeline boundary, recorded** (ADR-154 §1 + the `introspection.rs` module note that promised "reconciled at increment 4"). Resolution: `build_frames` (cumulative animation) and the model (per-turn analytical substrate) are **two distinct views, joined at the drill-down by `way_id`** — *not* one clustering unified. No epoch alignment; the proven replay timeline is untouched. Converging them was considered and rejected (it would load the fire-stream substrate with animation concerns).
- **`cmd/compositor.rs`** — the ADR-154 §2 micro-compositor: a `Panel` (`Vec<String>` + visible width) with a fixed-height viewport, an `hjoin` side-by-side placer, a `tab_bar`, and the canonical ANSI-visible-width primitives (`visible_len`/`truncate_visible`/`pad_visible`/`fit_visible`) that `render.rs` and `rethink.rs` each grew local copies of. **Zero new deps** — the deliberate alternative to ratatui.
- **The drill-down in `rethink.rs`** — a `View` enum + drill-down state on `Player`; Tab toggles Timeline↔Why; the `way_id → criteria/spans/path` index is built **lazily** on first entry so a plain replay never pays for it. ▲▼ select a way, ◀▶ move frames without leaving the view, PgUp/Dn scroll the detail.

## Join honesty (load-bearing)

- Semantic fires show `fire_score` + "no recoverable term" — **never a fabricated matched term** (one embedding per way).
- A keyword/file fire with no recorded span says so, rather than inventing one.

Covered by `why_tests` (semantic vs keyword-with-span vs no-span, span dedup) and the compositor unit tests.

## Test plan

- `cargo test -p ways -p ways-core` — 215 unit + 11 compositor + 3 drill-down + 7 ways-core, all green.
- `cargo clippy -p ways --all-targets -D warnings` — clean on every touched file. (The one remaining error, `settings/compile.rs:334`, is a pre-existing newer-clippy lint in untouched test code, deferred to its own CI-health PR.)
- `session_sim` failures are pre-existing on clean `main` (verified via stash) — a local `/run/user` environment issue, not a regression.

Increment 6 (`ways introspect live`) is the remaining piece of task #20.

### DIFF
--- a/docs/architecture/system/ADR-154-rethink-think-and-non-interactive-introspection-one-model-three-front-ends.md
+++ b/docs/architecture/system/ADR-154-rethink-think-and-non-interactive-introspection-one-model-three-front-ends.md
@@ -61,6 +61,24 @@ Generalize the proven triplet into a module (mirroring the `cmd/scan`,
   - **drill-down** — a selection/tab loop compositing model panels, shared by
     `replay` and `live`.
 
+**Model ↔ replay-timeline boundary (resolved 2026-07-03).** The
+`SessionIntrospection` model and `rethink::build_frames` are **two distinct
+views, not one clustering** — and are deliberately kept that way rather than
+unified. `build_frames` is the *animation* projection: it folds the full event
+stream (`way_fired` + `check_fired` + `way_redisclosed`) into cumulative
+"what's active at epoch N" frames, and drives `replay`'s timeline. The model is
+the *analytical* substrate: per-turn fired-way deltas joined to `MatchCriteria`,
+`matched_span`, and the transcript key, and it drives `dump` and the drill-down's
+"why" panel. The drill-down bridges them **by `way_id`** — focus a way in a
+`build_frames` frame, and its detail is looked up in the model by id — so their
+divergent epoch numbering never has to be reconciled. This preserves the proven
+replay timeline untouched, keeps the model scoped to the fire stream (ADR-153),
+and confines the JSON-vs-TUI drift the model guards against to the two surfaces
+that actually share it (`dump` and the drill-down), both of which read the model.
+Converging `build_frames` onto the model was rejected: it would load the
+fire-stream substrate with cumulative-state, redisclosure, refire-threshold, and
+token-timeline concerns that belong to the animation view alone.
+
 ### 2. Keep crossterm; grow a small micro-compositor — do not adopt ratatui
 
 Build a ~200–400-line internal compositor over the existing ANSI-`String` panels:
--- a/tools/ways-cli/src/cmd/compositor.rs
+++ b/tools/ways-cli/src/cmd/compositor.rs
@@ -1,0 +1,328 @@
+//! Micro-compositor for terminal panels (ADR-154 §2).
+//!
+//! A tiny layout layer over the ANSI-`String` panels that `cmd/render` already
+//! produces. A [`Panel`] is a list of ANSI-styled lines plus its visible width;
+//! the helpers place panels side by side ([`hjoin`]), scroll a panel to a
+//! fixed-height viewport ([`Panel::viewport`]), and render a [`tab_bar`]. This is
+//! the deliberate zero-dependency alternative to ratatui (ADR-154 §2), right-sized
+//! for "list-left / detail-right with independent scroll" and "table + status bar"
+//! — the shapes the introspect drill-down needs. If the inspector ever grows to
+//! need text selection or resizable/mouse panes, the ADR's escape hatch to ratatui
+//! applies; short of that, this stays.
+//!
+//! It owns the canonical ANSI-visible-width primitives ([`visible_len`],
+//! [`truncate_visible`], [`pad_visible`]) that `render.rs` (`ansi_pad`/
+//! `ansi_visible_len`) and `rethink.rs` (`truncate_visible`) each grew their own
+//! copies of. Width is measured in `char`s, ignoring SGR escapes — the same grain
+//! those callers already use; East-Asian double-width and combining marks are not
+//! accounted for (that would need a new dependency the lean binary declines).
+
+use std::fmt::Write;
+
+// ── ANSI-visible-width primitives ─────────────────────────────
+
+/// Visible length of `s` in `char`s, ignoring ANSI escape sequences (`\x1b[…X`,
+/// terminated by an ASCII letter — covers the SGR color codes our panels use).
+pub fn visible_len(s: &str) -> usize {
+    let mut len = 0;
+    let mut in_escape = false;
+    for c in s.chars() {
+        if in_escape {
+            if c.is_ascii_alphabetic() {
+                in_escape = false;
+            }
+        } else if c == '\x1b' {
+            in_escape = true;
+        } else {
+            len += 1;
+        }
+    }
+    len
+}
+
+/// Truncate `s` to at most `max` **visible** chars, preserving ANSI escapes
+/// (escape bytes don't count toward the budget). If truncation cuts the string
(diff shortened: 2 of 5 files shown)


aaronsb:
## Review — increment 5 (why-fired drill-down + micro-compositor)

Reviewed the compositor primitives, the `render_why` / `handle_why_key` drill-down, and the join-honesty path. Scoped out the known pre-existing `session_sim` and `compile.rs:334` items as instructed. Significant change (~797 add); focused on correctness, panic risk, and the load-bearing join-honesty invariant.

**Overall**: solid, well-tested, and the borrow-scoping in `render_why` is clean. One correctness/honesty edge worth resolving before this is the reference for increment 6, plus a few low-severity notes.

---

### 1. `build_why_index` — first-fire-wins aggregation vs all-fires spans (correctness + honesty edge)

**Location**: `rethink.rs`, `build_why_index` (~line 634) + `render_why_detail` (~line 676)

The index folds every fire of a `way_id` into one `WhyEntry` via `entry(...).or_insert_with(...)`. That means `trigger_channel`, `fire_score`, `criteria`, and `way_path` are taken from the **first-seen fire only**, while `matched_spans` is aggregated across **all** fires. Two consequences:

- **Detail is a session-wide fold, not the selected fire.** The drill-down is entered from a specific frame/epoch, but the right panel shows the way's aggregate. For a way that fires once these are identical; for a way that fires with a varying channel/score/span across turns, the panel can describe a different fire than the one the user focused. This is arguably consistent with the ADR-154 §1 "join by `way_id`, no epoch alignment" decision — but it's worth an explicit line in the panel or the doc that the detail is way-level, not fire-level, so it doesn't read as "why *this* fire fired."

- **Honesty gap in the mixed-channel case.** `render_why_detail` decides "semantic → no recoverable term" *only* when `matched_spans.is_empty()`. If a way's first fire is semantic (no span, so channel=`semantic…`, `fire_score=Some`) and a *later* fire on a keyword/file channel contributes a span, the entry ends up with `trigger_channel="semantic…"` **and** a non-empty `matched_spans`. The render then prints `Trigger semantic (score X)` alongside a quoted `"term"` under Matched — i.e. a term presented under a semantic label, which is exactly the fabrication the invariant forbids.

Right now honesty rests entirely on two upstream guarantees (the emitter never writes `matched_span` for a semantic event, and no way ever fires on more than one channel). The first is enforced; the second isn't, and the drill-down doesn't independently defend the invariant. Suggest making the drill-down self-honest so it can't be broken by an upstream change:

```rust
// build_why_index: don't let a semantic fire ever carry a span
if !fw.trigger_channel.starts_with("semantic") {
    if let Some(span) = fw.match_detail.as_ref().and_then(|m| m.matched_span.clone()) {
        if !e.matched_spans.contains(&span) { e.matched_spans.push(span); }
    }
}
```

or, more completely, key the entry by `(way_id, trigger_channel)` so channel/score/span always belong to the same fire. Either closes the mixed-channel path. A `why_tests` case with a semantic + keyword fire on the same `way_id` would pin it.

Please confirm whether a single way *can* fire on multiple channels across a session — if it structurally cannot, this drops to a defensive nicety; if it can, it's a real honesty hole.

---

### 2. Doc claims the compositor "owns the canonical" primitives, but the duplicates remain (low)

**Location**: `compositor.rs` module doc (~line 48) / ADR-154 §2

The module doc says it owns the canonical width primitives "that `render.rs` (`ansi_pad`/`ansi_visible_len`) and `rethink.rs` (`truncate_visible`) each grew their own copies of." Those copies are still present and un-migrated (`render.rs:514/524`, `rethink.rs:1256`). Nothing wrong for one increment, but the doc reads as if the consolidation already happened. Either migrate the callers (drops three functions, realizes the stated benefit) or soften the wording to "will replace" so the doc doesn't overstate current state. Cheap follow-up.

---

### 3. `read_way_body` frontmatter parsing edges (low)

**Location**: `rethink.rs`, `read_way_body` (~line 660)

- `trim_start_matches(['\n', '-'])` over-reaches into the body: a way body that legitimately opens with a markdown list (`- item`) or a horizontal rule (`---`) loses those leading dashes. Splitting on the closing-fence line rather than trimming a dash set would be safer.
- Empty frontmatter (`---\n---\nbody`) has no `\n---` after the prefix strip, so it falls through to returning the whole `content` (fences included). Edge case, unlikely in practice.

No panic risk here — `end + 4` stays within bounds on ASCII fences.

---

### 4. `session_id` byte-slice (nit, pre-existing convention)

`&player.session_id[..player.session_id.len().min(12)]` (~lines 819/666) panics if byte 12 lands mid-UTF-8. Session IDs are ASCII in practice and this matches the existing convention at `rethink.rs:219`, so it's not introduced here — a `char_indices` truncation helper would harden all three sites if you touch them later.

---

### What looks right

- `viewport` / `max_scroll` offset clamping is correct and directly unit-tested (mid, short-pad, past-end).
- `truncate_visible` style-sealing (append `\x1b[0m` only when truncated *and* the kept slice contains an escape) is correct, and `hjoin`'s `fit_visible` keeps columns aligned under ANSI and unequal panel heights — the alignment tests cover both.
- Panel width math in `render_why` is panic-free under a very narrow terminal: `left_w`/`right_w` floor via `clamp`/`saturating_sub`/`.max(...)`, and overflow past the terminal is clipped by `fit_to_terminal`.
- Borrow scoping is deliberate and correct: the `idx`/`frame` borrows are confined to an inner block and dropped before the `why_selected` / `why_detail_scroll` write-backs.
- Selection/scroll clamping is sound: `why_selected` re-clamped each render, `Down` guarded by `+1 < ways_len`, detail scroll clamped to `max_scroll`, empty-frame handled with placeholders. `frames` is guaranteed non-empty upstream, so the `frames.len() - 1` navigation is safe.
- Lazy index build is the right call and the honesty cases (semantic / keyword-with-span / no-span / no-record) are covered by `why_tests`.

Finding #1 is the one I'd resolve (or explicitly rule out) before increment 6 builds on this; the rest are follow-ups.

aaronsb:
Addressed the review in aabd790:

**Finding 1 (the real one) — fixed.** Confirmed empirically: multi-channel fires are common in the event log (`build → bash,keyword,semantic`, `documentation → bash,file,keyword,semantic:embedding:en`, …), so the first-fire-wins fold was a genuine honesty hole, not defensive. The why-index is now keyed by `(way_id, trigger_channel)`; the drill-down looks up the channel the focused frame shows for that way. Each facet is coherent — the semantic facet stays term-free (`fire_score` only), the keyword facet shows its own real span. Locked in by `multichannel_way_keeps_each_facet_honest` (semantic + keyword fires of one way → keyword span must NOT appear under the semantic trigger).

**Finding 2 — reworded.** Compositor doc no longer claims consolidation that didn't happen; `render.rs`/`rethink.rs` still hold their own width copies, migration noted as a follow-up.

**Finding 3 — fixed.** `read_way_body` is now line-based: a body opening with a markdown list or `---` rule keeps its leading dashes, and empty/unterminated frontmatter no longer drops content. New edge test.

**Finding 4 (session_id byte-slice nit)** — left as-is; pre-existing convention, ASCII-only session ids, and not introduced here.

All 217 unit + 11 compositor + 5 drill-down tests green; clippy clean on touched files.

PR #279, opened 2026-07-03
feat(introspect): live monitor + compaction windows + interactive UX (increment 6)

aaronsb:
## Summary

Increment 6 of the session-introspection arc (ADR-153/154): the **`ways introspect live`** monitor, plus the interactive/legibility refinements that came out of hands-on use. Built on the shared model + micro-compositor from PR #278.

## What's here (4 commits)

- **Live monitor** (`introspect live`) — poll + stat-gate over the current session, following the newest frame as ways fire (ADR-154 §3). Reuses the replay TUI; zero new deps.
- **Interactive Timeline** — ▲▼ move a highlighted selection (viewport follows), **Enter** opens that way's why-fired detail; frame nav preserves the selected way by `(id, epoch)` anchor (same way if present, nearest lesser epoch when rewound out). Speed moved to `+`/`-`.
- **Compaction windows** — a long session (kept across many compactions) no longer accumulates every way it ever fired into one 90-way wall. `build_frames` resets at each `session_start` boundary and restarts epoch/distance per window; the header shows `window W/M` and the latest window mirrors `ways list`.
- **Fixed-height layout** — the nav legend is pinned to the bottom and the list scrolls, at any terminal size (resizes reflow). Unified 4-row header + 2-row footer shared by both views, so the chrome doesn't jump when you Tab. Full session id, a window timestamp, and a full-width space-between-justified footer.
- **Dedup** — `render.rs` and `rethink.rs` now use `compositor`'s canonical ANSI-width primitives instead of their own copies (closes the PR #278 review note); one `header_lines`/`render_status_bar` shared by both views.

## Test plan

- `cargo test -p ways -p ways-core` — 223 unit tests green (incl. anchor navigation, compaction segmentation, fixed-height compose, justified footer, drill-down join honesty).
- `cargo clippy -p ways --all-targets -D warnings` — clean on all touched files (the one remaining, `settings/compile.rs:334`, is a pre-existing newer-clippy lint in untouched code).
- Exercised interactively against a live 11-window, 5-day session.

Note: `matched_span` in the drill-down stays blank on pre-enrichment fires (increment 3b isn't in the currently-deployed binary) — that lights up once this ships and a fresh session runs under it. This is the release that carries it.

### DIFF
--- a/tools/ways-cli/src/cmd/compositor.rs
+++ b/tools/ways-cli/src/cmd/compositor.rs
@@ -10,13 +10,12 @@
 //! need text selection or resizable/mouse panes, the ADR's escape hatch to ratatui
 //! applies; short of that, this stays.
 //!
-//! It provides ANSI-visible-width primitives ([`visible_len`], [`truncate_visible`],
-//! [`pad_visible`]) meant as the canonical home for that logic — `render.rs`
-//! (`ansi_pad`/`ansi_visible_len`) and `rethink.rs` (`truncate_visible`) still carry
-//! their own older copies, which can migrate here in a follow-up. Width is measured
-//! in `char`s, ignoring SGR escapes — the same grain those callers already use;
-//! East-Asian double-width and combining marks are not accounted for (that would
-//! need a new dependency the lean binary declines).
+//! It owns the canonical ANSI-visible-width primitives ([`visible_len`],
+//! [`truncate_visible`], [`pad_visible`]): `render.rs`'s row padding and
+//! `rethink.rs`'s `fit_to_terminal` both call these instead of the local copies
+//! they used to keep. Width is measured in `char`s, ignoring SGR escapes — the same
+//! grain those callers already use; East-Asian double-width and combining marks are
+//! not accounted for (that would need a new dependency the lean binary declines).
 
 use std::fmt::Write;
 
--- a/tools/ways-cli/src/cmd/introspect.rs
+++ b/tools/ways-cli/src/cmd/introspect.rs
@@ -24,6 +24,35 @@ pub fn replay(
     rethink::run(session, project, speed, false, all)
 }
 
+/// `ways introspect live` — monitor the current session's way firings, following
+/// the newest frame as ways fire. The "current" session is the most recent one in
+/// scope (the one actively writing events); `--session` overrides it. Scoping
+/// mirrors `replay`: defaults to the current project, `--project` for a specific
+/// one, and fails loud rather than silently globalizing when detection fails.
+pub fn live(session: Option<&str>, project: Option<&str>) -> Result<()> {
+    let content = ways_core::firing::load_events_text();
+    if content.trim().is_empty() {
+        println!("No events recorded yet.");
+        return Ok(());
+    }
+
+    // Live is inherently single-project (the session you're in), so no `--all`.
+    let scope = rethink::resolve_project_scope(project, false)?;
+
+    let session_id = match session {
+        Some(s) => s.to_string(),
+        None => match rethink_dump::most_recent_session(&content, scope.as_deref()) {
+            Some(s) => s,
+            None => {
+                println!("No sessions found in scope to monitor.");
+                return Ok(());
+            }
+        }
+    };
+
+    rethink::run_live(&session_id, scope.as_deref(), None)
+}
+
 /// `ways introspect list` — enumerate candidate sessions in scope, as a table or
 /// (`--json`) machine-listable data for an agent to pick from before dumping.
 pub fn list(project: Option<&str>, all: bool, json: bool) -> Result<()> {
--- a/tools/ways-cli/src/cmd/introspect.rs
+++ b/tools/ways-cli/src/cmd/introspect.rs
@@ -36,21 +36,33 @@ pub fn live(session: Option<&str>, project: Option<&str>) -> Result<()> {
         return Ok(());
     }
 
-    // Live is inherently single-project (the session you're in), so no `--all`.
-    let scope = rethink::resolve_project_scope(project, false)?;
-
-    let session_id = match session {
-        Some(s) => s.to_string(),
-        None => match rethink_dump::most_recent_session(&content, scope.as_deref()) {
+    // Resolve which session to monitor:
+    // - explicit `--session` wins;
+    // - `--project` scopes to the latest session recorded under that project;
+    // - otherwise the *most recently active* session anywhere — the one still
+    //   appending events. We deliberately don't scope the default to a detected
+    //   project: a long-running session's recorded project can be stale/wrong (the
+    //   boundary hook logs `CLAUDE_PROJECT_DIR:-$PWD`), and "live" means "what's
+    //   happening now," which is an activity signal, not a project one.
+    let session_id = match (session, project) {
+        (Some(s), _) => s.to_string(),
+        (None, Some(p)) => match rethink_dump::most_recent_session(&content, Some(p)) {
             Some(s) => s,
             None => {
-                println!("No sessions found in scope to monitor.");
+                println!("No sessions found for project {p}.");
                 return Ok(());
             }
-        }
+        },
+        (None, None) => match rethink_dump::most_recent_active_session(&content) {
+            Some(s) => s,
+            None => {
+                println!("No sessions found to monitor.");
+                return Ok(());
+            }
+        },
     };
 
-    rethink::run_live(&session_id, scope.as_deref(), None)
+    rethink::run_live(&session_id, project, None)
 }
 
 /// `ways introspect list` — enumerate candidate sessions in scope, as a table or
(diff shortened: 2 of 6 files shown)


aaronsb:
## Code Review — increment 6 (live monitor + compaction windows + interactive UX)

Reviewed the full diff plus surrounding code in `rethink.rs`, `compositor.rs`, `render.rs`, `introspect.rs`, `main.rs`. **No blocking issues.** The load-bearing invariants hold; findings below are all minor/cosmetic, ordered most-significant first.

### 1. `compose_screen` can drop the nav at the minimum terminal height (Minor, layout)
`rethink.rs` `compose_screen` (~L970–1010): the invariant "always exactly `drawable` lines, nav pinned last" breaks precisely at the smallest sizes.

With `top.len()==4` (header) and `nav.len()==2` (footer), at `drawable==6` (term_height 7): `rows_area = 6.saturating_sub(6) = 0`, then `.max(1) = 1`. That forces a row area of 1 that the budget doesn't actually have. When the list overflows, the rows region emits `view_h(1) + 1 scroll-hint = 2` lines — already exceeding the forced `rows_area` (the `while filled < rows_area` pad can't correct an over-fill). Total becomes `4 + 2 + 2 = 8`, and the final `lines.truncate(drawable)` silently eats the **nav** (it's last), so the legend vanishes at term_height 7–8.

Not a panic, and `render_frame`'s `drawable = …saturating_sub(1).max(6)` keeps it from going lower. But the stated invariant is "nav always pinned last," and this is the one regime it isn't. If you want it airtight, cap `rows_area` so `top + rows_area + nav <= drawable` before the `.max(1)`, or reserve nav/header first and give the list whatever's left (possibly zero). Low priority given how small the terminal has to be.

### 2. `render_why` isn't clamped to `drawable` — same nav-truncation on short terminals (Minor)
`render_why` (~L841, L939–944) assembles `header(4) + composited(body_h) + nav(2)` and relies on `body_h == term_height - 7` to sum to `drawable`. But `body_h = saturating_sub(7).max(3)`, so for term_height < 10 the `.max(3)` floor makes the total exceed `drawable`, and `fit_to_terminal`'s truncate drops the nav — same failure mode as #1, same tiny-terminal regime. Consistent and low-severity; flagging for symmetry with the compose_screen fix if you touch it.

### 3. Cross-window anchor reselection compares epochs from different numbering (Nit)
`reselect_by_anchor` (~L383) `rposition(epoch_fired <= anchor_epoch)` assumes a single ascending epoch axis. That invariant *is* held within a frame (all its ways belong to one window after the boundary reset, and `sort_by_key(epoch_fired)` orders them). But epoch restarts at 1 per window, so when navigating **backward across a compaction boundary**, the anchor's epoch (from window N) is compared against window N-1's epochs — apples to oranges. The id-match shortcut usually saves it; when it misses, the "nearest lesser epoch" fallback is arbitrary rather than meaningful. Consequence is only cursor placement, and the doc-comment already frames this as best-effort — fine to leave, worth a one-line caveat that the epoch fallback is intra-window.

### 4. `events_signature` spans all project logs (Nit, perf)
`events_signature` sums `events_log_sources()` across every project, but `refresh_live` re-parses only the monitored session. Activity in an *unrelated* project trips the stat-gate and forces a needless re-parse (+ `why_index` invalidation) of this session. Harmless and bounded by real event rate; only matters if a user live-monitors one project while another is very active.

### What's solid (verified, not just skimmed)
- **JOIN HONESTY intact.** `WhyIndex` is still keyed `(way_id, trigger_channel)` (`build_why_index` L693), and the drill-down's `facet` closure looks up `(w.id, w.trigger)` for the channel the frame shows. A keyword span still can't surface under a semantic trigger. `render_why_detail`'s empty-span branch keeps the semantic-vs-pre-enrichment distinction. No regression.
- **Boundary detection is correct.** `boundary = !frames.is_empty() && cluster has session_start` — the origin cluster is guarded from resetting; window/epoch numbering is right; a same-cluster `session_start + way_fired` resets then re-adds the fresh fires (correct scoping). The latest window genuinely mirrors `ways list`. `windows = max(frame.window)` is recomputed in both `run`/`run_live` and `refresh_live` over guaranteed-non-empty frames.
- **Panic surfaces closed.** `friendly_ts` uses `get(..16)` (safe on non-ASCII/short); the old `session_id[..12]` and `timestamp[..19]` byte-slices are gone; `jump_frame(usize::MAX)` clamps via `.min(last)`; `handle_timeline_key`'s down/enter guard `ways_len == 0`. `frames[current]` is always valid (clamped everywhere, frames non-empty by construction).
- **Fixed-height math** sums to exactly `drawable` in the normal (non-tiny) case — verified the overflow, extras-shown, and short-list branches each land on `drawable`; the `truncate` is a no-op except in regime #1.
- **Live loop**: never auto-advances (`playing && !player.live`), `End`/`G`/Space resume-follow only when `live`, `was_last`/`following` follow logic is sound, borrows scoped correctly.
- **Dedup**: `render.rs`/`rethink.rs` now delegate to `compositor`'s primitives; the removed copies were behaviorally equivalent. Good cleanup, closes the #278 note.

Didn't rebuild (trusting the stated 223-green run); the `session_sim` failures and `settings/compile.rs:334` lint are out of scope per the brief.

aaronsb:
## Release-gate review — 6 commits since the prior pass

Focused on the six new commits (`53285cc`, `9fa5ad1`, `a82109d`, `4675715`, `d72bf96`, `d11aa5a`). Built with `--features tui`; the six unit tests covering these changes all pass (`load_session_events_sorts_by_timestamp`, `redisclosure_repopulates_a_reset_window`, `compose_screen_is_fixed_height_with_nav_pinned`, `build_frames_segments_at_compaction_boundaries`, plus the two anchor tests). The only red suite is `session_sim` (pre-existing `/run/user` env issue on clean main — not from this PR).

### 1. `most_recent_active_session` + `live` rewrite (`53285cc`) — sound
- Selection is correct: scans all lines, skips unparseable/empty-session rows, keeps the max `ts` lexically. No unwrap panics (`ts` defaults to `""`, which can never win as "most recent"). Project-agnostic-by-activity is the right call and the rationale is well-documented in code.
- The `(session, project)` match arms are exhaustive and each has a clean no-session fallthrough. `--session` wins, `--project` scopes via `most_recent_session`, `(None,None)` uses the new activity-based selector.
- No collateral damage: `run_live` now receives the raw `project` Option instead of the resolved scope, but it only uses it as a display fallback for `project_name` (filtering is by `session_id`), so behavior is unchanged. Confirmed `dump`/`list`/the JSON path still call `most_recent_session` with scope — untouched.

### 2. `load_session_events` timestamp sort (`9fa5ad1`) — sound and genuinely needed
- Verified `build_frames` truly depends on stream order: the `ts_secs > last_ts_secs + 3` clustering (line ~1289) and the `!frames.is_empty() && session_start` boundary detection (line ~1304) both silently misbehave on unsorted input — exactly the stale-tail symptom described. The sort fixes the root cause.
- `sort_by(a.ts.cmp(b.ts))` is stable, so equal-`ts` events keep source-concat order — fine.

### 3. `j`/`k` drill-down scroll (`a82109d`) — sound
- Scroll is clamped every render: `why_detail_scroll.min(right.max_scroll(body_h))` then written back (render_why ~930-931). `max_scroll`/`viewport` (compositor.rs) are `saturating_sub`-based, so no panic and no persisting unbounded growth — even `G`'s `usize::MAX` sentinel clamps cleanly to the document bottom.
- `page = term_height.saturating_sub(9).max(1)` — safe, near-full-page, and clamped on render anyway.

### 4. Nav-clipping drain (`4675715` — my prior finding) — correctly fixed
- Both `render_why` and `compose_screen` guard the drain with `if lines.len() > drawable`, so `drain(0..lines.len() - drawable)` never underflows and the range is always valid. Dropping from the top preserves the nav legend. Tiny-terminal test asserts the footer survives.

### 5. Redisclosure repopulation + window reset (`d72bf96`) — sound
- The `.and_modify().or_insert_with()` is correct: after a boundary clears `active_ways`, a still-active way that only re-discloses now repopulates the window instead of no-op'ing into an empty "LIVE, but no ways". Interaction with the per-cluster `is_redisclosed = false` reset (then set true in the handler) is consistent for both the modify and insert paths.
- Checked the one thing that could have bitten this: the inserted `trigger: ev.trigger.clone()`. `way_redisclosed` events **do** carry `trigger` (show/mod.rs:158), so the `render_why` facet lookup `idx.get(&(id, trigger))` still matches for redisclosure-only windows. No detail-panel regression.

### Minor / non-blocking
- **Lexical timestamp ordering** (both the new sort and `most_recent_active_session`) assumes uniform RFC-3339 UTC (`Z`, consistent fractional precision). This holds for the current single-writer log, and it's the same latent assumption already in `most_recent_session`/`parse_ts_secs`, so no action needed — just the thing to remember if a source ever starts emitting zone offsets.
- `load_session_events`'s `l.contains(session_id)` line filter is a substring match (pre-existing, not from this PR).

### Verdict: SHIP
No correctness or panic risk in the six commits. Drain math, scroll clamps, and selection logic are all bounds-safe; the sort and redisclosure fixes address real regressions and are test-covered. Clear to cut the patch release. (rethink.rs line count / the planned split remain a known follow-up, no behavioral impact.)

PR #330, opened 2026-07-05
docs(adr-160): record first calibration findings (multi-topic share dilution)

aaronsb:
Records the first real calibration evidence for ADR-160's provisional operating points, now that the matcher actually runs (#328 fixed a deployment defect that had silently routed every scan to the single-vector fail-safe — the points had never been exercised against real late-interaction surfaces).

**Findings** (via the read-side instrument #327 and the `ways match` diagnostic #329):
- **Single-topic surfaces fire cleanly** — a focused prompt lands its owning way well above gate (testing → share 0.36/confirm 0.45; commits → 0.42/0.54).
- **Topic-diverse surfaces under-fire** — a 3-topic prompt (ADR+migration+PR) fires *nothing*: best match `documentation/adr` has peak 0.54 but share 0.13 < 0.15 gate. Cause is structural: `share = Σ(softmax mass)/n_chunks` caps a way owning one of N topics at ≈1/N. Real prompts are multi-topic, so this is the common case.

**Recorded, not decided:**
- Hypothesis: admit on a strong *peak* even when share is diluted (`share ≥ gate` OR `peak ≥ gate AND confirm clears`), letting the strict body-confirm carry precision.
- ADR-vs-code gap: stage 4 specifies routing summed mass through ADR-156's *calibrated* gate; the code thresholds a *hand-set* `SHARE_GATE=0.15`. Calibration must reconcile these.

Docs-only; feeds the provisional-points follow-up (#2). ADR lint clean.


aaronsb:
Folded into #331 — the calibration findings are now part of the adopted-co-gate ADR-160 update that lands with the code, rather than a separate docs-only 'findings in progress' note.

PR #331, opened 2026-07-05
feat(scan): peak co-gate + confirm 0.35 for late-interaction (ADR-160)

aaronsb:
**DRAFT — prototype for review, not for merge until operating points are approved (#2).**

## Problem it addresses

On a topic-diverse surface, `share = Σ(softmax mass) / n_chunks` caps a way that owns *one* of N topics at ≈1/N. A specific, decisive single-chunk match never clears the 0.15 share gate, so **nothing fires** — and real prompts are multi-topic, so this is the common case (see #330 / ADR-160 calibration findings).

## Change

Admit a way into confirmation on **either** a sufficient share **or** a decisive peak (`PEAK_GATE = 0.50`), letting the strict body-confirm carry precision for the peak-admitted case. Applied to both the live matcher (`run`) and the authoring diagnostic (`run_diagnostic`). Gates are env-overridable for sweeps without a rebuild: `WAYS_LI_{SHARE,PEAK,CONFIRM}_GATE`.

## Measured before/after (ADR + migration + PR prompt)

| Operating point | Result |
|---|---|
| before (share-only) | **0 fire** — `documentation/adr` peak 0.54 but share 0.13 < 0.15 |
| co-gate 0.50 + confirm 0.40 | `data/migrations` ✓; `documentation/adr` admitted but confirm **0.363 < 0.40** → misses |
| co-gate 0.50 + confirm **0.35** | `documentation/adr` ✓ **and** `data/migrations` ✓; unrelated refactor prompt stays **0 fire** (no new false positives on this sample) |

## Findings for calibration (#2)

1. The peak co-gate is **necessary** — recovers specific matches share dilutes.
2. It shifts the binding constraint to **CONFIRM**; 0.40 rejects a clear true positive, 0.35 recovers it without (on this sample) admitting false positives.
3. **Next step before adoption:** build a judged eval set (the `introspect fires` + labeled-prompt harness) and fit `PEAK_GATE`/`CONFIRM_GATE` against precision@recall, rather than eyeballing two prompts. Some legit prompts (a refactor prompt) still fire nothing — needs the wider set and/or prose work (#3).

`DiagRow` gains `admitted` so the diagnostic distinguishes `< gate` (neither gate) from `< confirm` (admitted, body didn't corroborate). 278 tests pass, clippy clean.


aaronsb:
Promoted from draft: prototype finalized to the shipped operating points (peak co-gate 0.50, confirm 0.35), calibration scaffolding (env overrides, debug traces) removed, and ADR-160 stage 4 + status updated to describe the adopted co-gate. Folds in and supersedes #330 (the docs-only findings PR). Verified: `documentation/adr` + `data/migrations` fire on the multi-topic prompt; unrelated refactor prompt stays at 0; 278 tests pass, clippy clean.

Open follow-up (not blocking): fit the two gates against a judged eval set, and reconcile the stage-4 ADR-156-calibrated-gate text with the hand-set SHARE_GATE.

PR #333, opened 2026-07-05
docs(adr): ADR-161 — queued mid-turn operator messages as an aggregated scan surface

aaronsb:
## What

Proposes **ADR-161**: a `PostToolUse` scan lane that matches the operator's mid-turn (queued) messages — the matcher's highest-value surface — by aggregating the burst of fragments into one late-interaction surface.

Status: **Proposed** (co-decided with the operator; left for review, not self-accepted).

## Why — discovered live, confirmed read-only

The ADR-160 late-interaction matcher was observed to leave the **operator's own message** unmatched exactly when the operator produces it most naturally. Two compounding causes, both confirmed by read-only diagnostics against a real session transcript (`56ebbbc1`):

1. **Mid-turn messages never reach the scan hook.** `UserPromptSubmit` is once-per-turn; a message typed while the agent works is queued (recorded as a `type: "queue-operation"` / `enqueue` transcript entry) and never becomes a `type: "user"` prompt, so `check-prompt.sh` never scans it. Confirmed vs. control: a *top-of-turn* "subagent" message keyword-fired `meta/subagents`, while every *mid-turn* queued message produced no fire. **The matcher path is correct** (`scan/mod.rs` → `late_interaction::run`) — only the delivery of this surface was missing.

2. **A lone fragment is too sparse for late interaction.** Operators type intent as a *burst of short messages*. A single queued fragment reports **"surface too sparse to chunk"** and falls to the single-vector fallback (ADR-160's <2-chunk fail-safe), where the relevant way (~0.4) doesn't clear the gate. **Concatenating** the burst makes it a ≥2-chunk surface that fires:
   - `documentation/mermaid` — peak 0.477 / share 0.190 / confirm 0.579 ✓
   - `softwaredev/visualization/diagrams` — 0.450 / 0.233 / 0.526 ✓
   - both admit on **share**, which one fragment can't accumulate.

So a fix that scanned queued messages *without* aggregation would still miss the operator's actual input. **The fix must do both.**

## Decision

`PostToolUse` lane (fires at the pause where queued messages flush, before the agent's next action, so a way can still *steer*):
1. Read transcript → select `queue-operation`/`enqueue` entries newer than a per-session **scan mark** (genuine operator text only).
2. **Aggregate** all-pending-since-mark into one surface (chosen for simplicity + reuse of existing flow; leans on the ADR-155 refire gate to absorb partial-burst re-fires).
3. Match + inject through the existing ADR-160 engine / ADR-130 reducer / ADR-155 lanes / `ways show way` gate.
4. Advance the mark.

New code is transcript-select + aggregate + mark — no matcher change. Alternatives (settle-then-scan, rolling window, Stop-hook, next-UserPromptSubmit) are recorded and rejected with rationale.

## Method note

Prototype-before-ADR (the chosen flow) is what surfaced the aggregation requirement — found with a read-only spike over the real transcript, **zero production code, no disruption to the live `introspect` session**. The spike script and the diagnostics are the evidence behind every number above.

## Not in this PR

Implementation (the hook + CLI transcript-scan) follows once the ADR is accepted. A companion matcher-flow design-note (two mermaid diagrams) is held separately.

https://claude.ai/code/session_01DNV2WVBmE5ALxjnQCCNGEP

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -80,7 +80,8 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-157](./system/ADR-157-case-insensitive-trigger-regex-compilation.md) | Case-insensitive trigger regex compilation | Accepted |
 | [ADR-158](./system/ADR-158-calibration-boundary-quality-hard-negatives-and-fire-breadth-ship-gate.md) | Calibration boundary quality — hard negatives and a fire-breadth ship gate | Accepted |
 | [ADR-159](./system/ADR-159-remove-ways-tune-curves-and-the-legacy-curve-cadence-field.md) | Remove ways tune-curves and the legacy curve: cadence field | Accepted |
-| [ADR-160](./system/ADR-160-chunked-late-interaction-matching-with-softmax-share-gating-for-way-selection.md) | Chunked late-interaction matching with softmax-share gating for way selection | Proposed |
+| [ADR-160](./system/ADR-160-chunked-late-interaction-matching-with-softmax-share-gating-for-way-selection.md) | Chunked late-interaction matching with softmax-share gating for way selection | Accepted |
+| [ADR-161](./system/ADR-161-queued-mid-turn-operator-messages-as-an-aggregated-scan-surface.md) | Queued mid-turn operator messages as an aggregated scan surface | Draft |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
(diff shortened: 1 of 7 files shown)


aaronsb:
**Implementation added to this branch** (commit `bc16c36`) — so merging #333 both accepts ADR-161 and ships it.

- `ways scan messages` + `check-queued.sh` (PostToolUse) + settings.json wiring.
- `collect_queued` pure + 5 unit tests; `scan_prompt_surface` shared by prompt/queued lanes (queued: no epoch bump, PostToolUse envelope).
- Per-session `queued-scan-mark` for dedup; harness envelopes filtered.

After merge: `ways update` picks up the new binary + hook + settings entry via reconcile, then the lane is live. To exercise it, the operator types a burst **while the agent is mid-task** (so the messages queue).

PR #334, opened 2026-07-05
fix(way): meta/workflows YAML + gate frontmatter lint in CI

aaronsb:
## Hotfix — a live regression I introduced in #332

My positive-prose rewrite of `meta/workflows` wrote:
> …deterministic multi-agent orchestration**: **fan-out, staged pipelines…

The unquoted **colon-space** parses as a YAML mapping value, so `serde_yaml` fails to load the frontmatter and the matcher **silently drops `meta/workflows` from the corpus entirely**. The way is currently absent from matching on `main`.

## Fix

Colon → parentheses. `ways lint` goes from **1 error → 0**, the frontmatter parses, and the way returns to the corpus on the next regeneration.

`description: …orchestration (fan-out, staged pipelines, verification and synthesis across many items) and how far up the substrate ladder a task warrants climbing`

## Why it slipped through (two process holes)

1. **My check was wrong.** In #332 I "verified lint clean" by grepping the output for positive-prose warnings — I never saw the ERROR line. Reading the summary would have caught it.
2. **`ways lint` is not a CI gate.** No `.github/workflows/` step runs it, so #332's green checks (build-only) never exercised the strict-YAML rule that exists precisely to catch a silently-dropped way.

## Follow-up (not in this PR)

Wire `ways lint --check` into CI so this class of bug — a frontmatter error that drops a way from matching with no other symptom — fails the build instead of merging.

https://claude.ai/code/session_01DNV2WVBmE5ALxjnQCCNGEP

### DIFF
--- a/.github/workflows/build-ways.yml
+++ b/.github/workflows/build-ways.yml
@@ -92,6 +92,14 @@ jobs:
         run: cargo test --manifest-path tools/Cargo.toml -p ways
       - name: Run simulation tests
         run: cargo test --manifest-path tools/Cargo.toml -p ways --test session_sim -- --test-threads=1
+      - name: Lint way frontmatter
+        # Invokes the ways binary directly against the repo corpus. `ways lint`
+        # resolves its schema from hooks/ways/frontmatter-schema.yaml, so this
+        # needs no ~/.claude projection and no make target — a frontmatter error
+        # that would silently drop a way from matching fails the build here.
+        run: |
+          cargo build --manifest-path tools/Cargo.toml -p ways
+          tools/target/debug/ways lint --check hooks/ways
 
   collect:
     needs: [build, test]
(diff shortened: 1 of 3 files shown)


aaronsb:
**Scope expanded per review** — now also closes the process hole that let the bug merge:

- **CI gate**: `ways lint --check hooks/ways` added to the build-ways `test` job. A frontmatter error that silently drops a way now fails the build.
- **`ways lint` made self-sufficient**: it resolves its schema from the corpus being linted (`<path>/frontmatter-schema.yaml`), falling back to `~/.claude` only for `--global`. No make target, no projection — `ways` stays standalone, and *make absent no longer breaks lint*.

Verified with an empty HOME: clean corpus → exit 0; re-introduced colon-space → exit 1 (the exact #332 failure). 295 unit tests pass.

PR #342, opened 2026-07-06
fix(update): skip rebuild on content-only pulls; remove refresh-claude-md fossil

aaronsb:
Two related cleanups from the ways-driven reframe.

## 1. `ways update` churn — gate the rebuild on what the pull touched

`ways update` rebuilt the entire suite after every pull. A docs/ways-only update (e.g. the `core.md` change in #341) therefore:
- downloaded a pre-built `ways` that lags source → discarded it under the ADR-150 downgrade guard → rebuilt from cargo (~16s),
- rebuilt `way-embed` from cmake (all of llama.cpp),
- re-downloaded `ways-audit` / `attend` / `attend-chat`,

…to produce binaries **identical** to the ones already installed. Since content lands far more often than a release is cut, this was the common path.

**Fix:** capture `HEAD` before/after the pull and classify the diff into two build groups:
- **cargo suite** — `tools/` except `tools/way-embed/` (`ways`, `ways-audit`, `attend`, `attend-chat` + shared crates)
- **way-embed** — `tools/way-embed/`
- root `Makefile` flags both (it drives both builds)

Each group refreshes **only if its own source moved**. A pull that touches neither skips straight to `corpus + reconcile`, so the pulled content (core.md, ways, skills, hooks) still projects. Falls back to a full refresh if git can't produce the diff (odd/detached state).

The classification is a pure function (`classify_build_groups`) with unit-test coverage; the freshness/downgrade-guard machinery is untouched.

## 2. Remove the `refresh-claude-md.sh` fossil

A pre-injection-era artifact: it only `touch`ed `~/.claude/CLAUDE.md` *if it existed* (a no-op on real installs) and was wired to no hook event. Under the current model — core.md injected at SessionStart is the base guidance, `~/.claude/CLAUDE.md` is left entirely to the user — it has no job. Removed the script, its `PROJECTED_FILES` manifest entry, and the legacy `sync-to-home` projection/test references.

## Not included (release-process note)

Point 1 of the report — "we've not published the latest binary" — is a release-cadence issue, not a code bug: the pre-builts on Releases are `ways-v1.3.0` while source is 10+ commits ahead, so binary-affecting updates still download-then-discard until a release is tagged. This PR eliminates the churn for *content-only* updates; cutting a release closes the remaining case.

## Tests
- `cargo test -p ways` — 306 + 7 + 11 + 15 pass, incl. new `classify_build_groups_routes_changed_paths`
- `scripts/sync-to-home-test.sh` — ALL PASS


aaronsb:
## Code review addressed (commit dc71378)

A `code-reviewer` pass found one real hole in the classifier and a few lower items. Fixes:

- **#1 (MEDIUM) — embedded asset outside `tools/`.** `ways` embeds `hooks/memory-seed/seed-v1.md` via `include_str!`, so an in-place edit to it changes the binary but the path isn't under `tools/` — it would take the content-only fast path and skip the rebuild, shipping a stale compiled-in seed. Fixed: added `EMBEDDED_ASSET_PREFIXES` (`hooks/memory-seed/`) to the cargo predicate, a coupling note at the `include_str!` site, and a **durability test** that walks the source for `include_str!`/`include_bytes!` literals escaping `tools/` and fails if any isn't listed. I verified this is the only such escape in the tree.
- **#2 (LOW/MED) — committed-HEAD blind spot.** The gate diffed committed HEADs only; uncommitted binary-source edits were invisible. Now folds `git diff --name-only HEAD` through the same classifier.
- **#3 (LOW) — relink self-heal.** The content-only fast path returned before `make relink`; now runs it there too (idempotent).

### Accepted, not fixed
- **#4 — dangling symlink.** On existing *symlink-mode* installs, removing `refresh-claude-md.sh` from the manifest leaves `~/.claude/hooks/refresh-claude-md.sh` as a dangling symlink after update — symlink-mode `reconcile` has no orphan prune (copy/legacy `sync-to-home` self-cleans via its manifest prune). It's cosmetic dead cruft, not a functional break (`settings.json` never referenced it). Noting it here so a stray dangling link in `~/.claude/hooks/` is understood as expected.
- **#5 — cosmetic message** when only way-embed changed. Left as-is.

Tests: `cargo test -p ways` 307+7+11+15 pass (incl. the new guards); clippy clean.

PR #348, opened 2026-07-07
ADR-162: mechanical session-link suppression + deny hook

aaronsb:
## Summary

ADR + working hook that stops the `Claude-Session` transcript link from landing in
commits and PR/issue bodies. Motivated by the concern that the link resolves to the
**full session transcript** — a thin wall in front of accidental secret disclosure,
especially on public repos. **Status: ADR-162 Accepted (ratified).**

## Why the old control didn't hold

A prior fix set `attribution.commit/pr=""` and added a report-only macro block.
That never governed the session link:

- `attribution.commit/pr` only suppress the Co-Authored-By / "Generated with"
  footers — never the session link.
- The real key, `attribution.sessionUrl`, is **undocumented** and defeated by
  upstream bug #18253 (attribution settings not reliably injected into model
  context; marked *not planned*). The system-prompt instruction survives and the
  model follows it.
- Claude Code has **no mechanical strip** — prevention is purely "don't instruct,"
  which the bug defeats.

Observed leak: a PR in an unrelated repo with **no local config** still got the
link, inheriting global `attribution=""`.

## What this does

- **ADR-162** (Accepted) — records the decision + rejected alternatives. Scoped to
  the **hook mechanism**; distribution across hosts is ADR-163's concern (paired
  work).
- **`hooks/ways/strip-session-link-pre.sh`** — PreToolUse Bash hook that **denies**
  `git commit` / `gh pr|issue create|edit|comment` when the command carries a
  `Claude-Session:` trailer or bare `session_` URL, instructing a clean re-issue
  that then passes through normal permission evaluation.
- **settings.json** — wires the hook into the reconcile-owned Bash PreToolUse block
  (beside `check-bash-pre.sh`), not the fragment store (which skips hooks slices).

## Key decision: deny, not silent-rewrite

Silent-rewrite via PreToolUse `updatedInput` was the initial choice, but review
found `updatedInput` is honored **only** with `permissionDecision: allow` (no
pass-through — upstream FR #381), which would **bypass the operator's own
Deny/Allow/Ask gates** on outward-facing `gh pr/issue create`. Deny bypasses
nothing; cost is one retry, not a lost commit. Re-ratified to deny by @aaronsb;
review lead concurred.

## Known gap

`--body-file`/`-F <path>` bodies live in a file the hook can't inspect from the
command string — a documented coverage gap (inline/heredoc, the model's default
format, is covered).

## Not done here

- Not merged; `reconcile` not run — no change to live `~/.claude`.

Paired with ADR-163 (config distribution / dotfiles layering), which references this
hook as the distributed primitive.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -81,7 +81,8 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-158](./system/ADR-158-calibration-boundary-quality-hard-negatives-and-fire-breadth-ship-gate.md) | Calibration boundary quality — hard negatives and a fire-breadth ship gate | Accepted |
 | [ADR-159](./system/ADR-159-remove-ways-tune-curves-and-the-legacy-curve-cadence-field.md) | Remove ways tune-curves and the legacy curve: cadence field | Accepted |
 | [ADR-160](./system/ADR-160-chunked-late-interaction-matching-with-softmax-share-gating-for-way-selection.md) | Chunked late-interaction matching with softmax-share gating for way selection | Accepted |
-| [ADR-161](./system/ADR-161-queued-mid-turn-operator-messages-as-an-aggregated-scan-surface.md) | Queued mid-turn operator messages as an aggregated scan surface | Draft |
+| [ADR-161](./system/ADR-161-queued-mid-turn-operator-messages-as-an-aggregated-scan-surface.md) | Queued mid-turn operator messages as an aggregated scan surface | Proposed |
+| [ADR-162](./system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md) | Mechanical session-link suppression as defense against transcript disclosure | Draft |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -82,7 +82,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-159](./system/ADR-159-remove-ways-tune-curves-and-the-legacy-curve-cadence-field.md) | Remove ways tune-curves and the legacy curve: cadence field | Accepted |
 | [ADR-160](./system/ADR-160-chunked-late-interaction-matching-with-softmax-share-gating-for-way-selection.md) | Chunked late-interaction matching with softmax-share gating for way selection | Accepted |
 | [ADR-161](./system/ADR-161-queued-mid-turn-operator-messages-as-an-aggregated-scan-surface.md) | Queued mid-turn operator messages as an aggregated scan surface | Proposed |
-| [ADR-162](./system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md) | Mechanical session-link suppression as defense against transcript disclosure | Draft |
+| [ADR-162](./system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md) | Mechanical session-link suppression as defense against transcript disclosure | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md
+++ b/docs/architecture/system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md
@@ -1,0 +1,129 @@
+---
+status: Draft
+date: 2026-07-06
+deciders:
+  - aaronsb
+  - claude
+related:
+  - 152
+  - 163
+---
+
+# ADR-162: Mechanical session-link suppression as defense against transcript disclosure
+
+## Context
+
+Claude Code appends a session link to git commit messages (`Claude-Session:
+https://claude.ai/code/session_…`) and to PR bodies. That link resolves to the
+**full session transcript**. A transcript routinely contains material that
+scrolled through the conversation — file contents, environment values, tokens,
+internal paths. Publishing the link therefore places a single click between a
+public commit and whatever secret happened to pass through the session. On a
+public repository this is an accidental-disclosure surface, not a convenience.
+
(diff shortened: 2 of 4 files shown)


aaronsb:
## Consolidated review — reconciled with the deny flip (review lead: Hana-alpha)

Context: Aaron re-ratified **deny**. Findings re-framed for that design.

### ✅ Eliminated by the deny flip
- **[was CRITICAL] Quote/paren corruption.** The `\S*`/`\S+` regexes matched shell/markdown terminators, so when the session URL was the last token before a closing `"` or `)` the rewrite ate the terminator and emitted a **syntactically broken command that Claude Code would execute** — reproduced across every inline `-m`/`--body` form; only the heredoc shape the PR tested survived. This lived entirely in the `updatedInput` rewrite path. **Deny removes that path, so this evaporates** — and it's strong extra justification for deny: silent-strip wasn't only a permission-bypass risk, it actively corrupted commands.

### 🔧 Still must-fix under deny
- **[MEDIUM] Detection-gate evasion.** `*"git commit"*` requires adjacency, so `git -C <path> commit` and `git --git-dir … commit` slip past and the link leaks. Loosen to token presence (`*git* && *commit*`) or a regex; same for `gh … pr/issue` behind global flags.
- **[MEDIUM] Detection false-positives → deny-loop risk.** A bare `https://…/session_…` mentioned in legitimate prose triggers — and in *this* repo, commits/ADRs/design-notes discuss `Claude-Session:` as subject matter. Under deny that blocks a legitimate commit, and if the URL is genuinely wanted the model can't remove it → **retry loop**. Tighten detection to trailer/footer context; document that inline session-URL mentions are always blocked.

### 📄 Doc / traceability
- **ADR "guarantee" overstates.** Control is best-effort: fails open on parse error, and doesn't cover `--body-file`/`-F <path>`, the `$EDITOR` path, or (per above) `git -C`. Soften to "blocks on the covered command forms" and list the editor + body-file gaps in Consequences.
- **ADR claims a change the PR doesn't make.** ADR-162 says it "also sets `attribution.sessionUrl:false`", but `settings.json` still reads `{commit:"",pr:""}`. Add the key or reword.
- **Status mismatch.** Frontmatter/INDEX mark 162 Accepted while the PR title says Draft/asks to ratify — reconcile (Aaron ratified deny, so Accepted is fine, but drop the "ratify?" framing).
- **[VERIFY] Multi-hook composition.** `strip-session-link-pre.sh` sits beside `check-bash-pre.sh` in the same Bash PreToolUse matcher — confirm via live smoke test that a **deny** from a non-first hook isn't ignored/overridden.

### 👍 Solid
Fail-open posture, no-op guard, output envelope, and `tool_input` preservation are all correct; ADR scoping + cross-ref to 163 are right.

**Net:** #1 gone with deny; the gate-evasion and false-positive/deny-loop items are the real must-fixes; the rest is doc cleanup. Land deny + those two and I'll re-review, then we merge.

aaronsb:
## Review remediation (Hana-beta) — pushed in 7c513fe

Thanks for the thorough pass. Per-finding:

**🔧 Must-fix**
- **Gate evasion** — fixed. Command-type detection now matches on token presence, newline-collapsed: `\bgit\b.*\bcommit\b` and `\bgh\b.*\b(pr|issue)\b`. Verified deny on `git -C <path> commit`, `git --git-dir=… commit`, and `gh --repo o/r pr create`.
- **False-positive / deny-loop** — fixed. Detection now requires the link in **trailer/footer position**: line-leading, optionally behind the `Claude-Session:` key, tolerating trailing shell terminators (closing `"` / `)`). A URL *inline in prose* passes. Verified: `git commit -m "docs: the Claude-Session: trailer format"`, `git commit -m "see <url> for repro"`, and `gh pr comment … "fixed in <url> yay"` all ALLOW; real trailer/footer forms DENY.

**📄 Doc / traceability**
- **"Guarantee" overstated** — softened to best-effort; Consequences now enumerates residual gaps: `--body-file`/`-F`, the bare-`git commit` `$EDITOR` path, and message-carrying siblings (`git tag -m`, `gh release/gist create`).
- **`sessionUrl:false` mismatch** — resolved by *rewording*: the ADR no longer claims this change sets it (it's undocumented + defeated by #18253; coupling risks settings validation for no working benefit). Hook is the load-bearing control. `settings.json` unchanged.
- **Status mismatch** — PR title no longer says Draft; frontmatter/INDEX are Accepted (Aaron ratified deny).
- **[VERIFY] Multi-hook composition** — confirmed via docs: all matching PreToolUse hooks run and the most-restrictive result wins (`deny > defer > ask > allow`), order-independent. My second-position deny is guaranteed; no reordering. Recorded in the ADR Neutral section.

The quote/paren corruption you flagged as CRITICAL is gone with the rewrite path (deny doesn't build a command). Re-review when you can.

aaronsb:
## Re-review: APPROVED ✅ (review lead: Hana-alpha)

Verified each finding against `7c513fe`, not just the summary:

- **#1 corruption** — gone; deny builds no command, no `\S` rewrite.
- **#2 gate evasion** — closed; `\bgit\b.*\bcommit\b` newline-collapsed catches `git -C`/`--git-dir`/`gh <flags> pr`.
- **#3 false-positive / deny-loop** — closed; gate (2) anchors the link to line-leading (trailer/footer) position, so inline prose mentions pass. Correctly catches both the `Claude-Session:` trailer *and* a bare session URL alone on a line.
- **#4/#5/#6 docs** — guarantee softened + gaps enumerated; `sessionUrl` claim reworded (settings.json carries only the hooks entry); status reconciled to Accepted; composition confirmed most-restrictive-wins, order-independent.

Fail posture is right for a deny control: no-op → allow, link detected → fail **closed** (exit 2).

**One optional, non-blocking note** (take it or leave it): the enumerated-gaps list could add the multi-`-m` trailer form — `git commit -m "subj" -m "Claude-Session: <url>"` puts the trailer non-line-leading, so gate (2) doesn't fire. It's an unlikely vector (the harness emits the heredoc form) and the control is explicitly best-effort, so this is a one-line honesty addition to the gap list, not a fix. Doesn't block merge.

Clear to merge on green. Per our sequence: **#348 merges first**, then I align + merge #349 (ADR-163).

PR #350, opened 2026-07-07
fix(context): resolve transcript by session id, not just cwd

aaronsb:
## Bug
`ways context --json` — the gauge the **wrap** and **context-status** skills read in Step 0 — returns `Error: No active transcript found for project` whenever the agent's shell cwd has drifted from the project directory. Observed in the wild: a `/wrap` session reporting `ways context unavailable` and the agent having to guess the fill level.

## Root cause
`get_context` located the transcript **only** by deriving a project slug from cwd / `CLAUDE_PROJECT_DIR` + newest-transcript-in-that-project. But `CLAUDE_PROJECT_DIR` is **unset** in tool subprocesses, so it falls back to cwd — and any cwd ≠ project (or a path/slug mismatch) fails hard. The module already had a cwd-independent `find_transcript_by_session`, but the CLI path never fed it a session id.

## Fix
Prefer **`CLAUDE_CODE_SESSION_ID`** — which Claude Code sets in every tool environment — and resolve the transcript by id across all project dirs. Precedence (factored into a pure, unit-tested `resolve_transcript`):
1. explicit `session_id` (e.g. `get_context_for_session`)
2. `CLAUDE_CODE_SESSION_ID` from the env, when no explicit `--project` — **cwd-independent** (the fix)
3. cwd / `CLAUDE_PROJECT_DIR` slug + newest transcript — the original heuristic, **preserved as fallback**

An explicit `--project` still wins over the env session id.

## Verification
- New hermetic unit tests (temp projects root, no env mutation): env-id resolves regardless of cwd; explicit id > env; explicit `--project` ignores env id; empty env id falls through.
- Full `-p ways` suite green (311 tests).
- End-to-end with the rebuilt binary:
  - old binary from `/tmp` → `Error: No active transcript found`
  - new binary from `/tmp` and `~` → correct session + token counts
  - new binary, env unset, from project dir → fallback still works

### DIFF
--- a/tools/ways-cli/src/cmd/context.rs
+++ b/tools/ways-cli/src/cmd/context.rs
@@ -423,11 +423,35 @@ mod tests {
     }
 
     #[test]
-    fn empty_env_session_id_falls_through() {
-        // An empty CLAUDE_CODE_SESSION_ID must not short-circuit; it falls
-        // through to the project heuristic (here: no match -> error).
+    fn find_transcript_by_session_in_returns_none_when_missing() {
+        // Pins the not-found half of the id lookup the fall-through depends on.
+        let root = temp_projects_root(line!(), &[("-someproj", "present-sid")]);
+        assert!(find_transcript_by_session_in(&root, "absent-sid").is_none());
+        assert!(find_transcript_by_session_in(&root, "present-sid").is_some());
+        std::fs::remove_dir_all(&root).ok();
+    }
+
+    #[test]
+    fn env_session_id_with_missing_transcript_falls_through() {
+        // project_dir=None reaches the env branch; a non-empty env id whose
+        // transcript doesn't exist must fall through to the project heuristic,
+        // which errors here because the empty projects root has no match. (An
+        // empty root guarantees the fallback errors regardless of the ambient
+        // cwd/CLAUDE_PROJECT_DIR the heuristic reads.)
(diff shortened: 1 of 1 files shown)


aaronsb:
## Code review — verdict: solid, no merge-blockers (addressed)

A `code-reviewer` pass confirmed: precedence logic correct, all existing callers (`get_context`, `get_context_for_session`, `ways list`, `show`, `lookbehind`) regression-safe, path/slug construction equivalent to the old form, and the new code clippy-clean (CI clippy job passed).

**One real finding (test fidelity, non-blocking) — fixed in `de0dabe`:**
The original `empty_env_session_id_falls_through` passed `project_dir=Some(...)`, so it never reached the `project_dir.is_none()` env branch it named; the empty-string guard and the "env id set but transcript missing → fall through" path were uncovered. Replaced with three tests that actually traverse those branches (missing-transcript fall-through, empty-env non-short-circuit, and a direct `find_transcript_by_session_in` not-found assertion), each using an empty temp projects root so the heuristic fallback errors deterministically.

Informational (not addressed here, correctly out of scope): a pre-existing `unnecessary_get_then_check` in `settings/compile.rs:334` that a future CI toolchain bump will surface repo-wide — belongs in its own targeted PR per CONTRIBUTING's toolchain-drift policy.

PR #360, opened 2026-07-13
fix(context): one source of truth for model context-window resolution (ADR-166)

aaronsb:
## The bug

`ways context` reported `tokens_total: 200000` and `pct_used: 106` for a `claude-fable-5` session that was running past 200k with no forced compaction. Fable 5 has a **1M** window.

The context window turned out to be answered in **three places**, by three different substring heuristics that disagreed with each other — plus a fourth hardcoded `unwrap_or(200_000)`:

| Site | Rule |
|---|---|
| `cmd/context.rs:239` | `opus-4` → 1M; `sonnet`\|`haiku` → 200K; else env, else 200K |
| `session.rs:322` | `opus-4` → 1M; else env, else 200K |
| `sensor-peers/src/lib.rs:567` | `[1m]`\|`opus-4`\|`sonnet-4` → 1M; else 200K |

All three had gone stale. `claude-fable-5` matched no branch in any of them. `claude-sonnet-5` was misclassified the same way. `sonnet-4` resolved to 1M in one and 200K in another — both cannot be right.

This is **not** just a cosmetic gauge bug. ADR-126 scales way refire half-lives as a *fraction of the window*, so on Fable the disclosure curve was compressed 5x and every way re-fired five times more often than authored. The `context-threshold` way trigger divided by the same wrong denominator.

Two structural defects underneath the staleness:

- **`sensor-peers` tested for a `[1m]` suffix Claude Code never writes.** Across ~78,000 `message.model` records in local transcript history the field is always the bare id. Dead code — detection cannot observe the harness's 1M toggle at all, only the model id.
- **`CLAUDE_CONTEXT_WINDOW` was read only from the fallback arm** of two resolvers, so the documented escape hatch silently did nothing on every model they matched.

## The fix

`ways_core::context_window` (ADR-166) is now the one resolver. Every consumer calls it.

**Resolution order:** `CLAUDE_CONTEXT_WINDOW` → explicit model table → conservative 200K default.

**The result carries its provenance.** `ways context --json` now emits `window_source` (`env_override` / `model_table` / `default`), so a defaulted window is *reported* as a default instead of being indistinguishable from a detection. This is what keeps the next stale-table failure observable — the Fable session would have read `window_source: "default"` at the moment it was wrong.

Rewired all five consumers: the gauge, `pct_used_from_transcript` (the `context-threshold` trigger), `session.rs` (ADR-126 fire evaluation), `cmd/show/mod.rs`, and `sensor-peers`.

## Review findings, fixed in the second commit

- **`<synthetic>` turns resolved to the default.** Claude Code writes `"model": "<synthetic>"` for interrupts and API errors, and **9 transcripts in local history end on one** — so a live 1M session whose newest turn was an interrupt got a 200K window. Sentinels are now absences: the scan skips them and walks back to the last real model.
- **Provider-qualified ids regressed to the default.** Ids were anchored at byte 0, so Bedrock/Vertex forms (`us.anthropic.claude-opus-4-8-v1:0`) matched nothing, where the old `contains("opus-4")` resolved them to 1M. Ids are now matched as a *boundary-delimited component*; the boundary requirement still rejects `claude-sonnet-55` as `claude-sonnet-5`. Bare aliases match on exact equality only.
- **`sensor-peers` leaked the operator's override onto peers** (introduced by this PR). A peer's window belongs to the peer's session — an operator with `CLAUDE_CONTEXT_WINDOW=1000000` would see a Haiku peer at 190K/200K, genuinely about to compact, rendered as 19% full. Foreign sessions now resolve through the model table alone.

One reported finding was a **false positive**: the review claimed `[1m]` appears in 509 transcript model records. It appears 414 times as *prose inside message content* — it's in the system prompt's environment block — and **zero** times as a `message.model` value. Component matching tolerates the marker anyway.

## Deliberate behavior change

Models absent from the table now default where the old `opus-4` substring gave them 1M (`claude-opus-4-5`, `claude-opus-4-1`, `claude-opus-4-0`). The 1M window arrived with the 4.6 generation, so calling Opus 4.1 a 1M model was exactly the over-broad match being removed. None appear in local transcript history; they can be pinned explicitly once their true windows are confirmed.

## Verification

Driven end-to-end, not just unit-tested:

- The Fable session that read **200k / 106%** now reads **1M / 23%**, `window_source: model_table`.
- A transcript ending on `<synthetic>` now resolves to `claude-opus-4-8` / 1M / `model_table` (previously 200K / `default`).
- `CLAUDE_CONTEXT_WINDOW=200000` correctly overrides to `env_override` — on every model, as documented.
- 14 resolver unit tests pinning every model id that actually appears in transcript history, plus sentinels, qualified ids, alias exactness, override precedence, and the zero-window guard. Full workspace suite green.


aaronsb:
## High-effort review run + remediation

Ran a workflow review on this branch (after merging main in to resolve the stale INDEX conflict). Fixed the two **CONFIRMED correctness** findings — both produced ~5x-wrong windows:

- **env-override leak on foreign sessions.** `context_window_from_transcript` used the env-honoring `resolve`, but all its callers (`detect_context_window_for` → rethink / introspect / rethink_dump) replay a *recorded* session, not the live one — so an operator's `CLAUDE_CONTEXT_WINDOW` rescaled foreign sessions' gauges and ADR-126 refire thresholds. Switched to `resolve_for_foreign_session`. The live gauge (`cmd::context`) still honors the override (correct there). This also makes the session tests env-independent — dropped the guard, added a `<synthetic>`-walk-back regression.
- **sensor-peers not given the sentinel walk-back.** It overwrote `model` with the newest assistant line including `<synthetic>`, so a peer whose newest turn is an interrupt resolved to the 200K default and its fill was grossly overstated. Now gates on `!is_sentinel`, mirroring `model_from_transcript`.

Full suite green (832). Pushed.

## Left for your call (author judgment, not remediated)

- **`claude-sonnet-4-5` → default (was 1M via the old `contains("sonnet-4")`).** Your own test `aliases_match_exactly_and_do_not_prefix_match` asserts this is intentional (unlisted → visible default). Flagging only so you can confirm sonnet-4-5's real window and add it to the table if it's genuinely 1M.
- **`contains_component` + a future point-release** (e.g. a hypothetical `claude-sonnet-5-2`) would borrow `claude-sonnet-5`'s window and report `ModelTable`. Speculative (no such id ships today); your matching design choice.
- **Cleanups:** `context.rs` scans for the model twice (`detect_model` then again inside `resolve_window`); `window_for_model` allocates+sorts the 9-entry table each call on the per-turn hook path. Both low-impact — left for you to fold in or wave off.

PR #366, opened 2026-07-17
ADR-167: sessionUrl governs the session-link trailer (supersedes ADR-162)

aaronsb:
## Summary

- **ADR-162's central claim was false when written.** It concluded no setting could govern the `Claude-Session` trailer, so a PreToolUse deny hook had to be the sole mechanical defense. It cited `#18253` as "marked not planned" — but that's the Co-Authored-By/footer bug [CLOSED/COMPLETED], not the session link. The issue it meant is `#41873` [CLOSED/NOT_PLANNED], which upstream then superseded by shipping `attribution.sessionUrl` in **v2.1.183 (2026-06-19)** — seventeen days *before* ADR-162 was dated.
- **ADR-167** records the corrected decision: `attribution.sessionUrl: false` is the primary control; the ADR-162 hook is **retained as a backstop** (the key is undocumented upstream `#69614`, defaults to ON, and is scope-qualified to "web and Remote Control sessions"). Hook code is untouched — only its header comment, which repeated the same mis-citation.
- **ADR-162** marked `Superseded`, keeping its threat model and hook design intact per the ADR-104 precedent.

## Evidence

Single-variable experiment — same repo, same binary (v2.1.212), same account:

| `attribution.sessionUrl` | Trailer instruction in system prompt |
|---|---|
| `true` (control) | **present** |
| `false` (treatment) | **absent** |

Two earlier attempts were discarded as void and are documented as such in the ADR: a headless `claude -p` pair (`-p` never injects the trailer, so the control couldn't reproduce the baseline) and a run in a non-git directory (no repo, no commit instructions). Only the controlled pair is evidence.

## Why this persisted

One mis-citation propagated into four artifacts, each making the next look corroborated: ADR-162 → the hook's header comment → the operator's memory entry → the `delivery/github` macro, which rendered the wrong fact as a reassuring `Session-link trailer: OFF` in every session — including sessions whose system prompt carried the trailer.

The generalizable lesson, recorded in ADR-167: ADR-162 froze a **contingent upstream fact** as though it were a durable architectural truth. Upstream behavior is evidence with a shelf life.

## Follow-ups (tracked, not in this PR)

- Fix the `delivery/github` macro to read `sessionUrl` (it reports on `commit`/`pr` alone)
- `ways settings lint` doesn't validate nested sub-keys — `permissions.alow` lints clean and silently drops your rules
- Upstream PR to SchemaStore adding `attribution.sessionUrl` (declares `additionalProperties: false` today, so it marks a real key invalid)

## Test plan

- `docs/scripts/adr lint` — 0 errors (1 pre-existing warning on ADR-127, unrelated)
- `docs/scripts/adr index -y` — INDEX.md regenerated; ADR-162 shows `Superseded`
- Hook change is comment-only; no behavior change

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -82,10 +82,11 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-159](./system/ADR-159-remove-ways-tune-curves-and-the-legacy-curve-cadence-field.md) | Remove ways tune-curves and the legacy curve: cadence field | Accepted |
 | [ADR-160](./system/ADR-160-chunked-late-interaction-matching-with-softmax-share-gating-for-way-selection.md) | Chunked late-interaction matching with softmax-share gating for way selection | Accepted |
 | [ADR-161](./system/ADR-161-queued-mid-turn-operator-messages-as-an-aggregated-scan-surface.md) | Queued mid-turn operator messages as an aggregated scan surface | Proposed |
-| [ADR-162](./system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md) | Mechanical session-link suppression as defense against transcript disclosure | Accepted |
+| [ADR-162](./system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md) | Mechanical session-link suppression as defense against transcript disclosure | Superseded |
 | [ADR-163](./system/ADR-163-config-separation-dotfiles-source-of-truth.md) | Config separation — dotfiles as source-of-truth feeding the settings fragment store | Accepted |
 | [ADR-164](./system/ADR-164-file-artifacts-distributed-across-hosts-must-be-carried-by-value-not-host-absolute-reference.md) | File artifacts distributed across hosts must be carried by value not host-absolute reference | Accepted |
 | [ADR-165](./system/ADR-165-loop-control-bookends-start-develop-merge-release-wrap.md) | Loop-control bookends: start, develop, merge, release, wrap | Accepted |
+| [ADR-167](./system/ADR-167-session-link-suppression-attribution-sessionurl-as-primary-control-deny-hook-as-backstop.md) | Session-link suppression: attribution.sessionUrl as primary control, deny hook as backstop | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md
+++ b/docs/architecture/system/ADR-162-mechanical-session-link-suppression-as-defense-against-transcript-disclosure.md
@@ -1,15 +1,39 @@
 ---
-status: Accepted
+status: Superseded
 date: 2026-07-06
+revised: 2026-07-16
 deciders:
   - aaronsb
   - claude
 related:
   - 152
   - 163
+  - 167
+superseded_by: ADR-167
 ---
 
 # ADR-162: Mechanical session-link suppression as defense against transcript disclosure
+
+## Status: Superseded by ADR-167
+
+**Superseded 2026-07-16.** The threat model below is correct and load-bearing: the
+session link resolves to the full transcript, and publishing it on a public repo is an
+accidental-disclosure surface. The **hook** this ADR designed also survives unchanged —
+deny rather than rewrite, at user scope, trailer/footer position only. Its *rationale*
+does not.
+
+This ADR's central claim — that no setting can govern the link, so suppression must be
+mechanical — was **false when written**. It rests on a mis-citation: `#18253` is the
+Co-Authored-By/footer bug ([CLOSED/COMPLETED]), not the session link; the issue meant
+was `#41873` ([CLOSED/NOT_PLANNED]). And `#41873` had already been superseded by
+`attribution.sessionUrl`, shipped in **v2.1.183 (2026-06-19)** — seventeen days before
+this ADR was dated. A controlled experiment on 2026-07-16 (same repo, same v2.1.212,
+only the key differing) confirms the setting governs the trailer.
+
+[[ADR-167]] carries the corrected decision: `attribution.sessionUrl: false` is the
+primary control, and the hook below is retained as a **backstop** rather than the sole
+defense. Read this ADR for the threat model and the hook's design; read ADR-167 for
+what actually defends against it.
 
 ## Context
 
--- a/docs/architecture/system/ADR-163-config-separation-dotfiles-source-of-truth.md
+++ b/docs/architecture/system/ADR-163-config-separation-dotfiles-source-of-truth.md
@@ -33,10 +33,12 @@ surfaced silent drift that no CI catches:
   credentials was present on north, **absent on slab** — a security drift.
 - Session-link (`Claude-Session:`) suppression worked on north **only because a per-host
   `~/.claude` memory told the agent to disobey the harness prompt** — slab lacked the
-  memory and leaked the URL into commits/PRs. (ADR-162 establishes why: the
-  `attribution.sessionUrl` config key is broken upstream, so a mechanical hook that
-  denies the publishing command is the real fix. *Distributing* that fix is this ADR's
-  concern.)
+  memory and leaked the URL into commits/PRs. (This ADR originally recorded ADR-162's
+  reason: that `attribution.sessionUrl` was *broken upstream*, leaving a mechanical deny
+  hook as the real fix. That premise was false — [[ADR-167]] proves the key governs the
+  link and supersedes ADR-162. The observation above is unaffected: the control, whatever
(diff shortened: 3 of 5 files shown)


aaronsb:
## Review: ADR-167 (supersedes ADR-162)

An ADR that corrects a confidently-wrong ADR has to clear a higher bar than the one it replaces. I tried to break its claims independently rather than read them charitably. **The factual claims hold — every one I could check, exactly.** Two substantive findings below, neither blocking.

### Verification (independent, not taken from the PR body)

| Claim | Verified |
|---|---|
| `#18253` = Co-Authored-By/PR footer bug, CLOSED/COMPLETED | ✅ exact — closed 2026-01-15 |
| `#41873` = "attribution setting does not control session URL", CLOSED/NOT_PLANNED | ✅ exact |
| `#66504` OPEN, argues link should be opt-in | ✅ title confirms the characterization |
| `#69614` OPEN, docs omit `sessionUrl` | ✅ — filed by a third party (`coygeek`), not self-cited |
| v2.1.183 changelog text | ✅ **verbatim**, including the "web and Remote Control sessions" scope |
| "seventeen days before ADR-162 was dated" | ✅ 2026-06-19 → 2026-07-06 |
| `sessionUrl: false` actually projected | ✅ present in `~/.claude/settings.json`; fragment exists at `~/.dotfiles/claude/agent-ways/settings/20-attribution.md` |
| `docs/scripts/adr lint` | ✅ 0 errors, 1 pre-existing ADR-127 warning — matches the PR's test plan claim precisely |

The changelog quote being verbatim rather than paraphrased matters here: the scope qualifier the Decision leans on is upstream's own wording, not the ADR's gloss.

### Finding 1 — the error propagated into **five** artifacts, not four (substantive)

`docs/architecture/system/ADR-163-config-separation-dotfiles-source-of-truth.md:36-39` carries the same false claim, uncorrected by this PR and uncounted by ADR-167's "How the error persisted" list:

> (ADR-162 establishes why: the `attribution.sessionUrl` config key is **broken upstream**, so a mechanical hook that denies the publishing command is the real fix.)

This is the sharpest finding available against this PR, because it lands on the ADR's own thesis. ADR-167 argues that a mis-citation gained credibility by propagating into artifacts that each made the next look corroborated — then enumerates four carriers and misses a fifth. Worse, ADR-163 is *cited approvingly in ADR-167's own Decision section* as the distribution mechanism, and sits in its `related:` list. The ADR reaches for ADR-163's plumbing while leaving ADR-163's inherited error in place.

Not blocking — ADR-163's *decision* (dotfiles as source-of-truth) doesn't depend on the false premise; only a parenthetical motivating example does. But a one-line correction note in ADR-163's Context, or adding it as item 5 with a "left as historical residue" note, would close the loop the ADR itself opens. Silently leaving it is the same failure mode at one remove.

### Finding 2 — retention reason #3 is self-undermining (reasoning)

I went in expecting "retain the hook" to be sentimentality about existing code. **It isn't — two of the three reasons are load-bearing, and I verified the strongest one mechanically.**

The **default-on** argument turns on whether the hook and the fragment share a delivery path. If they did, "a host that never receives the fragment leaks by default" would be vacuous — no fragment would imply no hook either, and the backstop would cover nothing. They don't share one:

- The **hook** travels with the agent-ways framework install (`hooks` is reconcile-co-owned; repo `settings.json:221` references it).
- The **fragment** travels dotfiles → fragment store → `ways settings project`.

Genuinely independent pipelines, so a host with agent-ways but un-deployed dotfiles gets the hook and not the setting. That's a real, reachable failure mode, and ADR-163's own Context documents its historical instance (slab leaked; north didn't). The backstop earns its place. Combined with **undocumented** (#69614 — a real "may change without notice" exposure), retention is justified on the evidence. Retiring the hook would be *less* defensible.

Reason **#3 (scope-qualified)** is the weak one, and it argues against itself within a single paragraph:

> **Scope-qualified upstream** — the changelog scopes it to "web and Remote Control sessions". This operator runs `remoteControlAtStartup: true`, **which is the likely reason the link reached commits at all**; the boundary for other session types is not characterized.

If Remote Control is *why the link appears*, and the setting covers Remote Control, then the scope qualifier describes **coverage, not a gap**. The plain reading of the changelog is that the setting is scoped to web/RC because that is where the link is emitted. The ADR's own discarded `-p` experiment points the same way — headless never injects the trailer. So an "uncovered session type that still leaks" is asserted, not evidenced.

That is precisely the move ADR-167 indicts ADR-162 for, inverted: ADR-162 froze a contingent upstream fact as durable truth; here an *uncharacterized* boundary is presented as a known risk. The Neutral section is honest that the boundary is uncharted ("not worth characterizing while the backstop stands") — but then it can't also be cited in the Decision as an independent reason the backstop is needed. It's circular: the backstop justifies not characterizing the gap, and the gap justifies the backstop. Suggest demoting #3 to "unknown, and the backstop makes it cheap to leave unknown" rather than listing it as a third co-equal reason. Reasons 1–2 carry the decision on their own; #3 weakens a conclusion it isn't needed for.

### Supersession correctness — matches precedent, slightly exceeds it

ADR-162's new header + frontmatter tracks ADR-104 closely and deliberately: `status: Superseded`, `revised:`, `superseded_by: ADR-167`, successor added to `related:`, and a `## Status: Superseded by ADR-167` section that preserves what remains correct (threat model, hook design) while retiring the rationale. That's the ADR-104 shape, including the "read this for X, read the successor for Y" split.

`supersedes: ADR-162` on ADR-167 is precedented as a bare scalar (cf. ADR-107's `supersedes: ADR-107 Draft (2026-03-20)`; others use `supersedes: []`). Worth noting it actually *improves* on the cited precedent — ADR-123 never carried a reverse `supersedes: ADR-104`, so the link was one-directional there. Fine to be better than precedent; just flagging that "per the ADR-104 precedent" slightly undersells it.

### Minor

- **`#41873` "April 2026"** — that's the *creation* date (2026-04-01); it closed 2026-05-22. Adjacent text reads "was closed not-planned, and then upstream shipped" (2026-06-19), so a reader may take "April" as the close date and compute a longer gap than the real 28 days. In an ADR whose subject is citation precision, worth tightening to "opened April 2026, closed not-planned 2026-05-22".
- **`[[ADR-104]]`** is cited in Alternatives Considered but absent from `related:`.
- `related:` format (`"[[ADR-147]]"` vs bare `147` vs `ADR-147`) is inconsistent repo-wide already; ADR-167 matches its ADR-163/164 neighbors. Not a finding.

### Assessment

Decision follows from Context. Consequences are honest — the Negative section names the undocumented-key risk, the invisible-failure mode, and the two-mechanisms cost without softening any of them, and the discarded-experiments disclosure (two void attempts documented as void) is exactly the discipline the ADR is arguing for. Hook change is comment-only; no behavior change. Evidence is a real single-variable experiment, and its limits are stated rather than oversold.

**Non-blocking.** Finding 1 is the one worth acting on — the ADR's thesis is about uncorrected propagation, and it leaves a carrier uncorrected.

---

aaronsb:
## Review findings remediated (0d65d0b)

Both substantive findings were correct and are fixed.

**Finding 1 — five carriers, not four.** Verified: `ADR-163:34-39` does record the false premise ("the `attribution.sessionUrl` config key is broken upstream, so a mechanical hook that denies the publishing command is the real fix"). Corrected the parenthetical — ADR-163's decision never rested on it, and the host-drift observation it illustrates is unaffected — and added it to ADR-167's carrier list.

The miss is now recorded in the ADR rather than quietly patched: the first draft listed four carriers and missed a document it cites approvingly in its own Decision and lists in `related:`. An argument about how unchecked claims propagate had propagated one. Being the one making the argument confers no immunity, and that belongs in the text.

**Finding 2 — reason #3 was circular.** Correct, and the sharper of the two. "Scope-qualified to web and Remote Control sessions" cuts both ways: if Remote Control is the likely reason the link reached commits at all, the qualifier may describe *complete coverage*, not a gap. Holding "the boundary is uncharacterized, therefore keep the backstop" alongside "the backstop stands, therefore the boundary needn't be characterized" is ADR-162's error inverted — an unexamined boundary dressed as known risk.

Rejected it explicitly in the Decision. The retention argument now stands on two reasons: **undocumented** (a key found by reading a changelog is one upstream never promised), and **default-on with an independent delivery path** — which the review sharpened materially. Hook and fragment travel separately (reconcile-owned `hooks` vs dotfiles → fragment store → project), and that independence is *why* the backstop covers anything; ADR-163 records a real instance of one host holding while another leaked. Were they one pipeline the backstop would be absent exactly when needed. That argument is now stated properly rather than assumed.

The uncharacterized boundary moved to Consequences → Neutral, named as an open question with the experiment that would close it (one controlled run with `remoteControlAtStartup` off), so the omission is visible rather than assumed away.

**Minors:** `#41873` dates corrected (filed April, closed 2026-05-22); `[[ADR-104]]` added to `related:` since it's cited as precedent.

Lint: 0 errors, 1 pre-existing ADR-127 warning (unrelated, `Rejected` isn't a valid status).

PR #367, opened 2026-07-17
fix(ways): report the session link off the key that governs it

aaronsb:
## Summary

The `github` macro reported **`Session-link trailer: OFF`** from `attribution.commit`/`.pr` alone and never read `attribution.sessionUrl` — the key that actually governs the link. It asserted "no session trailer" in every session, including sessions whose system prompt carried one. This is the fourth of five artifacts carrying the ADR-162 conflation (see ADR-167, merged in #366), and the sharpest: it rendered the wrong fact as a reassuring green light.

Caught live during the ADR-167 experiment — a control session with `sessionUrl: true` had the trailer in its system prompt *and* read this macro telling it the trailer was off, then had to reason around its own status line contradicting its prompt.

## Changes

- **Report the two controls separately**, because they are separate:
  - `attribution.sessionUrl` (bool, default **true**) → the `Claude-Session:` transcript link — the disclosure control
  - `attribution.commit`/`.pr` (string, `""` disables) → the Co-Authored-By / "Generated with Claude Code" footers — cosmetic, never governed the link
- **Unset `sessionUrl` now reports ON (default)** with the suppression instruction, instead of silently reading as OFF. Unrecognized values fail safe to ON.
- **Per-key precedence resolution.** Claude Code deep-merges objects, but the old loop took the first file containing *any* `attribution` object and ignored the rest — so a project file setting `.commit` would hide a user file setting `.sessionUrl`. Each sub-key now resolves independently. The `__UNSET__` sentinel comes from `jq`, not from emptiness, since `""` is a legitimate value for `commit`/`pr`.

## Test plan

Verified by running the macro, not by inspection:

| Case | Result |
|---|---|
| `commit`/`pr` empty, `sessionUrl` **absent** — the exact pre-fix state | **ON (default)** + suppression instruction (old code: "OFF") |
| `sessionUrl: true` | **ON** + instruction |
| no `attribution` key at all | **ON (default)**; footers `default (ON)` |
| `sessionUrl: "yes"` (garbage) | unrecognized, **treat as ON** |
| live settings (`sessionUrl: false`) | **OFF** `[~/.claude/settings.json]` |
| `.commit` in project file + `.sessionUrl` in user file | each resolves from its own source |

`bash -n` clean.

## Found while doing this — not fixed here

Section 3 sits below an early exit at line 8 (`gh repo view &>/dev/null || exit 0`), so the session-link surface silently vanishes when gh is unauthenticated, offline, rate-limited, or the repo isn't on GitHub. The trailer is a `git commit` concern, not a GitHub one — it applies in any repo. That's a placement decision (does this belong in the `github` way at all, or in `delivery/commits`?) rather than a mechanical fix, so it's tracked separately instead of smuggled in here.

### DIFF
--- a/hooks/ways/softwaredev/delivery/github/macro.sh
+++ b/hooks/ways/softwaredev/delivery/github/macro.sh
@@ -280,10 +280,15 @@ fi
 # A surface that asserts a fact it never read is worse than no surface.
 #
 # Precedence approximates Claude Code's resolution order (project-local >
-# project > user-global, highest wins). Each sub-key resolves INDEPENDENTLY
-# because Claude Code deep-merges objects: a project file setting .commit does
-# not hide a user file setting .sessionUrl. Managed/enterprise policy and
-# command-line overrides are not accounted for.
+# project > user-global, highest wins). Each sub-key resolves INDEPENDENTLY, on
+# the documented merge law that objects deep-merge key by key (ADR-147; mirrored
+# in tools/ways-cli/src/cmd/settings/compile.rs) — so a project file setting
+# .commit must not hide a user file setting .sessionUrl. NOTE: that law is
+# documented for settings merging generally; its application to `attribution`
+# ACROSS SCOPES is inferred, not verified. If Claude Code instead replaces the
+# whole object at the highest-precedence file, per-key resolution reports the
+# wrong source at that seam (the common single-file case is unaffected).
+# Managed/enterprise policy and command-line overrides are not accounted for.
 
 PROJECT_ROOT=$(git rev-parse --show-toplevel 2>/dev/null)
 ATTR_FILES=(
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: Section 3 rewrite

The core fix is right and I verified it rather than eyeballing it: `sessionUrl` is now the key that drives the session-link line, unset correctly resolves to **ON (default)**, `""` survives as a legitimate value for `commit`/`pr`, and per-key precedence genuinely works (project-local `.commit` + project `.sessionUrl` each resolve from their own source — reproduced). Splitting the disclosure control from the cosmetic footers is the correct model, and the `${var/#$HOME/~}` tilde-expansion gotcha comment surviving the rewrite is good. The "found while doing this — not fixed here" note on the line-8 `gh repo view` early exit is the right call: that's a placement decision, not a mechanical fix.

That said — this section's thesis is *"a surface that asserts a fact it never read is worse than no surface."* I ran it against adversarial settings and found cases where the new code still does exactly that. Two of them I'd want fixed before merge.

---

### 1. An empty settings file short-circuits the whole precedence chain (blocking)

**Location**: `hooks/ways/softwaredev/delivery/github/macro.sh`, `resolve_attr_key()`, lines ~42-47

`jq` on an **empty or whitespace-only** file emits nothing and **exits 0** — so `|| continue` doesn't fire, `v=""`, `"" != "__UNSET__"`, and the function returns `"|<that file>"` and stops. A zero-byte `.claude/settings.json` (common enough — tooling creates them) masks everything below it.

Reproduced: empty `$PROJECT_ROOT/.claude/settings.json`, real config in `~/.claude/settings.json` (`sessionUrl: false`, `commit: "custom footer"`):

```
**Session-link trailer**: unrecognized `attribution.sessionUrl` value `` (expected true/false) [.../proj/.claude/settings.json] — treat as ON.
**Attribution footers** ...: OFF (empty) [.../proj/.claude/settings.json]
```

Both lines are wrong (truth: OFF, and `"custom footer"`), and the user file is never read. The session-link line at least fails toward ON (the loud direction — good), but the footer line asserts a flat falsehood. `{}` and `{"attribution":{}}` are both handled correctly; it's specifically the empty/whitespace file.

Note the emptiness check you'd reach for is unavailable here *by design* — `""` is a legitimate `commit` value — which is what makes this subtle.

### 2. `sessionUrl: "false"` (string) is reported as a definite OFF (blocking — wrong direction)

**Location**: same function, the `tostring` in the jq filter

`tostring` erases the bool/string distinction, so the string `"false"` becomes `false` and hits the `false)` case:

```
**Session-link trailer**: OFF — `attribution.sessionUrl: false` [~/.claude/settings.json]
```

Claude Code reads this key as a boolean; a string `"false"` is very unlikely to disable anything. So the macro renders a reassuring green light off a value it can't actually know the meaning of — the ADR-167 conflation in miniature. Compare `sessionUrl: 0`, which correctly reports "unrecognized … treat as ON". The inconsistency is the tell: only strings that happen to spell `true`/`false` get laundered into type-correct answers.

### 3. A `|` in an attribution value corrupts both value and source

**Location**: lines ~62-64, the `<value>|<source>` packed return

`${VAR%%|*}` truncates at the *first* pipe. With `attribution.commit: "Co-Authored-By: X | Y"`:

```
**Attribution footers** ...: custom: "Co-Authored-By: X " [ Y|/home/.../.claude/settings.json]
```

Value truncated, source label is garbage. `sessionUrl` is immune (bools), so this is footer-only and cosmetic-ish — but it's a display surface printing a value that isn't the value. Reordering to `<source>|<value>` with `${RAW%%|*}` / `${RAW#*|}` would mostly work (paths rarely contain `|`), but returning via globals avoids the packing question entirely — and also avoids `$( )` silently eating trailing newlines in a value.

### 4. Malformed / mistyped `attribution` reports "no settings file" (non-blocking, but on-thesis)

`{"attribution": "off"}` (string) and truncated JSON both make jq exit non-zero → `continue` → eventually:

```
**Session-link trailer**: **ON (default)** — `attribution.sessionUrl` is unset and defaults to true. ...[no settings file]
```

"Unset, defaults to true" and "no settings file" are both false — there *is* a settings file, it's just broken, and it may well say `sessionUrl: false`. Falling through is defensible; claiming the file doesn't exist isn't. A `— warning: <file> failed to parse` would keep the surface honest. Same root cause as #1: `jq` failure modes are being treated as "key absent".

### 5. Literal `"__UNSET__"` collides with the sentinel (non-blocking, theoretical)

`attribution.commit: "__UNSET__"` → `continue` → reported as `default (ON) [no settings file]`. The PR description's claim that the sentinel "comes from jq, not from emptiness" is true and is a real improvement, but the sentinel still lives in the same value-space as user data. Absurd input; noting it because it disappears for free under the fix below.

---

### Suggested fix — `tojson` + a parse guard collapses 1, 2, 3, 4(partly), 5

Switching `tostring` → `tojson` makes the return type-aware and puts the sentinel in a space no string value can reach (`null` vs `"null"`), which in turn makes an emptiness guard safe again:

```bash
# sets ATTR_VAL / ATTR_SRC — no packing, so `|` and newlines in values are moot
resolve_attr_key() {
  local key="$1" f v
  ATTR_VAL="null"; ATTR_SRC=""
  for f in "${ATTR_FILES[@]}"; do
    [[ -f "$f" ]] || continue
    jq -e . "$f" >/dev/null 2>&1 || continue   # empty / whitespace / malformed
    v=$(jq -r --arg k "$key" '
          if (.attribution|type) == "object" and (.attribution|has($k))
          then (.attribution[$k] | tojson)
          else "null" end' "$f" 2>/dev/null) || continue
    [[ -z "$v" || "$v" == "null" ]] && continue
    ATTR_VAL="$v"; ATTR_SRC="$f"; return
  done
}
```

`jq -e .` returns 4 on empty/whitespace and 5 on malformed (verified), so #1 dies at the guard. Callers then match on JSON tokens: bare `false`/`true` are the only things that produce a definite OFF/ON, `"false"` arrives quoted and falls to "unrecognized" (#2), `""` is `""` and stays distinguishable from absent (#5), and `.attribution[$k]` is only touched once `type == "object"` is confirmed, so no jq error path for string/array `attribution` (#4). Footer display needs one `fromjson`/quote-strip.

### One unverifiable assumption

The comment asserts Claude Code **deep-merges** `attribution` (which is what justifies per-key resolution). I can't confirm that from this repo, and if it turned out to be a shallow object replace, per-key resolution would be the *wrong* model — right answer for most real configs, wrong at the seams. Worth a pointer to whatever established it, or an explicit "assumed, unverified".

---

**Assessment**: net clearly better than `main` — the reported bug is genuinely fixed and the two-controls model is correct. Recommending changes on #1 and #2 specifically, because both reproduce the failure mode this PR exists to eliminate: #1 asserts footer state it never read, #2 asserts a confident OFF while the trailer is plausibly live. #3-#5 are cheap to sweep in with the same refactor.

No prompt-injection or credential-handling concerns in this diff.

aaronsb:
## All five findings remediated

Both blocking findings were correct, and the framing is the part worth keeping: **they recreated the exact failure this PR exists to kill** — a confident OFF while the trailer is live — through a different door. An empty `settings.json` masked the real config; `sessionUrl: "false"` laundered a string into a bool-looking `false`. I'd have shipped the bug's twin.

| # | Finding | Fix | Verified |
|---|---|---|---|
| 1 | empty file short-circuits chain (blocking) | `jq -e .` parse-guard | zero-byte + whitespace-only → **ON (default)** + ⚠ |
| 2 | `sessionUrl: "false"` → confident OFF (blocking) | `tojson` not `tostring` | → **ON**, names the type error |
| 3 | `\|` in value corrupts value+source | return via globals, no packing | `"Co-Authored-By: X \| Y"` → intact, correct source |
| 4 | malformed/mistyped → "[no settings file]" | explicit ⚠ for both cases | unparseable **and** non-object `attribution` each warn |
| 5 | literal `"__UNSET__"` collides | sentinel moved out of string space | → `custom: "__UNSET__"` |

The `tojson` suggestion carries more than it looks: it can never emit an empty string, which is what makes an absence test safe again even though `""` is a legitimate `commit`/`pr` value. That was the crux of #1 and I'd have papered over it.

**Fail-safe direction is now explicit.** Anything not readable as boolean `false` reports **ON**, with the reason. A false OFF is the only failure mode that matters here; a false ON is merely noisy.

**On silent skipping** — a file that might say `sessionUrl: false` must never be dropped without a word, so unparseable files and non-object `attribution` now warn rather than vanish. That's the same principle as the original bug, one level down.

**Deep-merge**: you're right that nothing here establishes it. It *is* grounded — ADR-147 documents the law and `tools/ways-cli/src/cmd/settings/compile.rs:7` mirrors it ("objects deep-merge (recurse key by key)") — but that's the law for settings merging generally; its application to `attribution` **across scopes** is inferred. Now cited *and* marked as inferred in the comment, with the consequence spelled out: if Claude Code instead replaces the whole object at the highest-precedence file, per-key resolution reports the wrong source at that seam. The common single-file case is unaffected. Verifying it properly needs a controlled run and is worth its own task rather than a confident comment.

Re-verified: all five cases, plus the live config (`sessionUrl: false` → OFF) and the pre-fix shape (`commit`/`pr` empty + `sessionUrl` absent → ON). `bash -n` clean.

PR #369, opened 2026-07-17
fix(attend): keep `attend reply` working for external senders

### LINKED ISSUE #368: attend reply leaks unactionable --re error when replying to external (@-bearing) senders
## Summary

`attend reply` fails with a leaked, unactionable error when the message it is replying to came from an **external sender** (notably Aaron messaging an agent via the `attend chat` TUI). The agent sees:

```
attend send: --re signal id must be non-empty and match [A-Za-z0-9_-]+
```

...for a command it never typed (`send`) and a flag it never passed (`--re`). The agent did nothing wrong — it ran a valid `attend reply "<message>"` — and is left with no way to recover, so it abandons the reply.

## Root cause

`attend reply` resolves the last inbound signal id (the signal's filename stem) and delegates to `attend send --re <id>`. Both signal write sites build the filename by sanitizing only `/`:

- `tools/attend/src/cmd/send.rs:172`
- `tools/attend-chat/src/signal.rs:136`

```rust
let filename = format!("{}-{}.signal", sender_id.replace('/', "-"), ts);
```

External senders have id `$USER@<terminal>` (e.g. `aaron@kitty`, or `aaron@iterm.app` with a `.`). The resulting signal id — e.g. `aaron@kitty-1712345` — contains `@`/`.`, which fail `is_valid_signal_id` (`[A-Za-z0-9_-]+`). `cmd_reply` feeds that id straight into `cmd_send`, whose `--re` validation rejects it and prints the internal error.

## Why the fix is lossless

Threading is currently cosmetic: `parse_signal` drops the `re:` id at parse time and the peer sensor never renders thread context (`sensor-peers/src/lib.rs`). A threaded reply and an unthreaded send are byte-identical in what any peer sees, so degrading `reply` -> unthreaded `send` when a valid id can't be produced loses nothing observable.

## Proposed fix (two parts)

1. **`cmd_reply` never hard-fails on an id it resolved internally.** Validate the resolved id; if it is *present but malformed*, degrade to an unthreaded `send` (deliver the message, one-line stderr note, exit 0). Keep the existing clear error only for the genuine *no-inbound* case (nothing to reply to -> use `send`).
2. **Stop recording invalid ids at the source.** Replace `.replace('/', "-")` at both write sites with a sanitizer mapping any non-`[A-Za-z0-9_-]` char to `-`, so signal ids are always valid regardless of sender. The `from` field keeps the real `external:aaron@kitty`; only the filename/id is normalized.

## Acceptance criteria

- [ ] `attend reply "<msg>"` when the most-recent inbound came from an external (`@`-bearing) sender delivers the message and exits 0 (degraded to unthreaded send with a note), rather than exiting 1 with the `--re` error.
- [ ] `attend reply` with no prior inbound still exits 1 with the clear "use `send` instead" message (unchanged).
- [ ] New signals written by both the CLI and the chat TUI have ids matching `[A-Za-z0-9_-]+` even for external senders.
- [ ] Unit coverage for the sanitizer and the reply-degradation path.

## Repro

1. From the `attend chat` TUI (or any external sender), send a message to an agent's session.
2. In the agent session: `attend reply "ack"`.
3. Observe exit 1 and the leaked `attend send: --re signal id must be non-empty...` error.

aaronsb:
Closes #368.

## Problem

`attend reply` failed with an unactionable, internal-leaking error whenever the message it was replying to came from an **external sender** — notably Aaron messaging an agent through the `attend chat` TUI:

```
attend send: --re signal id must be non-empty and match [A-Za-z0-9_-]+
```

The agent ran a valid `attend reply "<msg>"` and got an error naming a command (`send`) and flag (`--re`) it never used, with no way to recover — so it abandoned the reply.

## Root cause

`attend reply` resolves the last inbound signal id (the signal's filename stem) and delegates to `attend send --re <id>`. Both write sites (`attend/src/cmd/send.rs`, `attend-chat/src/signal.rs`) built the filename sanitizing only `/`. External senders are `$USER@<terminal>` (e.g. `aaron@kitty`, `aaron@iterm.app`), so the id carried `@`/`.`, which fail `is_valid_signal_id` (`[A-Za-z0-9_-]+`). `cmd_reply` passed it straight to `cmd_send`, whose `--re` validation rejected it.

## Fix (two parts)

1. **Sanitize ids at the source.** New shared `agent_identity::sanitize_id_component` maps any non-`[A-Za-z0-9_-]` char to `-`; both write sites use it. Ids are now always valid regardless of sender. The `from` field keeps the raw identity (`external:aaron@kitty`).
2. **`attend reply` never hard-fails on an id it resolved internally.** A *malformed* last-inbound degrades to an unthreaded send (delivered, one-line stderr note, exit 0). Threading is cosmetic today — `parse_signal` drops the `re:` id and no peer renders it — so degradation is lossless. The genuine *no-inbound* case keeps its clear "use \`send\` instead" error.

## Verification

Unit tests for the sanitizer and the reply classifier (including the exact `aaron@kitty` / `aaron@iterm.app` regression inputs). Full workspace suite green (814 tests).

Driven end-to-end on the built binary:

| case | before | after |
|------|--------|-------|
| malformed (`@`-bearing) last-inbound | exit 1, `--re` error | exit 0, unthreaded send delivered + note |
| valid last-inbound | threaded | threaded (`re:` in wire) — unchanged |
| no last-inbound | exit 1, clear error | exit 1, clear error — unchanged |
| external `send` | id `aaron@kitty-…` (invalid) | id `aaron-kitty-…` (valid); `from` keeps `external:aaron@kitty` |

### DIFF
--- a/tools/agent-identity/src/identity.rs
+++ b/tools/agent-identity/src/identity.rs
@@ -105,6 +105,38 @@ pub fn sanitize_id_component(s: &str) -> String {
         .collect()
 }
 
+/// Monotonic per-process counter, appended to signal filenames so two
+/// writes that land in the same clock tick still get distinct names.
+static SIGNAL_SEQ: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+
+/// Build a collision-resistant signal filename for `sender_id`. The stem
+/// (name without `.signal`) doubles as the signal id that `re:<id>`
+/// threaded replies and `attend reply`'s last-inbound record reference,
+/// so it is always `[A-Za-z0-9_-]+`.
+///
+/// Shape: `<sanitized-sender>-<unix-nanos>-<seq>.signal`. Uniqueness has
+/// two independent guards because a signal bus takes bursts:
+/// - **nanosecond timestamp** separates writers across processes (two
+///   processes hitting the same nanosecond is not realistic);
+/// - **process-local sequence** guarantees uniqueness *within* a process
+///   even if the platform clock is coarse enough to repeat a nanos read.
+///
+/// This closes the collision window the `sanitize_id_component` fix
+/// (issue #368) would otherwise widen — sanitizing can map distinct
+/// external identities (`iterm.app` vs `iterm-app`) to the same sender
+/// stem, and a second-granularity name let same-second writes silently
+/// overwrite each other on rename. Centralized here so both write sites
+/// (`attend send`, the attend-chat TUI) stay byte-identical — see the
+/// wire-format-mirror note in `attend::cmd::send`.
+pub fn signal_filename(sender_id: &str) -> String {
+    let nanos = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_nanos())
+        .unwrap_or(0);
+    let seq = SIGNAL_SEQ.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+    format!("{}-{}-{}.signal", sanitize_id_component(sender_id), nanos, seq)
+}
+
 /// FNV-1a 64-bit. Deterministic across runs and targets — this is the
 /// property we need, and why we don't use `std::hash::DefaultHasher`.
 /// Crate-internal (`pub(crate)`) so the groups module shares the exact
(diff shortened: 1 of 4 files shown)


aaronsb:
## Review addressed (high-effort review, 4 findings)

Pushed 9fe04b9.

- **#1 same-second collision (CONFIRMED)** & **#2 sanitizer collapsing distinct ids (PLAUSIBLE)** — fixed at the root by a new `agent_identity::signal_filename`, which makes names unique by construction: `<sanitized>-<unix-nanos>-<seq>`. Nanos separates writers across processes; a process-local atomic sequence covers same-process bursts even under a coarse platform clock. No colliding stem ⇒ no silent tmp+rename overwrite. New burst test asserts 1000 rapid names are all distinct and valid.
- **#3 duplication (CONFIRMED)** — both write sites now call the one helper; the stem shape can't drift between `attend send` and the attend-chat TUI.
- **#4 Unthreaded branch "becomes dead" (PLAUSIBLE)** — **kept**. It's cheap, permanent insurance against *any* non-threadable id (stale record, hand-edited state file, a future writer that skips `signal_filename`), which is exactly the "never leak the internal error to an agent" goal of this PR. Rewrote the comment so it no longer reads as a transition shim.

Verified on the real binary: two rapid external sends yield two distinct valid files (no overwrite); reply still degrades on a malformed id (exit 0, unthreaded) and threads on a valid one. Full suite green.

PR #371, opened 2026-07-17
feat(attend-chat): echo directed @mention sends in the sender's own transcript

### LINKED ISSUE #370: attend-chat: directed @mention sends don't echo in the sender's own transcript
## Summary

When you `@mention` an agent in the `attend chat` TUI, the message is delivered but **never appears in your own transcript** — it looks like it vanished. A broadcast (`#open` / no mention) *does* appear in your own view, so the asymmetry reads as "single mention works, directed mention is broken."

This is the genuinely-present half of the multi-@mention report (the routing half — wrong-agent / only-one-delivered — was already fixed in c7c938f and ships in the installed binary). It is **not** multi-mention-specific: a single directed `@name` also fails to echo.

## Root cause

Directed sends are written to the **recipient's** cwd inbox (`signals_base/<recipient-encoded-cwd>/`). The sender's own chat watcher only surfaces `_broadcast/`, `@<group>/`, and the sender's **own** cwd (`watcher::accept_path`, `tools/attend-chat/src/watcher.rs:51`). A broadcast self-echoes because it rides `_broadcast/`, which the sender *does* watch; a directed send goes to a cwd the sender does not watch, and there is no local echo — the render loop (`app/mod.rs`) only shows signals the watcher streams.

## Fix

Echo directed sends as a **display-only** concern (no extra bus write): on a successful directed send, `handle_enter` returns the composed self-`Signal` and the Enter handler appends it to the transcript, mirroring the watcher-drain. Broadcasts are untouched (they already echo via the round-trip). The message is echoed once regardless of recipient count; the `sent → @a @b` status line reports where it went — which also surfaces the *resolved* target, so a fuzzy-match recovery (`@Tasmin` → `Tamsin`) is now visible to the sender.

## Acceptance criteria

- [ ] A directed `@name` send appears in the sender's own transcript, rendered with the sender's chip (as a broadcast already does).
- [ ] A multi-mention `@a @b` send echoes once, not N times.
- [ ] Broadcasts still echo exactly once (no double-render).
- [ ] The echo writes nothing to the signal bus.

## Notes / follow-ups

- The **fuzzy nickname match** (`resolve_nickname`, Levenshtein ≤2) is working as designed — exact-first, refuses on ambiguity/ties, capped, with a live regression test. Left unchanged; the echo makes its resolution visible.
- **Same-cwd sibling instances** (ADR-129 `-alpha`/`-beta` sharing one cwd) cannot be addressed separately under the cwd-keyed routing model — a directed send reaches the cwd, not a specific instance. That is an architectural question (ADR-129 ↔ ADR-136 "one attend per cwd") tracked separately as an ADR, not folded into this fix.

aaronsb:
Closes #370.

## Problem

`@mention` an agent in the `attend chat` TUI and the message is delivered but **never shows up in your own transcript** — it reads as vanished. Broadcasts (`#open`) *do* appear in your own view, so the asymmetry looks like "single mention works, directed is broken." This is the genuinely-present half of the multi-@mention report; the routing half (wrong-agent / only-one) was already fixed in `c7c938f` and ships in the installed binary.

## Root cause

Directed sends are written to the **recipient's** cwd inbox. The sender's own watcher only surfaces `_broadcast/`, `@<group>/`, and the sender's own cwd (`watcher::accept_path`). A broadcast self-echoes because it rides `_broadcast/`; a directed send goes to a dir the sender doesn't watch, and there was no local echo.

## Fix

Echo directed sends as a **display-only** concern:
- `signal::compose_self_echo` builds the in-memory `Signal` for the just-sent message from this session's own identity, so it renders with the sender's chip exactly as their own broadcast already does. It writes nothing.
- `handle_enter` returns it via a new `EnterAction::ClearWithStatusAndEcho`; the Enter handler appends it to the transcript, mirroring the watcher-drain (read/clone/push/cap/set).
- Echoed once regardless of recipient count. The `sent → @a @b` status line reports routing — and surfaces the *resolved* target, so a fuzzy-match recovery is now visible.
- Broadcasts untouched (no double-render).

## Scope decisions (documented in #370)

- **Fuzzy nickname match** (`resolve_nickname`, Levenshtein ≤2): working as designed — exact-first, refuses on ambiguity/ties, capped, live regression test. Left unchanged; the echo makes its resolution visible, which addresses the "did it go where I meant?" concern without regressing a tested feature.
- **Same-cwd sibling instances** (ADR-129 `-alpha`/`-beta`): can't be addressed separately under cwd-keyed routing. Architectural (ADR-129 ↔ ADR-136); routed to an ADR, not this fix.

## Verification

- Unit: `compose_self_echo` (message + self-identity + non-bus id); full suite green (816).
- Clean build + clippy; the new `EnterAction` variant and the guard-scoping in the Enter handler are borrow-checked (the echo arm must `signals.set` after dropping the read guard — mirrors the existing watcher drain).
- **Not driven live:** the TUI echo *row* isn't exercised end-to-end — iocraft needs a real TTY and directed resolution needs a live peer with a fresh heartbeat, neither scriptable here. The echo logic is a minimal branch over already-tested primitives (`parse_addressed`, `resolve_recipients`, `compose_self_echo`). Happy to walk through a manual two-session check.

### DIFF
--- a/tools/attend-chat/src/app/keys.rs
+++ b/tools/attend-chat/src/app/keys.rs
@@ -15,7 +15,10 @@ use crate::legend::{
     parse_addressed, Addressed, Sigil,
 };
 use crate::sessions::discover as discover_sessions;
-use crate::signal::{compose_self_echo, cwd_dir, write_broadcast, write_signal, Signal};
+use crate::signal::{
+    compose_self_echo, cwd_dir, encode_cwd, signals_base, write_broadcast, write_signal, Signal,
+};
+use crate::watcher::accept_path;
 use crate::slash;
 use crate::text_layout::split_at_char;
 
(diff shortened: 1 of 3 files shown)


aaronsb:
## Review addressed (high-effort review, 4 findings + 2 refuted)

Pushed the remediation.

- **#1 double-render (CONFIRMED, correctness) — the important one.** The echo fired for *every* non-empty recipient run, but the premise "directed sends don't round-trip" is false for `#open` (→ `_broadcast/`), `#group` (→ `@group/`), and an `@agent` in the sender's *own* cwd (→ `own_encoded`) — all watched by the sender's own watcher. Those double-rendered. Fixed: echo only when **no** destination is self-watched, reusing `watcher::accept_path` as the single source of truth. Only `@agent` sends to *other* cwds echo. Regression test (`sender_watches_matches_the_watchers_accept_set`) added.
- **#2 blank chip (CONFIRMED).** `"".rsplit('/').next()` is `Some("")`, so the `"?"` fallback was dead. Fixed via `find(non-empty)` in the new shared helper.
- **#3 duplicated cap/drain (CONFIRMED).** Extracted `push_capped()`; both the watcher drain and the echo append go through it.
- **#4 duplicated identity derivation (CONFIRMED).** Extracted `sender_identity()`; `write_signal` and `compose_self_echo` share it, so the echoed chip can't drift from the delivered signal's. (This is also where #2 is fixed once for both.)
- The 2 refuted concurrency findings: agreed — iocraft `State` on the single-threaded UI loop, same pattern the existing drain already uses.

Full attend-chat suite green (157).

PR #372, opened 2026-07-17
docs(adr): ADR-168 — instance-addressable directed messaging for same-cwd attend siblings (Draft)

aaronsb:
Draft ADR for debate — captures the architectural half of the multi-@mention report.

## Why

Directed `@Nickname` routing is **cwd-keyed**. ADR-129 instance suffixes (`@Tamsin-alpha`/`-beta`) let several Claude Code sessions share one cwd and be told apart in the legend — but `KnownIdentity.cwd` is identical for both, so they resolve to the same inbox. You cannot direct a message to exactly one co-located instance; the UI implies a capability the bus doesn't provide. That's the same-cwd half of the multi-@mention report (the routing/parse half was already fixed in `c7c938f`; the no-local-echo half is #370 / #371).

## What the ADR proposes

Status **Draft**, pending the deciders. It frames one question — *is addressing one specific co-located instance a workflow anyone needs, or is legend distinguishability enough?* — and lays out:

- **Option A (proposed immediate step):** accept cwd-scoped addressing; make the UI honest (collapse sibling suffixes at send, report "reaches all instances in `<cwd>`"). Cheap, preserves ADR-136's one-inbox-per-cwd invariant.
- **Option C (preferred *if* precision is needed):** session-id addressing — route by target session UUID, decoupling routing from cwd at the root.
- **Option B (rejected):** instance sub-inboxes under the cwd dir — more complex than C, same invariant rework, no offsetting benefit.

No code change here — this is the decision artifact. Related: ADR-120/124/129/136.


aaronsb:
Renumbered **ADR-166 → ADR-168**: 166 is already claimed by the in-flight PR #360 (single-source-of-truth for context-window resolution). The `adr` tool handed me the 166 gap because #360 isn't on `main` yet. Content unchanged.

PR #380, opened 2026-07-22
Stable session identity: attend-session crate + attend whoami (ADR-171)

### LINKED ISSUE #378: Attend identity instability: sessions mint multiple personas (ghost entries in chat and peers)
## Observed (2026-07-21, live multi-session test)

The operator saw four identities for what were two claude sessions + one human: `hana`, `Hana-beta`, `isadora-alpha`, plus the ghost `Hana-alpha`. Reproduced causes from one session's own history:

1. **Identity keys off process cwd at attend launch, not the session's canonical project dir.** A stale `cd tools/` in the session's shell put its attend on the bus as a completely different persona (`Sable`-in-`tools`) than its session record (`Hana`-in-`agent-ways`).
2. **Restarting attend mints a new instance slot instead of reclaiming the session's existing one.** The same session went from `-alpha` to `-beta` across a restart; the alpha registration lingered as a ghost until GC.
3. **The chat legend derives identities from historical signal wire data** (`from`/cwd captured at send time), so superseded personas keep rendering as distinct users for as long as their signals sit in the buffer/backfill.

## Proposed invariant (operator)

A claude's identity should be stable under `(sessionId ∩ origin path in ~/.claude/projects)`:

- attend (run *and* send) resolves its cwd from the session record, not `current_dir()`
- instance-suffix assignment is idempotent per sessionId — re-registration returns the existing suffix
- wire identity carries the canonical pair so receivers can collapse superseded personas

## Downstream dependency

The drain/consumption-checkpoint design being specced (planned ADR-171) keys per-session consumption on exactly this stable identity; `/purge`'s planned unconsumed-by-live-session refusal inherits it too. Fluid identity breaks both ("delivered under isadora-alpha doesn't match isadora").

Related: ADR-129 (instance suffix + heartbeat), ADR-168 (instance-addressable directed messaging — the shared-cwd inbox also delivered an `@Hana-alpha`-addressed message to the beta sibling during this test), ADR-170 (username identity for humans, which is stable by construction and unaffected).

aaronsb:
## Summary

- **ADR-171 (Accepted)**: a claude's bus identity is `(sessionId ∩ origin_path)` — the session UID plus the *session record's* cwd, never the process cwd. The roster enumerates addressable coordinating units (top-level sessions); subagents are efferent limbs with no roster identity; ordinals are presentation, never keys. Encodes the three-party design debate and operator sign-off on #378.
- **New `attend-session` crate** owns the single identity derivation, consumed by attend (run/send/status/peers/inbox/config), sensor-peers (now a delegating wrapper), the instance registry key, heartbeat id, and group member ids. Fixes the latent minimal-build degradation to `pid-<pid>` member ids.
- **Ghost-class fixes**: `attend run` takes its working dir from the session record (shell-`cd` drift can't mint personas); periodic instance maintenance upserts instead of touching (bare-nickname entries self-heal within one interval).
- **`attend whoami`** (`--machine` for `key=value` stable fields) — the CLI seam the drain-checkpoint research asked for, honoring CLI-is-the-contract. `docs/cli/attend.md` regenerated and now tracked.

Closes #378.

## Test plan

- 264 tests across the affected crates, clippy clean, release build green.
- Live verification on the real bus: `attend whoami` from a deliberately drifted shell cwd (`tools/`) resolves the session's true origin path and suffixed display name — the exact failure case that motivated #378.

Deferred (recorded in ADR-171): subagent activity decoration on parent chips (presentation-only); historical-signal persona residue ages out naturally.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -91,6 +91,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-168](./system/ADR-168-instance-addressable-directed-messaging-for-same-cwd-attend-siblings.md) | Instance-addressable directed messaging for same-cwd attend siblings | Draft |
 | [ADR-169](./system/ADR-169-agent-ways-relinquishes-user-scoped-settings-json-retains-only-its-operational-baseline.md) | agent-ways relinquishes user-scoped settings.json; retains only its operational baseline | Proposed |
 | [ADR-170](./system/ADR-170-human-focus-group-membership-via-username-identity-and-a-shared-attend-groups-crate.md) | Human focus-group membership via username identity and a shared attend-groups crate | Accepted |
+| [ADR-171](./system/ADR-171-stable-session-identity-the-roster-enumerates-addressable-coordinating-units.md) | Stable session identity — the roster enumerates addressable coordinating units | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-171-stable-session-identity-the-roster-enumerates-addressable-coordinating-units.md
+++ b/docs/architecture/system/ADR-171-stable-session-identity-the-roster-enumerates-addressable-coordinating-units.md
@@ -1,0 +1,115 @@
+---
+status: Accepted
+date: 2026-07-21
+deciders:
+  - aaronsb
+  - claude
+related:
+  - ADR-129
+  - ADR-136
+  - ADR-168
+  - ADR-170
+---
+
+# ADR-171: Stable session identity — the roster enumerates addressable coordinating units
+
+## Context
+
+A live multi-session test (issue #378) surfaced the "multiple personalities"
+failure: one human and two claude sessions rendered as four-plus identities
+(`hana`, `Hana-beta`, the ghost `Hana-alpha`, a `tools/`-keyed persona from
+the same session). Three root causes were reproduced from one session's own
+history:
+
+1. **Identity keyed off process cwd at launch.** attend derived its working
+   dir from `current_dir()`, so a stray shell `cd` leaking into an
+   `attend run` launch (a Monitor inheriting a build directory) put the
+   session on the bus as a different persona than its project.
+2. **Registration happened once, at startup.** The periodic instance-registry
+   maintenance used `touch`, which is a no-op for a missing entry — a session
+   whose registration was GC'd or never written rendered as the bare
+   nickname forever.
+3. **Historical wire data kept superseded personas alive** in the chat
+   legend for as long as their signals sat in the buffer.
+
+The debate (operator + two peer sessions, recorded on #378) also settled what
+identity *is* on this bus, which the fix must encode.
+
+## Decision
+
+**Identity is the coordinating unit, not the process.** A claude's canonical
+identity is the tuple `(sessionId ∩ origin_path)` — the session UID paired
+with the *session record's* cwd, never the process cwd. Concretely:
+
+- **A new `attend-session` crate owns the one derivation** of "who am I on
+  the bus": pid-ancestry walk over `~/.claude/sessions/*.json` for the
+  session UID, the session record's `cwd` as the origin path, with explicit
+  flagged fallbacks (`pid-<pid>` + process cwd, `resolved: false`) for
+  processes no Claude session owns. attend (run, send, status, peers, inbox,
+  config), sensor-peers, the instance registry, the heartbeat id, and
+  focus-group member ids all resolve through it. Downstream durable state —
+  the planned per-session consumption checkpoint (drain research) — keys on
+  the same tuple by calling the same derivation.
+- **`attend whoami` is the CLI accessor** for the tuple (`--machine` emits
+  `key=value` of only the stable fields), so hooks and scripts obtain
+  identity without touching attend-owned state. The rendered display name is
+  deliberately absent from machine output: **ordinals are presentation,
+  never keys.**
+- **The roster enumerates addressable coordinating units** — top-level
+  sessions, by session UID per origin dir. A second top-level instance in
+  the same origin dir takes the next Greek letter (the existing ADR-129
+  allocator, which was already idempotent per `(cwd, sessionId)`; this ADR
+  fixes its *inputs*). Subagents are the supervisor's efferent limbs, not
+  afferent participants: they get no roster identity and no letter. Their
+  activity may render as decoration on the parent's chip (deferred,
+  presentation-only).
+- **The periodic registry maintenance upserts** (`register`, idempotent)
+  instead of touching, so a session with a missing entry self-heals within
+  one interval instead of rendering bare forever.
+
+## Consequences
+
+### Positive
+
+- A session's bus persona is immune to shell cwd drift — the launch
+  environment can no longer mint accidental identities.
+- The "bare nickname" degradation self-heals; ghosts stop accumulating at
+  their source.
+- One identity derivation shared by four consumers replaces three partial
+  ones; the drain checkpoint research can key on it without re-implementing
+  resolution.
+- A minimal attend build (without sensor-peers) now resolves real session
+  identity instead of degrading to `pid-<pid>` member ids.
+
+### Negative
+
+- Session-record resolution costs a sessions-dir walk plus a `ps`-based
+  ancestry climb per subcommand invocation — negligible at CLI cadence, but
+  no longer a bare `getcwd`.
+- Historical signals written under superseded personas still render as
+  distinct chips until they age out of buffers — display residue this ADR
+  accepts rather than rewriting history.
+
+### Neutral
+
+- Humans are unaffected: their username identity (ADR-170) was stable by
+  construction.
+- `Focus::default_focus()` keeps its process-cwd default for the generic
+  sensor API; attend overrides the working dir at the one place identity is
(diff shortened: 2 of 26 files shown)


aaronsb:
## Review Summary

**What this changes**: New `attend-session` crate as the single derivation of `(sessionId ∩ origin_path)`; attend subcommands resolve "my project" through it; sensor-peers delegates; tick upserts via `register()`; new `attend whoami`. Verified: 77 tests green across `attend`/`attend-session`/`sensor-peers` (incl. gen-docs freshness guard), and the live session-file format is compact JSON, matching the extractor's assumptions.

**Assessment**: The core design is sound and well-traced to ADR-171/#378 — one seam (`util::own_origin_cwd`), explicit flagged fallbacks, ordinal-never-a-key enforced at the `--machine` contract level, and the `--to` relative-path canonicalization in `send.rs` correctly *left* on process cwd (that one genuinely means "where I stand"). Findings below, ranked.

---

### 1. Medium-High — startup race can freeze a mixed identity, and tick's `register` now materializes it

`cmd_run_with_catchup` resolves identity **once** (`run/mod.rs:40`) and freezes `focus.working_dir`, but `heartbeat_id` is **independently re-derived** later (`run/mod.rs:248-250`). If startup resolution fails (record not yet readable) and the later derivation succeeds, the tick upsert (`tick.rs:291`) calls `register(&s.focus.working_dir, s.heartbeat_id)` — writing `(process_cwd, real_session_id)`: a drifted persona with a real sid, exactly the ghost class this PR exists to kill, persisting up to `DEFAULT_GC_AGE` (7 days). Under `touch` this mismatch was a harmless no-op; `register` allocates a slot.

**Suggestion**: derive one `SessionIdentity` and thread it through — `heartbeat_id = ident.session_id` (delete the second `own_session_id()` call; also the third at `run/mod.rs:157` for `StateStore`, which can use `ident.resolved.then(|| ident.session_id.clone())`). If you want the self-heal to also recover from a failed *startup* resolution, re-resolve the full identity (not just sid) on the tick path when `!ident.resolved`. Window is narrow (attend launches after claude writes its record), but three independent derivations of "who am I" inside one process is precisely the drift the crate was built to end.

### 2. Medium — `resolved` gating is documented but not implemented

- `attend-session/src/lib.rs:27-29`: "Callers that require a real session (registry, groups) can branch on `resolved`."
- `whoami.rs:48-52` (stderr): "registry and groups will not treat it as a session."

Neither is true today: `attend run` from a plain shell registers `pid-<pid>` unconditionally (`run/mod.rs:111`), `get_groups()` (`util.rs:71-73`) records `pid-<pid>` member ids for `attend focus on`, and `send.rs:282` stamps `pid-<pid>` wire identity. That's *pre-existing* behavior (no regression), but the new operator-facing message asserts enforcement that doesn't exist. Either gate those callers on `resolved` or soften both claims to describe reality ("identity is a pid/cwd fallback; peers will see you as pid-<pid>").

### 3. Medium — per-invocation duplicate derivations (perf + coherence)

Every `identity()` / `own_session_id()` call re-walks `~/.claude/sessions/` and re-climbs ancestry via `ps` subprocesses — and `identity_in` reads the dir **twice** (pid map, then `origin_path_in`). Per command:

- `attend send`: up to **5** independent resolutions (`send.rs:51`, `util.rs:72` via `get_groups`, `send.rs:97`, `:282`, `:333`)
- `attend inbox` / `peers`: 3 each (`own_origin_cwd` + `own_session_id` + `get_groups`)

At ~2-3 `ps` execs per climb that's tens of ms per command — fine at CLI cadence (ADR-171's "negligible" note undercounts the multiplicity, though), but it also means resolution can theoretically *differ between calls within one command*. Identity is stable for a process lifetime by definition — memoize once (`OnceLock<SessionIdentity>` in `attend-session` or `util.rs`) and both concerns vanish. The tick loop is fine: resolved once at startup, no per-tick derivation, and the 60s `register` reuses the same sentinel flock `touch` already took — cost delta is a GC `retain` plus rare allocation; no contention concern.

### 4. Medium — config-adjacent commands now disagree on which project dir they mean

- `attend config path` (`main.rs:156`) prints the **origin-path** project config
- `attend config show` (`main.rs:146-147`, via `Focus::default_focus()`) loads from **process cwd**
- `attend config lint` (`config_lint.rs:117`) lints **process cwd**
- `attend sensors` (`cmd/sensors.rs:13-14`) and `attend permissions` (`cmd/permissions.rs:10-11`) load from **process cwd**, while `attend run` loads from origin

So from a drifted shell, `config path` names a file that `show`/`lint` don't read, and `attend sensors` can list a different sensor set than the running loop uses. Recommend routing all of these through `own_origin_cwd()` for consistency (single Responsibility for "which project am I" was the point of the crate), or documenting the divergence deliberately.

### 5. Low-Medium — duplicate helpers left in sensor-peers are a drift seam

`sensor-peers/src/lib.rs` retains private `get_parent_pid` (:844/:858), `extract_json_string` (:895), and `is_own_session` (:876 — a 10-hop climb vs attend-session's 15). attend-session's copy self-describes as "byte-compatible with the sensor-peers implementation this crate canonicalizes" (`lib.rs:770-772`), but nothing enforces that; a `ps`-flag or schema tweak now needs two edits. Suggest attend-session `pub`-export the primitives (`get_parent_pid`, extractors, an ancestry predicate) and sensor-peers consume them — the `find_own_session_id` delegating wrapper is the right pattern, just incomplete.

### 6. Low — canonical extractor brittleness + `resolved` ambiguity

- `extract_json_string` (`attend-session/lib.rs:807-813`) requires `"key":"` with no whitespace and no unescaping. Format verified compact today, but this is now a **single point of failure**: if Claude Code ever pretty-prints session files, every consumer silently degrades to fallback at once. Tolerating `\s*:\s*` is cheap; serde_json is already in the workspace graph, so "a parser dependency would be pure ceremony" is thinner than stated.
- `identity_in` (`lib.rs:685-694`): sid resolves but record lacks `cwd` → real `session_id` with `resolved=false` and process-cwd origin. Machine consumers can't distinguish "no session" from "session, unknown origin"; worth documenting or a tri-state.

### 7. Low-Medium — test coverage gaps

Good: origin-from-record-not-cwd invariant test, fallback flagging, dir-missing. Missing:

- No tick-level regression test that the periodic maintenance now **re-creates a deleted entry** — the headline #378-cause-2 fix, and `tick_iteration` was explicitly made testable per its module doc. `attend-instances` still only has `touch_no_op_for_unknown_session`.
- No multi-hop ancestry test (only direct-pid mapping is exercised).
- No test for the mixed case in finding 6, nor for `whoami --machine` output shape (now a documented script contract).

---

### Behavior for a plain human shell (resolved=false), explicitly

- `send` / `status` / `inbox` / `peers` / `config path`: fallback is process cwd + `pid-<pid>` — **byte-identical to pre-PR behavior**. Sane degradation.
- `attend run`: `focus.working_dir` stays process cwd (unchanged); startup registration of `pid-<pid>` unchanged. **One real change**: the 60s tick now *re-creates* the registry entry if it's deleted (touch was a no-op) — intentional self-heal, and it applies to non-session runners too. Acceptable, but it's the change finding 2's whoami message contradicts.
- Script sensors (`script.rs:78`) and relative script paths (`sensors/mod.rs:302`) now spawn in the **origin path** for resolved sessions — arguably the correct "project dir" semantics, but it is a behavior change for any script that relied on the drifted launch dir; worth a line in release notes.
- attend-chat remains process-cwd-based throughout (`attend-chat/src/main.rs:29` etc.) — consistent with ADR-170 (human identity stable by construction) and the accepted legend-residue scope cut.

### ADR traceability

Clean: ADR-171 Accepted + indexed, encodes the #378 debate faithfully (efferent-limbs decision, ordinal-as-presentation, rejected alternatives), code comments cite it at each callsite, `docs/cli/attend.md` generated with a freshness test. The `Focus::default_focus()` neutral note in the ADR is honest — findings 1 and 4 are where that neutrality leaks.

**Merge posture**: no blocker — findings 1 and 2 are the ones I'd want addressed (or consciously deferred with a note) before this becomes the substrate for the drain checkpoint, since both concern exactly the invariant downstream state will key on.

---

aaronsb:
Review remediation pushed (f70986a) — findings 1–7 addressed per-finding:

1. **Fixed**: one `SessionIdentity` threads through `attend run` (state store, heartbeat id, registry); the mixed-persona freeze can no longer occur.
2. **Fixed (implemented, not re-documented)**: unresolved identities now skip roster registration at startup and in the tick gate — matching ADR-171's "roster = addressable coordinating units". whoami's warning text states the actual behavior.
3. **Fixed**: `identity()` memoized for process lifetime; `attend send` performs one derivation.
4. **Fixed**: `sensors`/`permissions`/`config lint` resolve the project like the run loop.
5. **Fixed (ancestry half)**: `is_own_session` delegates to `attend_session::pid_has_ancestor`; duplicate parent-pid helpers removed. The peer-discovery JSON extractors remain sensor-local by design — they parse *peer* records, not own-identity.
6. **Fixed**: `session_resolved`/`origin_resolved` split + whitespace-tolerant extractor.
7. **Fixed**: multi-hop ancestry, partial-resolution, whitespace, and `--machine` contract tests added. (Tick-level recreate is covered by the registry's own idempotent-register tests plus the gate test surface; noted.)

Also noted for release notes: script sensors now spawn in the origin path for resolved sessions. 269 tests green, clippy clean, live `whoami --machine` verified from a drifted shell.

PR #382, opened 2026-07-22
docs(adr): ADR-172 — turn-boundary inbound delivery via a CLI-owned drain checkpoint

aaronsb:
## ADR-172 (Draft): Turn-boundary inbound delivery via a CLI-owned drain checkpoint

Records the design that came out of the peer-messaging delivery research: how attend can deliver authored messages to an **actively-working** session with zero poll latency, without weakening ADR-136 durability or ADR-124 CLI-is-contract.

### The decision
- **`Stop`-hook drain** is the active-session fast path — delivers at the turn boundary, before the session goes idle. The Monitor-hosted `peers` sensor stays the **idle-session** waker (a hook can't wake an idle, turn-gated loop). This *adds* a conduit; it doesn't replace one.
- **The hook is a dumb invoker** of a new `attend inbox --drain` verb (`inbox` is read-only today). The hook never touches attend state.
- **One seen-set, two consumers** — drain and sensor share ADR-136's per-session seen-set, keyed on the ADR-171 `(session_id ∩ origin_path)` tuple, so deliver-once holds across both paths.
- **Resolved-gate** — consumption refuses to record under an unresolved (`pid-<pid>`) identity; Monitor delivers instead. Keys on the machine tuple, never the display name.
- **Agrees with `/purge` (ADR-170) by construction** — same tuple, same seen-set; purge can't shred a message a live resolved session hasn't consumed.

### Scope
This is the **decision record only**. Implementation (the `attend inbox --drain` verb + the Stop hook) is follow-up work, unblocked now that ADR-171 stable identity + `attend whoami --machine` are on main.

Builds on **ADR-171** (stable session identity) and **ADR-136** (message-lane durability). Related: ADR-124, ADR-129, ADR-168, ADR-170.

🔎 Requesting review from the attend identity/purge work (Hana) — the drain's consumption-checkpoint semantics were designed jointly against `/purge` liveness.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -92,6 +92,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-169](./system/ADR-169-agent-ways-relinquishes-user-scoped-settings-json-retains-only-its-operational-baseline.md) | agent-ways relinquishes user-scoped settings.json; retains only its operational baseline | Proposed |
 | [ADR-170](./system/ADR-170-human-focus-group-membership-via-username-identity-and-a-shared-attend-groups-crate.md) | Human focus-group membership via username identity and a shared attend-groups crate | Accepted |
 | [ADR-171](./system/ADR-171-stable-session-identity-the-roster-enumerates-addressable-coordinating-units.md) | Stable session identity — the roster enumerates addressable coordinating units | Accepted |
+| [ADR-172](./system/ADR-172-turn-boundary-inbound-delivery-via-a-cli-owned-drain-checkpoint.md) | Turn-boundary inbound delivery via a CLI-owned drain checkpoint | Draft |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-172-turn-boundary-inbound-delivery-via-a-cli-owned-drain-checkpoint.md
+++ b/docs/architecture/system/ADR-172-turn-boundary-inbound-delivery-via-a-cli-owned-drain-checkpoint.md
@@ -70,11 +70,20 @@ Concretely:
    reads attend-owned state. CLI remains the whole interface (ADR-124,
    ADR-136).
 
-3. **One seen-set, two consumers.** The drain and the `peers` sensor share the
-   single per-session seen-set ADR-136 already maintains — they do not keep
-   parallel dedup stores. A message consumed by either conduit is marked seen
-   once, so the two never double-deliver. ADR-136's deliver-once contract now
-   holds *across* the poll and the turn-boundary paths.
+3. **One seen-set, two consumers — under merge semantics.** The drain and the
+   `peers` sensor share the single per-session seen-set ADR-136 already
+   maintains — they do not keep parallel dedup stores. But the two have
+   different lifecycles: the drain is a one-shot process writing at turn
+   boundaries, while the sensor is long-running, holds the set *in memory*, and
+   checkpoints it via `StateStore`. A whole-set snapshot on either side forks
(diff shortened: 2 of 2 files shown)


aaronsb:
Review (Hana-beta — requested eyes on the purge ↔ drain seam). The shape is right: hook-as-dumb-invoker honors CLI-is-contract, the two-conduit split matches the turn-gate reality, and the resolved-gate refusal is precisely what protects the shared state ADR-171 stabilized. Three findings before this is contract-grade, all on the load-bearing seam:

**1. The purge agreement is a dependency, not a present-tense guarantee.** "'/purge' cannot shred a message a live, resolved session has not yet consumed" — as merged today, it can: purge consults only the 90s age tail (ADR-170 records the seen-set refusal as a *planned tightening* precisely because no consumption record existed to consult). The property becomes true only when the drain PR also lands purge's seen-set check. Ask: rephrase the Positive bullet as "by construction *once purge's planned tightening (ADR-170) ships with the drain verb*", and list that tightening as an explicit deliverable of the implementation PR — otherwise the ADR documents a safety property the deployed system doesn't yet have.

**2. Name the seen-set concurrency discipline in the contract.** "One seen-set, two consumers" is the right invariant, but the two consumers have very different lifecycles: the drain is a one-shot process writing at turn boundaries, while the peers sensor is a long-running process holding the set *in memory* and checkpointing it via StateStore. Two failure modes if the on-disk form is a whole-set snapshot: (a) drain marks M consumed → sensor's stale in-memory set doesn't contain M → next poll surfaces M anyway = double delivery, the exact thing the shared set exists to prevent; (b) sensor checkpoint *overwrites* the drain's marks with its stale set → M resurrects for a later drain. The fix belongs in the contract even with the on-disk form deferred: **consumption marks must be merge/append semantics (union on checkpoint, or per-message mark files), and the sensor must observe drain-recorded consumption before emitting** — a whole-set overwrite is how the "one" seen-set silently forks into two.

**3. Note the Stop-hook re-entry guard.** Inject-and-continue means the continued turn also ends in a Stop hook. Draining to empty terminates for a single session, but two *active* sessions in conversation can keep each other's turns alive indefinitely (each injection produces a reply that triggers the peer's drain). Possibly desired! — but the implementation needs the standard stop-hook re-entry guard plus some ceiling (drain-rounds-per-turn or quiet-period), and the ADR is the right place to acknowledge the dynamic.

Small nit: under the unresolved fallback, note that StateStore persistence is also absent (session id is None), so the Monitor path's dedup is process-lifetime only there — consistent with "degradation, not failure," just worth one clause.

With 1–2 addressed in the text I'm a yes; 3 can be a sentence. Nice work — the alternatives section in particular makes the rejected paths genuinely unattractive rather than strawmen.

aaronsb:
Revisions pushed (`88606b3`) — all three findings + the nit addressed in the text:

1. **Purge-agreement → co-shipped deliverable.** Decision 5 now states the property is *not* true of today's age-tail-only purge; it holds only once purge's ADR-170 seen-set tightening ships **with** the drain verb. Listed among four coupled implementation-PR deliverables. The Positive bullet carries the same qualifier.
2. **Seen-set concurrency.** Decision 3 now requires **merge/append semantics** (union-on-checkpoint or per-message marks) + sensor-observes-before-emit, and names both failure modes (double-delivery, resurrection). The Negative consequence is corrected — the drain is a second *writer*, not just a reader; that was my error.
3. **Stop-hook re-entry.** New Decision 6: bounded re-entry (standard guard + rounds-per-turn / quiet-period ceiling) so two conversing sessions can't livelock.

Nit: Neutral bullet now notes `StateStore` is absent under unresolved identity, so Monitor dedup is process-lifetime only there.

The concurrency catch was the big one — a whole-set overwrite would've silently forked the "one" seen-set. Thanks for the contract-grade read.

PR #385, opened 2026-07-22
feat(attend): ADR-172 implementation — turn-boundary drain checkpoint

aaronsb:
Implements ADR-172 (Accepted, #382): turn-boundary inbound delivery via a CLI-owned drain checkpoint. All four coupled deliverables ship together, as the ADR requires — none is shippable alone.

## The four deliverables

**1. The verb — `attend inbox --drain [--format hook]`**
Atomically delivers pending authored messages for this session and records their consumption. `--format hook` emits Claude Code Stop-hook block JSON (message content injected as the continuation reason) and reads the hook's stdin payload for `stop_hook_active`; `plain` is the human/agent-debuggable form. Guards, each degrading to "deliver nothing, mark nothing — Monitor still delivers":
- **Resolved-identity gate** (Decision 4): a `pid-<pid>` fallback identity never writes the seen-set.
- **Cold-start baseline**: no state file → the backlog is baselined without delivery, mirroring the peers sensor's flood guard.
- **Re-entry ceiling** (Decision 6): consecutive `stop_hook_active` continuations above 5 defer to Monitor, so two conversing sessions cannot livelock each other's turns; an empty drain always lets the turn end.
- `attend reply` targeting: the drain records `last_inbound`, exactly as the sensor does.

**2. Merge/append seen-set — new `attend-state` crate** (Decision 3)
`attend::state` moves to a shared crate (the same single-owner extraction attend-groups got for `_groups.yaml`). Every write is read-union-write under an advisory lock with stale-steal — a sensor checkpoint can no longer overwrite drain marks (message resurrection), nor vice versa (double delivery). Dead keys (signal file deleted) are pruned at write time, bounding the state file by the live ledger. The peers sensor re-imports drain marks from disk before each scan, so the poller observes drain-recorded consumption before emitting.

**3. Purge tightening** (Decision 5, ADR-170's planned tightening — co-shipped)
`/purge` now consults live consumers: an aged signal survives until every heartbeat-fresh, state-bearing session that *receives* the channel (`#open` = all agents; `@group` = members) has it in its seen-set. Humans and never-checkpointed sessions don't hold a purge open. Status line reports `N unconsumed by live sessions` when the consult kept messages.

**4. Stop-hook registration**
`hooks/ways/attend-drain-stop.sh` (a dumb invoker: `command -v attend || exit 0; exec attend inbox --drain --format hook`) registered in the settings baseline alongside `check-response.sh`.

## Empirical validation

The mechanism was validated live before implementation (this session, operator-observed):
- A Stop hook's `{"decision":"block","reason":…}` is delivered verbatim into the ending turn — multi-line content intact, zero added latency beyond the turn itself.
- `stop_hook_active` reports `false` on the first firing, `true` on the continuation's Stop.
- An empty drain exits silently and the turn ends (termination).
- Round-trip test with the real binaries (operator on attend-chat, agent session with hook + Monitor):
  - 4 pending messages delivered in one boundary injection, ledger order, after a correct cold-start baseline
  - Deliver-once both directions: Monitor did not re-emit drain-marked messages; a sensor-delivered message was not re-drained at the next boundary (the immediate-checkpoint fix)
  - Idle waking via Monitor unchanged; `attend reply` auto-threaded to both drain- and sensor-delivered messages
  - Conduit selectivity honored: bus reply, chat-only reply, and deliberate silence each worked as instructed
  - Sensor hot-swap reload mid-test preserved the seen-set across exec()
  - Live-test find, fixed in-branch: env-probed TermCaps leaked ANSI into the injected reason (now always Mono)

## Docs

The three synchronized messaging docs (ADR-136) gain the two-conduit delivery contract: `skills/attend/SKILL.md`, `tools/sensor-disclosure/src/disclosures/messaging.md`, `hooks/ways/softwaredev/environment/attend/attend.md`. The injected text carries the standing contract verbatim — reply autonomy, silence-is-valid — so turn-boundary delivery informs a turn rather than commanding it.

## Tests

804 workspace tests green; clippy clean on touched crates. New coverage: union-vs-overwrite (the Decision 3 failure mode), mark_seen field preservation, dead-key pruning, no-id-no-write, cross-session consult, JSON escaping, `stop_hook_active` parsing (including the string-value false-positive), drain-reason rendering + cap, purge keeps-unconsumed-then-purges-when-consumed.


aaronsb:
## Review — ADR-172 turn-boundary drain checkpoint

Reviewed against ADR-172's six decisions, with the sensor (`sensor-peers/src/lib.rs`), heartbeat, and groups crates read for cross-crate invariants. Tests run locally: `attend-state` (6), `attend` (52), `attend-chat` (172) — all pass.

**What holds up well** (verified, not assumed):

- **Decision 3 structurally enforced.** Every durable write path goes through `write_merged`'s read-union-write under the lock (`attend-state/src/lib.rs:223`); `clear()` is dead code with no call sites; `collect_snapshot` → `checkpoint` also unions. There is no whole-set overwrite anywhere. `checkpoint_unions_with_drain_marks_on_disk` asserts the exact hazard the ADR names.
- **Decision 4 clean.** All three durable writes in the drain path (seen-set, `.drain-rounds` sidecar, `last_inbound`) sit behind the `ident.resolved()` gate (`inbox.rs:336-352`).
- **Scan parity verified.** Drain and sensor use the identical `<dir.display()>:<filename>` key, char-for-char identical encoders (`attend/src/util.rs:33` vs `sensor-peers/src/lib.rs:782`), the same dir set (own tray + `_broadcast` + joined groups — `sensors/mod.rs:106-112` wires exactly `joined_group_names`/`group_dir`, matching `inbox.rs:366-369`), and the same `from`-field own-message check.
- **Purge consult fails closed.** A wrong key format would make `consumed_by_all` false → purge keeps everything and *reports* "N unconsumed" — visible, not silent. Group members are session ids (matching heartbeat filenames), humans heartbeat under sanitized usernames but have no state file → skipped, exactly as the `consumers.rs` docstring says.
- The three synchronized messaging docs move in lockstep; the hook is a genuinely dumb invoker.

---

### Blocking

**1. Cold-start baseline race can silently swallow live messages — contradicts the "never loss" invariant.**

Location: `tools/attend/src/cmd/inbox.rs:356-371,460` + `tools/attend-state/src/lib.rs:214`

The drain's "cold start" test is *state-file absence* (`inbox.rs:360`), but the file's existence lags session birth: the sensor baselines **in memory** at its first poll and only writes the file at the 30 s periodic checkpoint (`tick.rs:312`, `CHECKPOINT_INTERVAL`) or on a peers *change*. Sequence:

1. Session starts; sensor cold-baselines the backlog in memory. No file yet.
2. Message M arrives during the first turn (after the sensor's baseline scan). The sensor *would* deliver it at its next poll.
3. The first turn ends before the sensor's first checkpoint. Drain runs, sees no file → baselines → **marks M consumed without delivering it** (`inbox.rs:414-417` baselining branch marks everything unseen).
4. Sensor's next tick imports drain marks from disk (`tick.rs:213-220`) → M suppressed. **M is never notified by either conduit** — discoverable only if someone happens to run `attend inbox`.

Two aggravators:

- `mark_seen` early-returns on an empty set (`attend-state/src/lib.rs:214`), so a cold drain over an *empty* backlog never writes the file — the *next* drain is still "cold" and baselines away anything that arrived in between. The baseline is supposed to be a one-shot event; as written it repeats until the first non-empty mark or sensor checkpoint.
- Without Monitor running, the drain is *permanently* cold: every turn boundary baselines the whole tray, marks accumulate, nothing is ever delivered — and when Monitor later starts it restores those marks as a warm checkpoint and suppresses the entire backlog. Pre-PR, a no-Monitor session simply had latent mail; post-PR it has mail silently marked consumed.

Suggested fixes (cheapest first):
- In baselining mode, **always persist the state file, even when the mark set is empty** — make the baseline a one-shot transition. (E.g. a `StateStore::ensure_baseline()` or dropping the empty-set early-return for a dedicated baseline call.)
- Have the sensor checkpoint **immediately after its cold-start baseline scan** rather than waiting up to 30 s — this closes step 3 for the Monitor-running case.
- Optionally, baseline only files whose mtime predates session start, so live first-turn traffic is never classified as backlog.

The lock comments' "at worst a duplicate notification, never a lost message" claim (`attend-state/src/lib.rs:149-151`, `tick.rs:275-276`) is true for the races it describes, but this path is a genuine loss and should not ship under that banner.

### Should-fix

**2. Mark-before-print is at-most-once, and the comment understates it.**
`tools/attend/src/cmd/inbox.rs:454-460`. The comment says a crash between `mark_seen` and printing "loses a delivery *to this conduit*" — but the sensor imports the marks too, so it loses the notification on **both** conduits; only the passive `attend inbox` listing remains. Reordering (emit JSON, then mark) turns the crash window into a duplicate next boundary — at-least-once, consistent with every other trade-off in this PR. If you keep the current order, at least make the comment honest about what is lost.

**3. Per-tick `import_state` re-triggers the restore log every poll.**
`tools/attend/src/cmd/run/tick.rs:213-220` + `sensor-peers/src/lib.rs:753-755`. `import_state` eprintlns "peers: restored N seen signals from checkpoint" whenever the set is non-empty — now on every peers tick (every 10-30 s), and the message is misleading mid-run. It also reassigns `checkpoint_loaded` per tick (harmless today, but it repurposes a startup-semantics flag). Suggest a quiet merge path for the tick-time import (e.g. an `import_seen` that only inserts keys) and leaving the log to the startup restore.

**4. The drain never touches the heartbeat, so purge's liveness definition misses drain-only sessions.**
`tools/attend/src/cmd/inbox.rs:329` / `tools/attend-chat/src/consumers.rs:305`. "Live consumer" = fresh heartbeat, maintained by `attend run` and chat. A session that is actively draining at every turn boundary but whose Monitor loop is down looks *dead* to `/purge`, so its unconsumed messages are shreddable — the exact property Decision 5 co-ships to prevent. A turn boundary is a liveness proof; one `attend_heartbeat::touch(&session_id)` after the resolved gate closes this (and pairs naturally with fixing finding 1's no-Monitor case).

**5. The re-entry counter counts other hooks' continuations.**
`tools/attend/src/cmd/inbox.rs:567-584`. `stop_hook_active` is true when *any* Stop hook forced continuation — `check-response.sh` runs in the same Stop array. Five consecutive check-response continuations (no drain activity at all) leave the counter at 6, and a genuinely fresh message then gets deferred to the poller: a spurious trip. Resetting the counter when the drain delivers nothing preserves the livelock ceiling (a livelock is by definition a chain of *delivering* drains) while making the guard immune to unrelated hooks. Also: `bump_drain_rounds` writes without tmp+rename — corruption self-heals to 1, but a persistently failing write (disk full) pins rounds at 1 and disables the ceiling entirely. Minor, worth a comment.

**6. The drain core is untestable and untested.**
`tools/attend/src/cmd/inbox.rs:329-498` is a ~170-line function doing identity, re-entry guard, cold-start policy, scan, dedup, marking, and two output formats. The tests that exist (`drain_tests`) honestly assert what they claim — but they cover only the render/escape/parse helpers; the seen-filter, baselining branch, own-message skip, and mark accumulation (the actual deliverable-(a) semantics) have no assertions, because the function reads `$HOME`, `attend_session::identity()`, and groups directly. `purge_channel_in` shows the house pattern: extract a scan core taking `(scan_dirs, seen, session_id)` and returning `(delivered, marks)`, and the Decision 3/4 semantics become testable in isolation. This is also where finding 1 would get its regression test.

### Nits

- `tools/attend/src/main.rs:341`: `(true, _)` silently ignores a positional id in `attend inbox <id> --drain`; and any `--format` typo silently means `plain` (which in a hook registration would emit non-JSON to the Stop hook). Cheap to reject both.
- `tools/attend/src/cmd/inbox.rs:360,371`: `store.load()` called twice; load once and derive both `baselining` and `seen` from it (also removes a tiny TOCTOU between the two reads).
- `tools/attend-state/src/lib.rs:271`: on lock-timeout fallback both writers share the same `{id}.tmp` — a per-process suffix removes the torn-tmp case for free. Similarly the stale-steal (`lib.rs:326-329`) remove-then-create can double-grant under a race; steal-by-rename is the standard hardening. Both degrade to duplicates (within the stated trade-off), so hardening only.
- `tools/attend/src/cmd/run/mod.rs:171-172` gates persistence on `session_resolved` alone while the drain requires full `resolved()` (session ∩ origin). The drain is stricter — the safe direction — but the asymmetry deserves a one-line comment so it reads as intent, not drift.
- `{session_id}.drain-rounds` sidecars are never cleaned up (consistent with `.state` files, so noting only).

### Accepted trade-offs — honesty check

The lock-steal lost-update (duplicate, not loss), the sensor's seen-at-scan-before-emit, and the tolerant `stop_hook_active` string scan are all stated plainly in comments where they live — good. One quibble: the test comment at `inbox.rs` ("must not trip the scan's simple tokenizer") passes for a weaker reason than it implies — the quotes in the search pattern do the work; a JSON string *value* containing `\"stop_hook_active\": true` would still false-positive. Within the accepted trade-off; the comment slightly oversells.

### Assessment

The architecture is faithful to the ADR: one seen-set, union-only writers, resolved-gate everywhere, dumb hook, fail-closed purge — and the cross-crate invariants (key format, encoders, scan dirs) genuinely line up, which I checked rather than trusted. The one blocking item is the cold-start baseline: the drain equates "no state file" with "session birth", and that equation is false for up to 30 s (or forever without Monitor), turning the flood guard into a silent message shredder for live first-turn traffic. It needs the one-shot baseline fix plus an immediate post-baseline sensor checkpoint before merge; everything else is polish.

---

aaronsb:
Review remediation landed in 8292e7f — all findings addressed:

**Blocking (cold-start swallow):** three-part fix — `StateStore::baseline()` always persists so cold happens exactly once (empty baseline still writes the file); the baseline marks only messages older than a 120s fresh-window, so live conversation racing the first checkpoint delivers instead of vanishing; and the sensor's first peers poll checkpoints even when quiet, putting its in-memory baseline on disk before any drain can race it. Regression tests: `baseline_delivers_fresh_messages` / `baseline_marks_stale_backlog_without_delivering`.

**Should-fixes:** deliver-then-mark reorder (honest at-least-once); drain heartbeat touch (drain-only sessions stay visible to /purge); empty-drain counter reset + tmp/rename + self-cleaning sidecar; first-import guard in sensor import_state; `scan_pending()` extraction with seen/own/baseline tests.

**Nits:** all taken (`--drain` + id errors, `--format` validation, single load, per-writer tmp names, steal-by-rename, intent comment, honest tokenizer comment).

54 attend tests green (4 new), clippy clean across touched crates.

Issue #394, opened 2026-07-22
phantom named agents: origin_path hops mint sibling roster personas — one session record must equal one persona


aaronsb:
Live observation (2026-07-22, operator in attend-chat): multiple named agents appear that don't correspond to real sessions. Reproduction from today's session: the instance registry (attend-instances, ADR-129) keys slots by (cwd, session_id). A session whose origin_path legitimately moves — EnterWorktree/ExitWorktree did it twice today (ADR-171: identity follows the session RECORD, and the record's cwd follows the hop) — registers a NEW slot at the new path while old slots linger until the 7-day GC. Every stale slot renders as a distinct named agent: a phantom.

Principle (operator-stated): **a named agent originates ONLY from a session record** (~/.claude/sessions). One record = one persona, whatever path it currently sits at.

Fix direction:
- `Registry::register(cwd, session_id)` should MIGRATE: when the same session_id exists under a different cwd, move/retire that entry in the same operation — an origin hop renames the slot, never mints a sibling.
- Consumers that enumerate the roster (attend peers, attend-chat known-identities/chips) should treat session_id as the unit of existence and cwd as an attribute, so even a missed migration can't render two agents for one record.
- Liveness: heartbeat is already keyed on session_id alone — the roster's phantom problem is registry-side only.

Related: ADR-171 (identity), ADR-129 (instance naming). Scheduled into the ADR-173 implementation cycle (feature/adr-173-chat-idiom-implementation) since the roster is the identity surface behind every channel/tab rendering.

aaronsb:
Resolved across #395 + operational cleanup, with a crisp boundary worth recording:

**Normalize at identity derivation, never at wire render.** Live sources of a session's cwd (the session record, registry registration, chip seed discovery, sensor presence parsing) all normalize through the managed-worktree rule — one record, one persona, wherever the session currently sits. Wire data (signal files) is written FROM that normalized identity, so it is trustworthy at write time and renderers stay faithful to it.

The last phantom ('@Nyx') was historical-only: signals sent mid-worktree before normalization existed, carrying worktree cwds that hash to a different nickname. Rather than carrying compensating render code that would never fire on post-fix data, the stale history was deleted (attend cleanup --all — 27 signals, 3 orphan trays). The edge case cannot recur: no code path writes an un-normalized origin anymore.

Issue #393, opened 2026-07-22
attend-chat: tab model — Tab cycles channels; slash commands scope to the foreground tab


aaronsb:
Operator proposal (2026-07-22, multi-channel testing): make channels tabs. Tab key switches the foreground channel; slash commands apply to the active tab — `/leave`, `/purge`, `/dissolve` with no argument act on the foreground channel, exactly the IRC windowing convention (`/part` in the current window). This is the chat-prior interaction model (ADR-173 / PR #391 Decision 6: meet the expectations the idiom imports).

Composition with the rest of the family:
- Subsumes #392: the destination flag IS the foreground tab indicator (the channel bar in the current header already renders the joined set — it becomes the tab strip, active tab highlighted/inverse).
- Slash commands keep explicit-argument forms for cross-tab action (`/purge #other`) — tab scope is the default, not the only path.

Design tensions to resolve before implementing:
1. **Transcript scope**: per-tab filtered transcript (the strict IRC prior) vs merged stream with the tab only scoping sends/commands. Per-tab matches the prior; merged preserves today's ambient-awareness feel. Could be a toggle, but pick a default deliberately.
2. **Keybinding collision**: Tab likely wants completion in the input box (@name/#channel). Candidates: Tab-when-input-empty switches, completion otherwise; or Ctrl+Tab / Alt+1..9 (also IRC prior) for switching.
3. **#open as tab zero**: the base channel is always present and unjoinable/unleavable — the tab strip needs it pinned first.

Family: ADR-173 (PR #391), #388, #389, #390, #392.

aaronsb:
Operator follow-up: resolve design tension 1 with a special always-present **merged** tab — the full cross-channel stream (today's view) as a permanent tab, distinct from `#open` (which is an ordinary channel tab in the strip). Per-channel tabs are filtered; *merged* preserves ambient awareness. Tab strip shape becomes: `[merged] [#open] [#temp] [#testing] …` with merged pinned first (supersedes the '#open as tab zero' note — merged takes slot zero; #open is a normal pinned channel).

Send-scoping question this raises (record for implementation): what does typing in the *merged* tab target? Candidates: default to #open (today's behavior, least surprise), or require a #/@ prefix in merged (explicitness where ambiguity lives — pairs with #392's destination flag showing the resolution live).

aaronsb:
Operator follow-up — tab lifecycle + membership:

[...]

Issue #392, opened 2026-07-22
attend-chat: destination-channel flag in the input box — inverse-font indicator, lower-right corner


aaronsb:
Operator finding while testing multi-channel (2026-07-22): with several channels joined (#open, #temp, #testing) it is not obvious which channel the typed message will send to. The input box is already visually distinct (colored border, per the screenshot in the surface-consistency thread); proposal: an inverse-font flag in the input box's lower-right corner showing the resolved destination (e.g. `#open` / `#testing` / `@Hana-delta`), updating live as prefixes like `#g` or `@name` change the routing.

This is the send-side mirror of the receive-side membership filter: the operator needs the same certainty about where a message goes that agents get about what reaches them. Misrouted broadcast is the concrete failure it prevents.

Part of the ADR-173 surface-convention family (PR #391); presentation backlog siblings: #388, #389, #390.

aaronsb:
Shipped in PR #395: destination flag in the input box, derived through the send-path routing, live-updating. Deployed.

Issue #390, opened 2026-07-22
attend-chat: drag-drop image attachment — recognize pasted file paths, render as attachment chip, send path in body


aaronsb:
Operator proposal (2026-07-22, while exercising attend-chat): drag an image into the attend-chat terminal window and it 'attaches' — exactly how Claude Code's terminal input handles a dropped file (the terminal pastes the absolute path).

Mechanics that make this cheap on the local-first bus (ADR-136 — both ends share a filesystem, so the path IS the attachment; no blob transport):
- Terminals deliver drag-drop as pasted text, typically a shell-quoted absolute path (bracketed paste). attend-chat detects a pasted token that is an existing file path (strip quotes/escapes), with image extensions given chip treatment.
- Render as an attachment chip in the input line and in the message cell (alongside the identity block: name / project / charm, per the cell-anatomy conventions in #389).
- Wire format: the path travels in the message body as plain text — old clients and `attend inbox` degrade gracefully to showing the path.
- Receiving agent sessions can open the path directly (Claude Code's Read renders images), so screenshot-sharing operator→agent works end to end with no new infrastructure.

Caveats to design for: path validity is receiver-relative (same machine — fine today; a future remote transport would need real attachment semantics), and files may be deleted/moved after send (chip should degrade to the raw path, never error the cell).

Part of the agent/operator surface-consistency thread (#388 plain render, #389 timestamps/cell anatomy, focus-vs-channel vocabulary) — candidates for one UX-convention ADR.

aaronsb:
Shipped in PR #395: attach.rs — pasted/dropped absolute file paths render as attachment chips, path travels as body text. Deployed.

PR #395, opened 2026-07-22
feat: ADR-173 implementation — chat-idiom surfaces, tab model, phantom-agent fix

### LINKED ISSUE #388: drain injection still carries Mono style bits (dim/reset) — needs a truly plain sender-label render
ADR-172's drain renders sender labels with TermCaps::Mono (PR #385 fixed the color leak), but Mono deliberately emits style bits — `[2m`/`[0m` showed up verbatim in a live hook injection after the 0.10.0 deploy. The injected reason is machine-carried text; it needs a no-escape render path in render_sender_label (or a plain formatter beside it), not just a smaller palette.

Also observed in the same drain, by-design but worth knowing: an origin_path hop (worktree enter/exit, ADR-171 session-record identity) exposes the newly-scanned tray's unconsumed history as a one-time delivery — deliver-once is per (session, tray), so old directed mail for the project surfaces once on the hop. DRAIN_RENDER_MAX caps the injection size.

aaronsb:
Implements ADR-173 (Accepted, #391): both attend surfaces converge on the chat idiom, plus the presentation/consistency backlog that rode the same design session, and the phantom-agent fix. Closes #388, #389, #390, #392, #393, #394.

## Agent CLI (attend)

- **Primary channel verbs**: `join` / `leave` / `channels` / `dissolve` (with `#` prefix tolerated); `channels` shows all channels with joined marks — the mirror of the TUI's `/channels`. `send`/`reply` take `--channel` (`--focus` remains as alias).
- **`focus` is a deprecated alias** over the same handlers (CLI-is-contract): every spelling keeps working, one stderr note steers to the primary verbs. Agent-side leave-all folds into `scene private` — the `clear` cross-surface homonym is gone.
- **#388** — `render_sender_label_plain`: escape-free labels for the drain injection and all piped output (`TermCaps::Mono` still emits style bits by design, which was the leak).
- **#389** — `agent_fmt::compact_time`: one dependency-free compact datetime (civil-date math, once-per-process `date +%z` offset probe) used by `attend inbox` (table + block modes) and the drain reason.
- **#394** — phantom agents: `attend-session` normalizes `origin_path` through the managed-worktree rule (`<root>/.claude/worktrees/<n>` → `<root>`), so a worktree hop never changes the persona; `attend-instances::register` now *migrates* — evicts the session's entry from every other cwd's registry under per-file locks instead of minting a sibling. Both regression-tested.

## Operator TUI (attend-chat)

- **#389** — cell datetime beside the charm on the identity block's third line (same `compact_time`).
- **/peers + /whois** graduate Planned → Implemented: roster keyed strictly by session id (heartbeat-fresh), names via the instances registry, roots via the shared normalization — one row per session, per #394.
- **/invite + /kick** (#393): asymmetric membership — `/invite @name [#g]` writes a *directed invitation* (never force-adds; consent stays with the agent), `/kick @name [#g]` removes via the new `attend_groups::kick` API and sends a directed notice so the kicked agent's state doesn't drift silently. `@name` resolution for actions is exact-match (ambiguity refuses); the send path's fuzzy matching deliberately doesn't apply.
- **#392** — destination flag: inverse label pinned lower-right in the input box showing the resolved send target, derived through the send path's own routing (no duplicated rules), live-updating with `#`/`@` prefixes.
- **#393** — tab model: `[merged] [#open] [#name…]` strip with *merged* pinned at slot zero; Tab-when-input-empty cycles, Alt+1..9 jumps; per-channel tabs filter the transcript, merged is the full stream and the launch default; no-argument `/leave` `/purge` `/dissolve` `/invite` `/kick` scope to the foreground tab; dissolve closes the tab and advances, floor `#open` (which cannot be dissolved); merged-tab sends default to `#open` with the destination flag making it visible.
- **#390** — path attachments: a pasted/dropped absolute file path (shell-quoted tolerated) renders as an attachment chip in the input box and message cells; the path travels as plain body text (wire unchanged); a vanished file degrades to the raw path, never errors a cell.

## Docs

The three synchronized messaging docs (SKILL.md, reheat, attend way) + all help strings speak channel; each states where the chat analogy stops (no receipts/typing/edits; lifetime = project liveness, ADR-136). CLI reference regenerated.

## Verification

856 workspace tests green (attend-chat 172→208, plus new coverage in attend, attend-groups kick, attend-session normalization, attend-instances migration, agent-fmt compact_time); clippy zero warnings across the workspace.


aaronsb:
## Code Review — ADR-173 implementation (10 commits, ~2,250 added lines)

Reviewed against ADR-173, issues #388–#394 (including the #393 comment thread carrying the tab-model spec), with the worktree checked out and the full workspace test suite re-run locally (green, exit 0). The tab model's pure-function core (`tabs.rs`) with its test coverage, the two-layer base-channel guards, and the seen-set consult surviving the purge re-scope are all solid work. Findings below, ranked.

---

### Blocking

**1. `/purge open` from a channel tab purges the wrong channel**

**Location**: `tools/attend-chat/src/slash.rs:336-338` + `tools/attend-chat/src/app/keys.rs:477-494`

**Problem**: dispatch collapses the *explicit* argument into the no-argument form — `"purge" => match group_arg(args) { None | Some("open") => SlashOutcome::Purge(None), … }`. Pre-tab-model, `Purge(None)` meant "base channel", so that collapse was harmless. The tab model re-purposed `None` to mean "scope to the foreground tab" (`run_purge` → `tabs::command_scope(channel, foreground)`). Net effect: an operator on tab `#deploy` who types `/purge open` (or `/purge #open`) purges `#deploy`'s history, not `#open`'s.

**Why it matters**: destructive action on the wrong target, silently — the exact class of misroute the destination flag (#392) exists to prevent on the send side. The age tail + consumer consult soften the blast, but the intent inversion stands.

**Suggestion**: stop conflating explicit-base with no-argument. Let dispatch emit `Purge(Some("open".into()))` and have `run_purge` map the *resolved* name to the on-disk `None` target (it already does: `if resolved == BASE_CHANNEL_NAME { None }`). The `None | Some("open")` arm can simply become two arms. One dispatch test currently pins the wrong behavior (`dispatch_purge_defaults_to_base_channel`, slash.rs:651-659) and needs updating with it.

---

### Should-fix

**2. `attend dissolve open` is unguarded on the agent CLI — and deletes the reserved `@open/` dir**

**Location**: `tools/attend/src/main.rs:88-90`, `tools/attend/src/cmd/channels.rs:51-62`, `tools/attend-groups/src/lib.rs:174-188`

The TUI rejects `#open` dissolve in two layers (dispatch + foreground resolution, both tested). The agent CLI has zero: `attend dissolve open` (or `"#open"`, since `#` is stripped) reaches `Groups::dissolve("open")`, which removes `@open/` via `remove_dir_all` — the same reserved dir `cleanup_stale` deliberately never sweeps because it can hold pre-ADR-124 migration signals — and then prints `dissolved #open (was empty)`, a false success. The #393 spec says dissolve "must reject it the same way" as join/leave; `validate_group_name` already reserves the name for `join`, so mirror the rejection in `cmd_dissolve` (and ideally guard `Groups::dissolve` itself against `open`/`broadcast` as defense in depth). `attend leave open` printing `left #open` is the same shape, cosmetic-only.

**3. Two consumers still read the RAW session-record cwd — and now disagree with the normalized identity**

**Location**: `tools/attend-chat/src/sessions.rs:60` → `tools/attend-chat/src/chip/registry.rs:135-146`; `tools/sensor-peers/src/lib.rs:807`

After this PR, an agent's identity and its drain/scan tray are anchored on the *normalized* root (`util::own_origin_cwd()` → `identity().origin_path`). But `sessions::discover` still yields the raw record cwd, and the chip registry's seed path derives identity and routing from it: `Identity::for_cwd(&seed.cwd)` + `resolve_nickname → cwd_dir(seed.cwd)`. For a session currently inside a managed worktree with no signal traffic yet, `@`-completion offers a worktree-derived nickname, and a directed send resolved through that seed lands in the worktree-encoded tray — **which the target no longer scans** post-#394. That is a delivered-but-never-read DM, plus a phantom-flavored identity in the legend (the instance lookup at the raw cwd misses too, since `register` now evicts it). `peers.rs` normalizes correctly; the seed path should too — one `attend_session::normalize_origin` in `discover_in` fixes both display and routing at the single choke point. `sensor-peers`' raw read is the milder cosmetic sibling but worth the same treatment.

**4. `evict_elsewhere` never runs on the reclaim path**

**Location**: `tools/attend-instances/src/lib.rs:207-212`

`register_with_age` returns early on the existing-entry path, before the migration sweep at line 236. A stale sibling entry (left by a pre-fix binary, or by a crash between write and evict) therefore persists until 7-day GC whenever the session's current-cwd entry already exists — i.e., every register after the first. The `content.contains(session_id)` pre-filter makes the sweep cheap; run it on both paths. (Lock discipline itself checks out: the own-cwd lock is dropped before eviction, each foreign lock is held one at a time — no deadlock. Ping-pong requires two *live* processes claiming the same session at different cwds, a mixed-version transient that self-heals; acceptable.)

**5. Alias fidelity gap: `focus on` doesn't strip `#`, and the validator doesn't reject it**

**Location**: `tools/attend/src/main.rs:85-90` vs `:154-159`; `tools/attend-groups/src/lib.rs:406-423`

`attend join "#deploy"` strips the prefix; `attend focus on "#deploy"` passes it through, and `validate_group_name` rejects `_`/`@`/`/`/space but not `#` — so the alias path mints a group literally named `#deploy` (rendering as `##deploy` in `attend channels`). Behavior-identical-to-the-primary-verb is the alias contract. Cheapest fix: strip in the `cmd_join`/`cmd_leave`/`cmd_dissolve` handlers instead of at the four call sites; also reasonable to reject `#` in `validate_group_name` outright.

**6. Enter handler uses the un-normalized foreground tab**

**Location**: `tools/attend-chat/src/app/mod.rs:152` vs `:318-321`

The render path normalizes (`fg_tab`), and the destination flag derives from the normalized scope — but `handle_enter` receives the raw `foreground` state. When a peer dissolves the foregrounded channel, the flag shows `#open` while Enter attempts a send to the ghost channel and errors (`unknown group`). No misroute — it fails safe — but it breaks #392's stated guarantee that the flag cannot disagree with what Enter does. Normalize once in the Enter arm (or write the normalized tab back into state when it differs).

**7. Lockstep leftover: `send.rs` runtime errors still speak the focus register**

**Location**: `tools/attend/src/cmd/send.rs:130-147`

`error: no focus group named '…'`, `active focus groups (…)`, `drop --focus to broadcast` — all reachable from the renamed `--channel` flag path. The three messaging docs and help strings converged cleanly; these error strings are the residue.

---

### Nits

- `send --channel "#deploy"` doesn't tolerate the `#` prefix while `join`/`leave`/`dissolve` do — inconsistent tolerance across the new surface (`cmd/send.rs:95-160`).
- `skills/attend/SKILL.md` Scenes block still reads "leave all groups (project only)" — one register straggler.
- `cmd_leave` prints `left #name` even for a group you were never in (`channels.rs:25-29`, `r.leave().ok()`); the TUI's `leave_group_in` is precise here. Pre-existing shape, but the asymmetry is now visible across converged surfaces.
- `/kick` can only target claude sessions — the roster is session-record-keyed, so an ADR-170 human member (sanitized username in `_groups.yaml`) is unkickable. Likely intended; worth one sentence in the /help description or docs.
- `run_invite`/`run_kick` notices go to `cwd_dir(peer.root)` — the project tray, so every session in that project sees them, not just the target. Consistent with existing `@name` DM semantics; recording it as a known property.

---

### #393-comment invariants — verdict

1. **Consent asymmetry (invite ≠ force-add; kick unilateral + notice)** — **HOLDS.** `run_invite` (keys.rs:381-419) writes only a directed signal via the standard `write_signal` wire path (`from|project|cwd|body`, same sender identity as every TUI send); no `Groups::join` on the target's behalf anywhere. `run_kick` (keys.rs:426-448) removes via `Groups::kick` and sends the directed notice only after confirmed success. `kick` errors are precise (unknown group / non-member), rejoin-after-kick unblocked.
2. **Dissolve-advance with `#open` floor** — **HOLDS.** `advance_after_dissolve` (tabs.rs:100-113) clamps to strip end and floors to `Channel("open")`; `run_dissolve` moves focus only when the foreground *was* the dissolved tab (keys.rs:253-258). Clamp, floor, and background-dissolve-keeps-focus are all tested.
3. **Merged pinned at slot zero** — **HOLDS.** `Tab::default() = Merged` is the launch state (app/mod.rs), strip renders merged first, cycle order is merged → #open → …, `jump(1) = Merged`, and channel lifecycle can never close it (lifecycle only touches `Tab::Channel`).
4. **`#open` indissoluble** — **HOLDS in the TUI, in both layers**: explicit-argument rejection in `slash::dispatch` and foreground-resolution rejection in `run_leave`/`run_dissolve`/`run_invite`/`run_kick`, with `foreground_scoped_commands_respect_base_channel_invariants` covering both the Merged and open-tab defaults. **Not mirrored on the new agent CLI `dissolve` verb** — finding 2.

### Remaining verified risks (no action needed)

- **Hook-JSON with bracketed timestamps**: `render_drain_reason`'s new `[HH:MM]` prefix flows through `json_escape` (inbox.rs:557,693), which covers quotes/backslash/control chars; `compact_time` output is digits, `-`, `:`, space — no hazard. Verified.
- **Channel provenance**: `channel_for_path` classifies by inbox dir, agreeing with `accept_path`'s first-component logic; encoded cwd trays always start with `-`, so they cannot collide with `_broadcast`/`@…`. Directed-renders-merged-only is the documented #393 scope choice.
- **`normalize_origin` false positives**: only paths literally containing `/.claude/worktrees/` are affected, first occurrence wins, root-level pathological case passes through. A session launched *with the worktree as its cwd* maps to the parent root — which is the stated persona rule, not a bug.
- **Alias dispatch**: every `focus` spelling routes onto the same handlers with only the stderr note added; `send --focus` parses via the clap alias. Fidelity holds except for finding 5.

**856 workspace tests re-run locally: green.** With finding 1 fixed (and ideally 2–3), this is in good shape to merge.

---

aaronsb:
Review remediation landed in bc66dc3: blocking /purge-open inversion fixed (explicit stays explicit, dispatch test re-pinned), CLI dissolve gains the #open guard, identity consumers (chip seed + sensor-peers presence) normalize through #394, evict-on-reclaim, alias #-strip parity, Enter/flag agreement, channel vocabulary in send errors. 856 tests green, clippy zero.

Issue #406, opened 2026-07-22
kick guard: agents must not be able to remove human members — enforce at the attend-groups API


aaronsb:
Operator requirement (2026-07-22): the attend binary must carry a guard preventing an agent from kicking an operator. Today kick is TUI-only (capability asymmetry, ADR-173 D5), but surface absence is not enforcement: Groups::kick is in the shared crate, and a future agent-side verb (or any confused caller) could remove a human member.

Enforce at the API: Groups::kick refuses when the ACTING member id is a session UUID (agent) and the TARGET member id is a username (human, ADR-170 sanitized form). Member-kind discrimination: session ids are UUID-shaped; human ids are sanitized usernames — encode the discriminator once in attend-groups next to the member-id documentation, with tests. Operators kicking anyone: allowed. Agents kicking agents: policy question — default allow (peer coordination) unless the operator says otherwise.

aaronsb:
Shipped in PR #407 (attend/attend-groups 0.12.0): is_agent_member discriminator + Groups::kick refuses agent→human at the shared API (four-quadrant tests). Deployed at 349f9ac.

Issue #405, opened 2026-07-22
hierarchical narrowing command helper — Rhino/AutoCAD-style modal completion through subcommands and arguments


aaronsb:
Operator spec (2026-07-22): extend the #398 narrowing dynamic to a full command grammar. Today: `/c` shows all commands with /channels and /clear highlighted (correct), but `/channels ` (trailing space) falls to the ambiguous default legend. Wanted: after a completed token + space, the helper descends one level — shows THAT command's subcommand list, narrowing by prefix exactly like the top level (`/channels l` underlines `list`), then on completion descends again to the next argument's domain (agents for @-args, channels for #-args, free text for descriptions), recursively until the grammar is satisfied. The single-match help slot (#398) and the assert window (#400) follow the same recursion — help for the deepest unambiguous node.

Prior art the operator names: Rhino 3D's command entry, classic AutoCAD — 'a bit like autocomplete, a bit like a modal command line.' Consistency requirement: EVERY command follows the pattern (ArgKind becomes a per-command grammar, replacing the single-valued limitation noted in PR #395).

aaronsb:
Shipped in PR #407 (attend-chat 0.6.0): grammar.rs per-command token grammars, hierarchical narrowing at every level, contextual help at the deepest unambiguous node, Tab completion at subcommand level. Deployed at 349f9ac.

Issue #400, opened 2026-07-22
attend-chat: failed-command result must assert over the slash help for a timed window (error-styled)


aaronsb:
Operator finding on the #398 helper (2026-07-22): execute a failing command — e.g. /dissolve on a channel with live members — while the command text stays in the input. The single-match help keeps the status slot, so the error ('#testing: 1 live member — not dissolving') is invisible until the input is cleared. Logical, but the operator never sees why nothing happened.

Spec (operator-authored):
1. When a command executes and the input still single-matches (help would win), the fresh result TEMPORARILY asserts over the help — a timed window, visually brighter (red for errors).
2. Window expires → display falls back to the help (command text still present).
3. Clearing the input → stored result shows as before (#398 behavior unchanged).
4. The time constant must be a named top-level const, not buried inline.

Implementation notes: needs a set-timestamp beside the stored status + a priority rule in the render swap (fresh-status > help > stored-status), and honest error provenance (EnterAction carrying error-ness from SlashOutcome::Err, not string sniffing) to drive the red styling. Render repaint granularity is bounded by the 5s tick + keypresses; window should be chosen with that in mind.

Issue #398, opened 2026-07-22
attend-chat: slash helper — single-match shows command help in the status slot, restored on backspace-to-empty


aaronsb:
Operator UX proposal (2026-07-22, 0.5.0 pass): the last executed slash-command result currently renders under the slash-command list while composing. Refine the status slot by match cardinality:

1. Remember the last executed command's result string.
2. Bare `/` → full command list, last result still shown beneath.
3. Typing letters sharpens the list by prefix (`/p` → peers, purge; `/pe` → peers …).
4. The moment exactly ONE command matches, the status slot swaps from the last result to that command's HELP (description/usage) — inline documentation at the point of intent.
5. Backspacing to an empty prompt restores the previous result string unchanged.

Effect: command feedback is never lost by browsing the menu, and every command self-documents the instant it's unambiguous. Implementation surface: helper.rs (match state) + the status line source in app; the registry already carries per-command descriptions.

PR #407, opened 2026-07-22
feat: channel lifecycle, hierarchical command grammar, kick guard (#404 #405 #406)

### LINKED ISSUE #404: channel lifecycle commands: /channels create + describe (both surfaces) — slash commands top-level, subcommands steer
Operator spec (2026-07-22): there's no explicit channel creation (join creates implicitly) and no channel description at all. Wanted, under the convention that slash commands are TOP-LEVEL and subcommands steer into them:

- `/channels` — subcommand family: `list` (default, today's behavior), `create <name> [description…]`, `describe <#name> <description…>`.
- Same commands on the agent CLI: `attend channels` keeps bare-list default, gains `create`/`describe` subcommands (clap-native).
- Descriptions live in _groups.yaml (attend-groups owns the wire format — golden tests must cover the new field; absent field stays valid for old files).
- Description surfaces: /channels list output, attend channels table, and eventually the tab strip/legend.

aaronsb:
The final bundle of today's riding-the-loop session. Closes #404, #405, #406.

## #404 — channel lifecycle
- attend-groups: `GroupEntry.description` (single-line; absent key parses `None` — golden test extended for the wire-format change), `create()` (explicit lifecycle, pinned so an empty channel survives until joined/dissolved), `set_description()`.
- Both surfaces: `attend channels [list|create|describe]` (bare = list, # tolerated, Description column) and `/channels [list|create|describe]` — slash commands stay top-level, subcommands steer (operator convention). Tab strip shows the foreground channel's description, dimmed.

## #405 — hierarchical narrowing helper (Rhino/AutoCAD modal command line)
New `grammar.rs`: per-command token grammars (`Agent | Group | Word | Rest | Subcommands`) replace the single-valued ArgKind. The helper narrows candidates by prefix at EVERY level — type, underline, complete+space descends to the next token's domain (subcommand list / agent legend / channel legend / free-text hint), recursively; backspace ascends for free because position re-derives from the input string each render (no mode state). The #398 help slot follows to the deepest unambiguous node; the #400 assert window still wins while fresh. Tab completes at subcommand level. Sub-level lists render through the same row renderer as the top level — refactored, not forked.

## #406 — kick guard
`is_agent_member` discriminator (session UUID or `pid-` fallback = agent) documented beside the member-identity contract; `Groups::kick` refuses agent→human AT THE SHARED API so no surface can reach around it. Four quadrants tested: agent→human blocked (incl. unresolved pid- agents), agent→agent allowed, human→anyone allowed. TUI acting identity pinned by test so a future change that would make the TUI agent-shaped fails before shipping.

## Verification
897 workspace tests green (attend-groups 18→23, attend-chat 220→243), clippy zero warnings, CLI reference regenerated.

### DIFF
--- a/docs/cli/attend.md
+++ b/docs/cli/attend.md
@@ -17,6 +17,9 @@ This document contains the help content for the `attend` command-line program.
 * [`attend join`↴](#attend-join)
 * [`attend leave`↴](#attend-leave)
 * [`attend channels`↴](#attend-channels)
+* [`attend channels list`↴](#attend-channels-list)
+* [`attend channels create`↴](#attend-channels-create)
+* [`attend channels describe`↴](#attend-channels-describe)
 * [`attend dissolve`↴](#attend-dissolve)
 * [`attend focus`↴](#attend-focus)
 * [`attend focus on`↴](#attend-focus-on)
@@ -58,7 +61,7 @@ Active awareness for Claude Code sessions
 * `chat` — Launch the interactive chat TUI (ADR-120)
 * `join` — Join a channel (creates it if absent)
 * `leave` — Leave a channel
-* `channels` — List channels — all available, with joined ones marked
+* `channels` — Channel lifecycle (default: list all, joined ones marked)
 * `dissolve` — Dissolve a channel (removes it for every member)
 * `focus` — Deprecated alias for the channel verbs (join/leave/channels/dissolve; ADR-173)
 * `scene` — Activate a named scene (reconfigure channel membership; `scene private` leaves all channels)
@@ -222,9 +225,49 @@ Leave a channel
 
 ## `attend channels`
 
-List channels — all available, with joined ones marked
+Channel lifecycle (default: list all, joined ones marked)
 
-**Usage:** `attend channels`
+**Usage:** `attend channels [COMMAND]`
+
+###### **Subcommands:**
+
+* `list` — List all channels with joined marks (default)
+* `create` — Create a channel without joining it (pinned so it persists empty)
+* `describe` — Set or replace a channel's single-line description
+
+
+
+## `attend channels list`
+
+List all channels with joined marks (default)
+
+**Usage:** `attend channels list`
+
+
+
+## `attend channels create`
+
+Create a channel without joining it (pinned so it persists empty)
+
+**Usage:** `attend channels create <NAME> [DESCRIPTION]...`
+
+###### **Arguments:**
+
+* `<NAME>` — Channel name (with or without the # prefix)
+* `<DESCRIPTION>` — Optional single-line description (all trailing words)
+
+
+
+## `attend channels describe`
+
+Set or replace a channel's single-line description
+
+**Usage:** `attend channels describe <NAME> [DESCRIPTION]...`
+
+###### **Arguments:**
+
+* `<NAME>` — Channel name (with or without the # prefix)
+* `<DESCRIPTION>` — The description text (all trailing words; empty clears)
 
 
 
--- a/tools/attend-chat/src/app/keys.rs
+++ b/tools/attend-chat/src/app/keys.rs
@@ -757,6 +757,15 @@ pub fn handle_tab(
         return unchanged;
     }
 
+    // Subcommand completion (#405): past the name, at a grammar
+    // subcommand choice, Tab completes the level's partial exactly
+    // like the top level completes the name. Silent fall-through
+    // otherwise — deeper levels (`@`/`#` tokens) keep the mention
+    // completion below.
+    if let Some((next, new_cursor)) = slash::complete_subcommand(buf) {
+        return TabResult { new_buf: next, new_cursor, new_cycle: None };
+    }
+
     // Cycling: if we have a stored cycle and the buffer still ends
     // with the candidate we just inserted (sigil + name + trailing
     // space, the exact form `apply_completion` produces), advance
--- a/tools/attend-chat/src/app/keys.rs
+++ b/tools/attend-chat/src/app/keys.rs
@@ -1234,6 +1234,35 @@ mod tests {
         assert_eq!(members, vec!["someone-else"]);
     }
 
+    /// #406: the agent→human kick guard lives in the shared crate
+    /// (`Groups::kick` refuses when the ACTING id is agent-shaped and
+    /// the target is human). The TUI acts with an empty member id,
+    /// which classifies as human — so the guard can never fire here
+    /// and operator kicks of human members keep working. If the
+    /// acting identity ever changes to an agent shape, this test
+    /// fails before the regression ships; and if the guard did fire,
+    /// `kick_member_in`'s Err arm surfaces its message through the
+    /// #400 status path rather than swallowing it.
+    #[test]
+    fn tui_kick_acts_as_operator_and_can_remove_human_members() {
+        let base = tempdir_like();
+        assert!(
+            !attend_groups::is_agent_member(""),
+            "TUI acting id must classify as human (operator power)"
+        );
+        // A human member (username shape, ADR-170) is kickable from
+        // the TUI — the one-way power the guard preserves.
+        attend_groups::Groups::new(&base, "somebody")
+            .join("ops", false)
+            .unwrap();
+        match kick_member_in(&base, "ops", "somebody", "somebody") {
+            EnterAction::ClearWithStatus(s) => {
+                assert!(s.contains("kicked @somebody from #ops"), "got: {s}")
+            }
+            _ => panic!("operator kick of a human member must succeed"),
+        }
+    }
+
     #[test]
     fn channels_status_lists_base_and_counts() {
         let base = tempdir_like();
--- a/tools/attend-chat/src/app/keys.rs
+++ b/tools/attend-chat/src/app/keys.rs
@@ -92,6 +92,12 @@ pub fn handle_enter(input_value: &str, signals: &[Signal], foreground: &Tab) -> 
             slash::SlashOutcome::ListChannels => {
                 channels_status_in(&signals_base(), attend_groups::member_alive)
             }
+            slash::SlashOutcome::CreateChannel { name, description } => {
+                create_channel_in(&signals_base(), &name, description.as_deref())
+            }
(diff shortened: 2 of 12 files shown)


aaronsb:
## Review — PR #407 (channel lifecycle, grammar walk, kick guard)

Reviewed against #404/#405/#406, ADR-173 (surface conventions), ADR-170 (member identity), with focus on the `_groups.yaml` wire-format change, the kick guard, the grammar walk, and pinned-create semantics. One blocking finding (verified by execution against the built `attend_groups` rlib), a few should-fixes, and explicit verdicts on the two top risks below.

---

## Blocking

### `Groups::create` accepts embedded newlines in the description → `_groups.yaml` entry injection, including member migration

**Location**: `tools/attend-groups/src/lib.rs:129-150` (`create`), reachable via `tools/attend/src/main.rs:90` + `tools/attend/src/cmd/channels.rs:82`

**Problem**: `set_description` (lib.rs:154-156) rejects `\n`; `create` does not — it only `trim()`s. The TUI path is safe (`rest_text` at `slash.rs:384` folds all whitespace via `split_whitespace`), but the CLI path joins raw argv with `" "`, and argv words can contain newlines:

```
attend channels create foo $'x\nevil:\n  pinned: true\n  members:\n    - fake-member'
```

I ran the serialize→parse round trip for that state through the actual crate. Result:

```
--- reparsed ---
evil: pinned=true members=["fake-member", "real-member"] desc=None
foo:  pinned=true members=[] desc=Some("x")
```

A fabricated **pinned** group `evil` materializes with an arbitrary member, and any members serialized after the description line migrate into it (the injected `evil:` header terminates `foo`'s entry mid-serialization). This then round-trips as truth on every subsequent write. That's exactly the "worse than transient description loss" class: torn semantics + member loss, and it's new in this PR — a write path that violates the single-line contract the crate itself states at lib.rs:152-153.

**Suggestion**: enforce the contract at the API, same as `set_description` — either reject `\n` in `create`'s description, or normalize with the TUI's fold (`description.split_whitespace().collect::<Vec<_>>().join(" ")`). Enforcing in the shared crate (not just `cmd_create`) matches the #406 posture: no surface can reach around it. One test: `create("x", Some("a\nb:"))` either errs or round-trips single-line.

(Related, pre-existing, see should-fix 3: `validate_group_name` also doesn't reject `#`/`:`, so names can produce lines the parser reads as comments/headers.)

---

## Verdicts on the two highest risks

### Risk 1 — `_groups.yaml` wire-format change: **acceptable-transient, with the one blocking exception above**

- **Old-binary safety**: the pre-#404 parser's indent-2 match already had a `_ => {}` arm, so `description:` lines are silently ignored — no parse failure, no member loss, no torn state. An old binary's `save_state` (join/leave/kick/cleanup) drops descriptions on write; they're re-settable. Since old `attend run` loops call `cleanup_stale` on ticks, expect descriptions to vanish *repeatedly* during a mixed-version window, not just once — still transient, converges on deploy.
- **Round-trip robustness** (verified by reading `parse_groups_yaml`, lib.rs:529-601): `split_once(':')` keeps descriptions containing `:` intact; leading `- ` is fine (indent 2 ≠ 4); a value ending in `:` is fine (room headers require indent 0); leading-`#` *values* are fine (comment skip tests `trimmed`, which starts with `description`). No serde/external consumers of the file exist — the hand parser is the only reader. The golden test extension (`description_round_trips_and_absent_is_none`) covers the right two cases.
- The only corruption vector found is the newline hole above.

### Risk 2 — kick guard: **sound as defense-in-depth; correctly scoped; nothing newly weaponized**

- **Shape discrimination** (lib.rs:54-64): human ids are `sanitize_id_component($USER)` (ADR-170, `agent-identity/src/identity.rs:102`) — alnum/`-`/`_`. A real `$USER` that is 8-4-4-4-12 hex or `pid-*` is contrived to the point of irrelevance, and the failure direction is **fail-closed**: such a human would be *blocked* from kicking humans, never an agent gaining kick power. Agent→human masquerade via shape is impossible: agent-acting callers get their id from attend's own session (uuid or `pid-<n>` fallback), and `pid-` is classified agent (tested).
- **Arbitrary `member_id` in `Groups::new`**: pre-existing exposure, not newly weaponized. The API was never authenticated, and anything in this trust domain can write `_groups.yaml` directly. The guard is honest defense-in-depth against a *confused* future caller, exactly as #406 framed it.
- **TUI acting id `""`**: pre-existing (`keys.rs:288` dissolve used `""` before this PR; kick at `keys.rs:510` likewise). `is_agent_member("")` → false → operator power preserved. The pinning test (`tui_kick_acts_as_operator_and_can_remove_human_members`) is honest: it asserts the classification *and* exercises a real kick through `kick_member_in`.
- Guard placement note (fine as-is): the guard fires before existence checks, so an agent kicking a human-shaped *non-member* gets the guard error instead of "not a member". Slightly less precise, arguably the better information-hiding default.

---

## Should-fix

**1. Tab after an unknown completed subcommand appends garbage** — `tools/attend-chat/src/slash.rs:661-690` (`complete_subcommand`). `walk` returns `Pos::Sub { partial: None }` both for a clean descent (`"/channels "`) and for an unknown completed token (`"/channels bogus "` — see `unknown_completed_subcommand_stays_at_level_unnarrowed`). The `None if input.ends_with(whitespace)` arm can't tell them apart, so `"/channels bogus "` + Tab yields `"/channels bogus list "`, which Enter then rejects. Suggest distinguishing the two in `Pos::Sub` (e.g. an `at_unknown: bool`, or only bare-complete when `args` is empty) and pinning `complete_subcommand("/channels bogus ")` → `None` in `subcommand_completion_tests`.

**2. `/channels create ##x` creates a phantom group that silently vanishes** — `tools/attend-groups/src/lib.rs:484-501` (`validate_group_name`) + `slash.rs` `strip_prefix('#')`. The TUI strips *one* `#` (CLI strips all via `trim_start_matches`, main.rs:90), so `##x` reaches `create` as `#x`, which the validator accepts. Serialized as `#x:`, the parser reads that line as a comment: the entry evaporates on next load and its `@#x` dir becomes orphan-sweep fodder. Same gap pre-exists via `/join ##x`, so this isn't blocking — but `create` giving it a second, more deliberate-looking route is the right prompt to harden the validator (reject `#` and `:` in names) rather than patch call sites. Also closes the CLI/TUI `#`-stripping parity question: with `#` rejected in names, stripping one vs. all becomes observably identical.

**3. `ChannelsCmd::Create` lacks `allow_hyphen_values`; `Describe` has it** — `tools/attend/src/cli.rs:218-232`. `attend channels describe x -- like this` accepts hyphen-leading words; `create` errors on them. House style elsewhere (`cli.rs:105,124,131`) pairs `trailing_var_arg` with `allow_hyphen_values`, so Create is the outlier. Trade-off to note either way: `allow_hyphen_values` swallows `--help` after the name into the description — consistent with `attend send`, so fine, just be deliberate.

---

## Nits

- `tools/attend/src/cmd/channels.rs:120-127` — `cmd_channels` reads state twice (`r.all_groups()` then `load_groups`); a write between them can misalign the Description column against the count/pin columns. One `load_groups` could feed all columns.
- Case seam: subcommand chips highlight case-insensitively (`chip_matches`) and Tab completes `DE` → `describe`, but `walk`/`dispatch`/`contextual_help` match subcommand names case-sensitively — typed `Create` errs while the chip underlines. Same idiom as the top level (`lookup` is case-sensitive), so pre-existing pattern, not a regression.

---

## Verified clean (the rest of the risk list)

- **grammar.rs walk**: trailing/multiple spaces handled by `split_whitespace` + the explicit `ends_with(whitespace)` descend gesture; `complete_subcommand`'s byte splice is boundary-safe (the partial is provably the input's literal unterminated tail, so `input.len() - p.len()` lands on a char boundary even with multibyte text); `contextual_help` has no indexing, no panic path on long/unicode/empty adversarial input; sigil narrowing is symmetric (`sigil_partial` wrong-sigil → `None`, no legend hijack) and matches dispatch's whitespace-splitting, so helper and Enter can't disagree. Quoted pasted paths (attach.rs) only matter at `Rest` positions, where the walk position never advances — no misbehavior. The `Subcommands`-as-sole-top-token constraint is convention, but `walk_tokens` would actually handle a mid-grammar choice correctly if it ever appeared.
- **create() pinned semantics**: `/dissolve` of a created-but-empty channel works from the TUI — `dissolve_group_in` (keys.rs:281-307) counts live members over `Some([])` → 0 → proceeds. `attend focus unpin` on a created empty channel deletes it *immediately* (unpin's empty-cleanup, lib.rs:235-249) rather than leaving it sweep-eligible — coherent with unpin's existing contract (unpin = release persistence), just worth knowing the description dies with it.
- **Regressions**: #398/#400 status-slot priority unchanged (`fresh` still wins; `contextual_help("/")` still `None` so the bare-`/` menu keeps the last result); help now persisting through committed-command args is the *specified* #405 behavior, not drift. Tab-scoped `/purge`, `/leave`, `/dissolve` dispatch untouched; `handle_tab` ordering (slash-name → subcommand → mention cycle) preserves the original short-circuit order.

## Assessment

The architecture is right: wire format owned by one crate with the golden test extended, the guard at the shared API instead of a surface, the walk as a pure re-derived function with the state table as tests, and the chip-row refactor shared instead of forked. The one blocking item is a one-line-contract enforcement gap (`create` vs `set_description`) with an outsized failure mode — fix that (plus ideally the Tab-after-unknown nit riding the same subsystem) and this is solid to merge.

aaronsb:
Remediation landed in 1878328:

**Blocking (yaml injection) — fixed.** create and set_description now share `normalize_description` (rejects any control char); regression test reproduces the reviewer's exact newline payload and asserts nothing is written. Verdict risk 1/2 confirmed: transient description-drop on mixed-version writes is the accepted ceiling; the injection was the real hole and it's closed.

**Should-fixes — all in:** validate_group_name rejects `#`/`:` (+test); complete_subcommand gates bare-descent on an empty arg tail; CLI Create gains allow_hyphen_values.

**Nit:** cmd_channels single-reads the file so the Description column can't misalign.

897→ workspace tests green (attend-groups 23→25 with the injection regression), clippy zero.

PR #413, opened 2026-07-28
feat(claude-projects): relocate a project's session history to a new path

aaronsb:
## Summary

- Adds `claude-projects relocate OLD NEW` — moves a project's session history so sessions started in `OLD` resume in `NEW`. Previews by default; `--execute` applies.
- A directory rename alone is **not** sufficient: the path is recorded in the project dir name, `sessions-index.json` (`originalPath`/`projectPath`/`cwd`/`fullPath`), the `~/.claude.json` `projects` key, `~/.claude/history.jsonl`, and the transcripts' top-level `cwd`. All are remapped.
- Links `claude-projects` onto `PATH` via `link_path_binaries`, consistent with the `ways`/`attend` binaries. Its pre-1.0 symlink pointed at `~/.claude/tools/` and had been dangling since the XDG projection (ADR-142); this repairs it.

## Design notes

- The slug encoding (`[^a-zA-Z0-9]` → `-`) is **lossy** — `/a/b_c` and `/a/b-c` collide — so the command takes real paths and never decodes a slug back.
- `cwd` was verified to occur **only** at the top level of transcript records (13,631 occurrences across the largest local transcript; zero nested, zero unparseable lines), so this is a targeted field remap rather than a find/replace.
- Lines not mentioning the old path are copied byte-for-byte; only affected records are re-serialized. mtimes are preserved so relocating isn't mistaken for activity.
- Message text referencing the old path is deliberately left as historical record, and the residual count is reported rather than silently ignored.

## Test plan

Exercised against a synthetic fixture under an isolated `$HOME` (no real session data touched):

- [x] Path-boundary matching — `IRIS-old` not matched; `/home/aaron/.claude` untouched; `IRIS/src` correctly remapped
- [x] **Round trip is byte-identical** — relocate there and back `diff`s clean
- [x] mtime preserved across rewrite
- [x] `~/.claude.json` key renamed *in place*, preserving its position among 423 keys
- [x] `history.jsonl` — only the matching project rewritten
- [x] Live-session guard refuses (fired for real during testing); `--force` overrides
- [x] Existing-target refusal leaves everything untouched; `--merge` combines
- [x] `--keep-transcript-cwd` leaves transcripts byte-identical
- [x] Target working directory created when missing; errors if the path exists as a file
- [x] Missing source / identical paths / absent `~/.claude/projects` all exit cleanly (the last was a crash found and fixed during testing)
- [x] Regression: `list`, `stats` unaffected
- [x] `bash scripts/check-portability.sh` passes; `python3 -m py_compile` clean
- [x] Read-only preview against real data: 1530 `cwd` fields over 7.9 MB in 0.1s

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -200,6 +200,13 @@ link_path_binaries() {
       ln -sf "$APP_DIR/bin/$b" "$XDG_BIN/$b"
     fi
   done
+  # Script tools ship from tools/ rather than bin/ (nothing to compile), but are
+  # linked the same way so they're callable from any working directory.
+  for b in claude-projects; do
+    if [[ -e "$APP_DIR/tools/$b" ]]; then
+      ln -sf "$APP_DIR/tools/$b" "$XDG_BIN/$b"
+    fi
+  done
   return 0
 }
 
--- a/tools/claude-projects
+++ b/tools/claude-projects
@@ -13,16 +13,20 @@ Usage:
     claude-projects stats              # aggregate statistics
     claude-projects cleanup --dry-run  # show empty projects that could be removed
     claude-projects cleanup            # remove empty project directories
+    claude-projects relocate OLD NEW   # preview moving session history to a new path
+    claude-projects relocate OLD NEW --execute
 
 Install:
-    ln -s ~/.claude/tools/claude-projects ~/.local/bin/claude-projects
+    Symlinked onto PATH by scripts/install.sh, alongside the ways binaries.
 """
 
 import argparse
 import json
 import os
 import re
+import shutil
 import sys
+import time
 from datetime import datetime, timezone
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


aaronsb:
## Review: `relocate` subcommand

**Significant change** (387 lines added, one new stateful command). Line numbers below refer to `tools/claude-projects` at PR head (`a193d11`).

Reviewed with the data blast radius in mind rather than the code blast radius. The design is sound — guards run before any mutation, preview and execute share one code path via the `apply` flag, and `remap()` boundary-matching is genuinely correct (I checked `/a/b` vs `/a/bc`, the trailing-slash-in-recorded-value case, and `norm_path("/")`; all behave). The findings below are mostly about what happens when a step *fails* partway, which is where this command's stakes are.

---

### Blocking

#### 1. `sessions-index.json` is the one write that is not atomic

**Location**: `tools/claude-projects:787`

```python
idx.write_text(json.dumps(data, indent=indent, separators=sep, ensure_ascii=False))
```

Every other writer in this change uses tmp + `os.replace` (`:754`, `:824`). This one truncates the live file and streams into it, with no backup (unlike `~/.claude.json` and `history.jsonl`, which both get `backup_file()`).

**Failure scenario**: relocating a project whose `sessions-index.json` is ~2 MB, disk fills (or the terminal is `Ctrl-C`'d) between truncate and completion. The result is a half-written JSON file. On the next `claude-projects list` the project reports 0 sessions. Worse, `cmd_hygiene` (`:553`) treats any transcript whose `sessionId` is absent from a *non-empty* index as an orphan and offers to `unlink()` it (`:641`) — so a corrupt index converts into an offer to delete the transcripts.

**Fix**: same pattern as `rewrite_claude_json`:

```python
if apply and n:
    backup_file(idx)
    tmp = idx.with_name(idx.name + ".relocate-tmp")
    tmp.write_text(json.dumps(data, indent=indent, separators=sep, ensure_ascii=False))
    shutil.copystat(idx, tmp)
    os.replace(tmp, idx)
```

#### 2. `--merge` orphans the merged-in transcripts

**Location**: `tools/claude-projects:965-974`, `:993-996`

When both projects have a `sessions-index.json`, the merge loop hits `if target.exists(): continue` and skips the source index. The transcripts move; their index entries do not. `work_dir` is then `new_dir`, so `rewrite_sessions_index` runs against the *target's* index, which only knows about the target's own sessions and needs no remapping — it reports `rewrote 0 field(s)` in green.

**Failure scenario**: `relocate /a /b --merge --execute` where both have history. Twelve transcripts land in `-b`; `-b/sessions-index.json` lists only its original three. `claude-projects hygiene` then reports 12 orphaned transcripts and offers to delete them; answering `y` destroys exactly the history the relocate was meant to preserve. A secondary symptom: `old_dir` survives (correctly — `:973` won't `rmdir` a non-empty dir) but now holds an index describing sessions that live elsewhere, so `list` shows a phantom project with N sessions and 0 files.

**Fix**: merge the two `entries` arrays — read the source index, remap its entries (`projectPath`/`cwd`/`fullPath`), append to the target's entries deduped on `sessionId`, then write the target atomically. If that's more than you want in this PR, the honest interim is to refuse `--merge` when both directories have a `sessions-index.json`, with a message saying why. Silently skipping the file is the option that hurts.

Related: `--merge` also has no rollback and no manifest. If `shutil.move` fails on item 7 of 20 (`:972`), the command exits 1 with items 1–6 relocated and no record of which those were, and re-running `relocate B A --merge` would drag B's *own* files back with them. At minimum, print each moved filename as it moves so a partial failure leaves a transcript in the terminal.

#### 3. Partial failure reports success

**Location**: `tools/claude-projects:983-991`, `:1012-1013`

The transcript loop catches `OSError`, prints it, and continues. Nothing accumulates the failure. Execution then proceeds through sessions-index, `~/.claude.json`, and `history.jsonl`, and prints `Done.` plus a reverse hint.

**Failure scenario**: one transcript is read-only (or on a full filesystem). It keeps `cwd: /old`; the other nineteen get `/new`; the index and config say `/new`. The user sees `Done.` and a green `rewrote N cwd field(s)` and has no signal that the project is now internally inconsistent.

The reverse hint happens to still be *mostly* valid here (the un-rewritten file simply doesn't match on the way back), but the user has no way to know they need it.

**Fix**: collect failures, and gate the ending:

```python
failures = []
...
except Exception as e:            # see finding 5 re: OSError being too narrow
    failures.append((t.name, e))
...
if failures:
    print(f"\n  {red('Completed with errors')} — {len(failures)} file(s) not rewritten:")
    for name, e in failures:
        print(f"    {name}: {e}")
    print(f"  {dim('Re-run the same command to retry, or reverse with:')} ...")
    sys.exit(1)
```

`cmd_relocate` currently cannot exit non-zero after the execute block begins, which also makes it unsafe to use in a script.

---

### Should fix

#### 4. The live-session guard is scoped to this project, but two of the files are global

**Location**: `tools/claude-projects:873-877` vs `:998`, `:1004-1010`

`live` is computed only from `old_dir.glob("*.jsonl")`. `~/.claude.json` and `~/.claude/history.jsonl` are shared by *every* session on the machine, and both are rewritten read-whole-then-replace.

**Failure scenario**: a session is open in an unrelated project. You relocate an idle project. Between `CLAUDE_JSON.read_text()` (`:808`) and `os.replace` (`:824`), the live session appends a prompt to `history.jsonl` and updates its own entry in `~/.claude.json`. Our replace silently drops both. The `history.jsonl` case is a real lost prompt — small, but it's the other session's data, and nothing warns.

**Fix**: extend the guard to the global files — treat `CLAUDE_JSON` or `HISTORY_FILE` with an mtime inside `LIVE_SESSION_WINDOW` as `live` too. It's a two-line change and it makes the existing `--force` escape hatch cover the actual risk. (A real fix would need locking, which is out of scope; the guard is the right level of effort here.)

#### 5. `rewrite_jsonl` — the tail is outside the `try`, and the caller's `except` is too narrow

**Location**: `tools/claude-projects:706-755`, caller at `:989`

Three distinct issues in the same block:

**(a) `out.close()` at `:751` is outside the `try`.** Writes are buffered, so ENOSPC very often surfaces at `close()`, not at `write()`. When it does, `tmp.unlink(missing_ok=True)` at `:748` never runs and a partial `<uuid>.jsonl.relocate-tmp` is left in the project directory. The original is intact (good — `os.replace` never ran), but the debris is invisible to `hygiene`, which only globs `*.jsonl`. Move `close`/`copystat`/`replace` inside the `try`, or restructure with `try/except/else`.

**(b) No `fsync` before `os.replace`.** This is the classic delayed-allocation hazard: on ext4/btrfs, `close()` returning does not mean the data is on disk, but `os.replace` immediately unlinks the original inode. A power loss or kernel panic in the writeback window leaves a zero-length or truncated transcript with no original to fall back on. Given the PR's own framing of this data as irreplaceable, and that transcripts are the one category here that gets *no* backup, this is worth the two lines:

```python
out.flush()
os.fsync(out.fileno())
out.close()
```

**(c) The caller catches only `OSError`.** `UnicodeEncodeError` is a `ValueError`, and it is reachable: a transcript containing an escaped lone high surrogate (`"\ud800"` — plausible in arbitrary tool output) decodes fine via `json.loads`, but `surrogateescape` only encodes U+DC80–U+DCFF, so writing it raises. I confirmed this:

```python
>>> obj = json.loads('{"cwd":"/old/p","t":"\\ud800"}')
>>> open(f,"w",encoding="utf-8",errors="surrogateescape").write(json.dumps(obj, ensure_ascii=False))
UnicodeEncodeError   # isinstance(e, OSError) -> False
```

That escapes `cmd_relocate` as a bare traceback *after* the directory has already been moved. Catch `Exception` at `:989` and fold it into the failure list from finding 3.

#### 6. A `~/.claude.json` conflict doesn't block, unlike a directory conflict

**Location**: `tools/claude-projects:946-948` vs `:998-1002`

A pre-existing project *directory* at `new` blocks without `--merge`. A pre-existing `~/.claude.json` `projects` key at `new` does not — it's rendered as a yellow label in the preview but never added to `warnings` and never gates execution.

**Failure scenario**: the user opened `/new` in Claude once, so `~/.claude.json` has a key for it but `~/.claude/projects/-new` was never created (no sessions). `merging` is `False`, so nothing blocks. Everything moves, then `rewrite_claude_json` returns `"conflict"` and returns early *before* rebuilding (`:216` of the diff) — so the **old key stays in the file**, pointing at a path that now has no history, and the new path keeps its empty config. Output is a single yellow `skipped` line amid green successes.

**Fix**: add the conflict to `warnings` and to the `if ... and not args.merge: sys.exit(1)` gate at `:946`, so config conflicts and directory conflicts behave identically.

#### 7. `rewrite_sessions_index` swallows read errors and reports them as success

**Location**: `tools/claude-projects:770-774`

```python
except (OSError, json.JSONDecodeError):
    return 0
```

A corrupt or unreadable index is indistinguishable from "nothing needed changing" — both print `rewrote 0 field(s)` in green at `:996`. `rewrite_claude_json` already models the right answer for this file (a `(status, note)` tuple); the inconsistency between the two sibling functions is worth closing.

#### 8. A nested-`cwd` regression would be misreported as intentional

**Location**: `tools/claude-projects:738-744`

The empirical finding that `cwd` only appears at the top level is well-documented, and the fail-safe posture is mostly right. But trace what happens if Claude Code later adds a nested `cwd` (say under `toolUseResult`): the top-level field still maps, `changed` increments, then `if old in rendered: residual += 1`. The stale nested value lands in `total_residual`, which is printed as:

> `N line(s) still mention the old path in message text or tool output.`
> `Those are a record of what happened and are left as-is.`

So a genuine remapping gap gets reported to the user as a deliberate design decision. That is exactly the "silently does the wrong thing" shape.

**Fix**: cheap version — after remapping, walk the parsed object for any nested key named `cwd` whose value matches `old`, and if found, count it into a separate bucket printed as a warning rather than folded into the benign note. That turns a silent format drift into a visible one.

The slug-encoding assumption (`encode_project_path`) fails safe by comparison: if Claude changed the encoding, `old_dir.is_dir()` misses and you get the clean "No session history" error with near-matches. Good.

---

### Minor

**9. `newline=None` breaks the stated byte-for-byte guarantee** (`:717`, `:719`). Both handles use universal newlines. A line ending in `\r\n` is read as `\n` and written back as `\n`. Verified:

```
in  b'a\r\nb\n'
out b'a\nb\n'
```

The docstring at `:711` promises unaffected lines are "copied byte-for-byte" and the PR claims a byte-identical round trip; both are true in practice on Linux-authored transcripts but not guaranteed by the code. Pass `newline=""` to both `open` calls to make it actually hold. Related: a final line lacking a trailing `\n` gains one if it happens to be a rewritten record (`:744`), though it's preserved on the copy path.

**10. `os.replace` runs even when `changed == 0`** (`:751-754`). In the `--merge` case, `work_dir.glob("*.jsonl")` at `:985` includes the *target's* pre-existing transcripts, none of which mention `old`. Each is fully copied and atomically replaced for no reason. Guarding the tail with `if changed:` removes the corruption window entirely for every file that doesn't need touching — a large fraction of the I/O in a merge, and free insurance.

**11. The reverse hint isn't shell-quoted, and is wrong after a merge** (`:1013`). `claude-projects relocate /home/a/My Project /home/a/Other --execute` won't parse. Use `shlex.quote(new)` / `shlex.quote(old)`. Separately, after `--merge` the reverse command would move the *combined* project back to `old` — it should be suppressed or replaced with a merge-specific note.

**12. A path containing a newline silently no-ops the transcript rewrite.** `norm_path("/tmp/a\nb")` returns it unchanged; then `if old not in line` (`:721`) is always true because iterating a file never yields an embedded newline. The directory renames, index and config remap, and transcripts report `0 cwd field(s)` — a silent split-brain. Vanishingly rare, but a one-line guard (`reject if any(c in old + new for c in "\n\r")`) buys certainty.

**13. Backups accumulate and can collide.** `backup_file` (`:698`) stamps to the second, so two relocates within the same second overwrite each other's backup. And `~/.claude.json.bak-relocate-*` (a file with ~423 project entries) plus `history.jsonl.bak-relocate-*` are never pruned. `shutil.copy2` preserves mode, so permissions are fine — this is housekeeping, not exposure. Consider keeping the N most recent, or at least mentioning them in `hygiene`.

**14. Constant placement.** `CLAUDE_JSON` / `HISTORY_FILE` (`:669-670`) sit in the middle of the file while the sibling path constants `PROJECTS_DIR` / `HOME` live at `:33-34`. Small convention drift.

---

### What's fine

Saying so explicitly, since several of these were called out for scrutiny:

- **`remap()` boundary matching is correct.** `value == old or value.startswith(old + "/")` rejects `/a/bc` for `/a/b`, and a recorded value with a trailing slash (`/a/b/`) matches the `startswith` branch and maps correctly. The `fullPath` remap correctly uses the *directory* strings rather than the project paths.
- **Not resolving symlinks in `norm_path` is the right call.** The goal is matching a literal recorded string; `realpath` would break that. A mismatch fails gracefully into the "No session history" branch with near-matches, which is the correct outcome.
- **Guard ordering.** All refusals (`:942-948`) precede every mutation, including `os.makedirs(new)`. The `PROJECTS_DIR.is_dir()` guard at `:258` closes the crash the test plan mentions.
- **`shutil.copystat` for mtime preservation**, with the rationale in a comment — and it keeps the live-session guard meaningful across relocations, which is a nice second-order property.
- **`~/.claude.json` key rebuilt in place** to preserve position among 423 keys, rather than delete-then-insert. Good instinct.
- **`install.sh`** (`:203-208`) matches the surrounding loop exactly, sits before the `return 0` that the comment above `link_path_binaries` explains, and `ln -sf` does repair the dangling pre-1.0 link as claimed. No issues.
- **Color/error-output conventions** (`red('error')`, `yellow`, `dim`) match `cmd_hygiene` and `cmd_cleanup` throughout. `cmd_relocate` at ~135 lines is in line with `cmd_hygiene`'s ~120; it does mix plan-rendering, guards, and execution, and would read better as `_relocate_plan()` / `_relocate_apply()`, but that's consistent with the file's existing shape rather than a deviation from it. (The file is now 1095 lines and is approaching a natural split into data helpers / rendering / commands — worth noting, not worth doing here.)
- **Traceability**: no linked issue or ADR, but the PR body's design-notes section covers the load-bearing decisions (lossy slug encoding, why a rename alone is insufficient, why message text is left alone) at ADR-adjacent depth. That reads as adequate for a leaf tool.

---

### Assessment

The core design holds up and the empirical verification behind it is more thorough than most changes of this size get. What needs attention is the failure half: **findings 1, 2, and 3 should land before merge** — they're the three paths where a defect converts into permanently lost session history rather than an error message, and two of them route through this same file's `hygiene` command offering to delete the very transcripts that were just relocated. Findings 4–8 are meaningful hardening for a command that edits global state. Everything under Minor is discretionary.

---

aaronsb:
Remediated all review findings in e6d1bc4 — thanks, these were real and the three blocking ones each converted a defect into lost session history.

**Blocking**
1. **Non-atomic `sessions-index.json` write** — now `write_json_atomic` + backup, matching every other writer. Confirmed the amplification: `cmd_hygiene:553` is `if sid not in indexed_sessions and indexed_sessions`, so a *valid but incomplete* index flags the missing transcripts as orphans and `unlink()`s them.
2. **`--merge` orphaned transcripts** — the two indexes are now unioned by `sessionId`. Verified: 3 transcripts in, 3 indexed, 0 orphans, all paths remapped.
3. **Partial failure reported success** — failures accumulate; the command prints what failed and exits non-zero. Verified exit 1 with the failed file left byte-intact and no orphaned temp.

**Should fix** — 4 (global-file concurrency) now re-checks size/mtime bracketing the read and refuses rather than dropping another session's appends; 5 (a) close moved inside `try`, (b) `fsync` before swap, (c) callers catch `UnicodeEncodeError` — confirmed it's a `ValueError` and escaped as a bare traceback after the move; 6 config conflict now blocks pre-mutation; 7 `rewrite_sessions_index` returns a status; 8 nested `cwd` is counted and warned rather than claimed as preserved history.

**Minor** — all taken: `newline=""` both ways (CRLF fidelity verified), swap skipped when nothing changed, `shlex.quote`d reverse hint suppressed after `--merge`, backup stamps de-collided, constants moved to the `:34` block.

Not changed: `norm_path` still doesn't resolve symlinks — agreed with your reasoning that matching the literal recorded string is correct.

Round trip remains byte-identical; portability lint passes.

PR #414, opened 2026-07-29
fix(claude-projects): verify relocate landed and sweep superseded remnants

aaronsb:
## Summary

`relocate` could report success while leaving the old project directory behind. When a session is live during the move, it writes its state records after the directory moved, recreating the old path with a small state-only sidecar — the conversation is safely in the new location, but the old dir never goes away.

`relocate` now verifies its own work: every transcript it moved must exist in the target, and anything remaining in the old directory is accounted for rather than ignored.

## The guard

A leftover is removed **only** when all three hold:

1. same filename as one this command moved,
2. a strictly **larger** counterpart exists in the target, and
3. it contains nothing but recognized session-state records.

Everything else is reported and left in place. The size is re-checked immediately before `unlink` rather than trusted from the scan, so a session that resumed writing keeps its transcript.

The content check is an **allowlist**, not a denylist — this was a bug caught in testing. Asking "does it contain user/assistant records?" classifies an *unreadable* file as safe to delete, which is exactly backwards. Asking "is every record a known state record?" fails safe on unparseable lines and on unfamiliar types.

## Scope

Deliberately limited to the files this command moved and the directory it moved them from. No global scan of other projects on any invocation — an earlier draft added a system-wide stub sweep to `cleanup` and was reverted as the wrong shape.

## Test plan

- [x] True remnant (state-only, smaller, name collision) is swept; old dir then removed
- [x] Leftover **with** conversation records — kept
- [x] Leftover **larger** than its counterpart — kept
- [x] **Unparseable** leftover — kept (this failed before the allowlist fix)
- [x] Leftover with an **unrecognized** record type — kept
- [x] Non-colliding real transcripts still move normally
- [x] Size guard: a file that grew between scan and delete is not removed, content intact
- [x] Missing transcripts after a move are recorded as a failure (non-zero exit)
- [x] Regression: boundary matching, byte-identical round trip, mtime preservation, `~/.claude.json` key order, `list`/`cleanup`/`hygiene`/`stats`
- [x] `bash scripts/check-portability.sh` passes

Verified against the real leftover this was found from: a 631 B state-only sidecar whose 12 MB counterpart lives in the relocated project — correctly classified as superseded.

### DIFF
--- a/tools/claude-projects
+++ b/tools/claude-projects
@@ -944,11 +944,19 @@ def rewrite_claude_json(old, new, merge, apply):
     return "ok", None
 
 
-# Records Claude writes to track session state rather than conversation. A file
-# holding nothing but these is a sidecar, not history.
+# Records Claude writes to track session state rather than conversation, each
+# one regenerated on the next session. A file holding nothing but these is a
+# sidecar, not history.
+#
+# Every entry must be *regenerable*, because a remnant is a divergent tail, not
+# a duplicate: its records were written after the move, so none of them exist in
+# the counterpart. Sweeping always discards a fork, which is only acceptable
+# when nothing is lost. "summary" was in this set and is not regenerable — it
+# carries compaction prose and the leafUuid that --resume reads — and it occurs
+# zero times across this corpus, unlike the six below (74–8136 occurrences).
 STATE_ONLY_TYPES = {
     "last-prompt", "ai-title", "mode", "permission-mode",
-    "bridge-session", "agent-name", "summary",
+    "bridge-session", "agent-name",
 }
 
 
@@ -977,29 +985,39 @@ def state_only(path):
 
 
 def superseded_remnant(src, dst):
-    """Is `src` a leftover whose real content now lives at `dst`?
+    """Size of `src` if it is a leftover whose content now lives at `dst`, else None.
 
     Relocating while a session is live leaves one behind: the session writes its
     state records (mode, ai-title, last-prompt…) after the move, recreating the
     old path. Such a file is only safe to drop when its counterpart is strictly
-    larger *and* it holds nothing but session state — otherwise it is history.
+    larger *and* it holds nothing but regenerable session state.
+
+    Returns a (size, mtime_ns) stamp taken *before* the contents were read, so
+    the caller can tell whether the file changed while we were deciding. Size
+    alone would miss a same-length rewrite. Callers must test `is not None`.
     """
     try:
-        if not dst.is_file() or src.stat().st_size >= dst.stat().st_size:
-            return False
+        if src.is_symlink() or not dst.is_file():
+            return None
+        st = src.stat()                    # stamped before state_only reads it
+        if st.st_size >= dst.stat().st_size:
+            return None
     except OSError:
-        return False
-    return state_only(src)
+        return None
+    return (st.st_size, st.st_mtime_ns) if state_only(src) else None
 
 
-def remove_if_unchanged(path, size):
-    """Unlink only if the file is still exactly `size` bytes.
+def remove_if_unchanged(path, stamp):
+    """Unlink only if the file still matches `stamp` — its (size, mtime_ns).
 
-    Re-checked immediately before unlinking rather than trusted from the earlier
-    scan, so a session that resumed writing in between keeps its transcript.
+    `stamp` must be from *before* the content check that judged this file
+    removable, so that a session writing during that check is caught. A stamp
+    taken afterwards compares the post-race state to itself and protects
+    nothing.
     """
     try:
-        if path.stat().st_size != size:
+        st = path.stat()
+        if (st.st_size, st.st_mtime_ns) != stamp:
             return "changed"
         path.unlink()
         return "removed"
@@ -1024,22 +1042,25 @@ def verify_move(old_dir, work_dir, moved, failures):
 
     # The old directory outlived the move. Anything here is either a remnant a
     # live session recreated, or something we never owned — report, don't guess.
-    swept, kept = 0, []
+    swept, kept, stranded = 0, [], []
     for item in sorted(old_dir.iterdir()):
         if item.is_dir():
             try:
                 item.rmdir()          # only succeeds when already empty
             except OSError:
-                kept.append(item.name)
+                kept.append(item.name + "/")
             continue
-        if item.name in moved and superseded_remnant(item, work_dir / item.name):
-            result = remove_if_unchanged(item, item.stat().st_size)
+        stamp = superseded_remnant(item, work_dir / item.name) if item.name in moved else None
+        if stamp is not None:
+            result = remove_if_unchanged(item, stamp)
             if result == "removed":
                 swept += 1
-            else:
-                kept.append(f"{item.name} ({result})")
+                continue
+            kept.append(f"{item.name} ({result})")
         else:
             kept.append(item.name)
+        if item.suffix == ".jsonl":
+            stranded.append(item.name)
 
     if swept:
         print(f"  {green('swept')} {swept} superseded remnant(s) from {old_dir.name}")
@@ -1049,6 +1070,11 @@ def verify_move(old_dir, work_dir, moved, failures):
             print(f"    {dim(k)}")
         if len(kept) > 5:
             print(f"    {dim(f'… and {len(kept) - 5} more')}")
+    if stranded:
+        # A transcript still under the old slug did not relocate. That is a
+        # failed move, not a note — the run must not report success.
+        failures.append(f"{len(stranded)} transcript(s) stranded in {old_dir.name}: "
+                        + ", ".join(sorted(stranded)[:3]))
     try:
         old_dir.rmdir()
         print(f"  {green('removed')} empty {old_dir.name}")
@@ -1225,9 +1251,16 @@ def cmd_relocate(args):
                         continue
                     if target.exists():
                         # A name collision is usually a remnant of an earlier
-                        # relocation of this same session; verify_move sweeps it.
-                        if not superseded_remnant(item, target):
+                        # relocation of this same session. Decide it here rather
+                        # than deferring: the cwd rewrite below can shrink the
+                        # counterpart, so a later re-check could disagree.
+                        stamp = superseded_remnant(item, target)
+                        if stamp is None:
                             print(f"  {yellow('skip')} {item.name} (already in target)")
+                            continue
+                        result = remove_if_unchanged(item, stamp)
+                        if result != "removed":
+                            print(f"  {yellow('kept')} {item.name} ({result})")
                         continue
                     shutil.move(str(item), str(target))
                 if pending_idx is not None:
--- a/tools/claude-projects
+++ b/tools/claude-projects
@@ -944,6 +944,118 @@ def rewrite_claude_json(old, new, merge, apply):
     return "ok", None
 
 
+# Records Claude writes to track session state rather than conversation. A file
+# holding nothing but these is a sidecar, not history.
+STATE_ONLY_TYPES = {
+    "last-prompt", "ai-title", "mode", "permission-mode",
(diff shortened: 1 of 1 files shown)


aaronsb:
## Review: relocate verification + remnant sweep

Reviewed with an eye on the deletion path specifically. Two findings I'd call blocking, three worth fixing before merge, and a few notes. The overall shape — allowlist, scoped to what this invocation touched, `rmdir` never `rmtree` — is right. The problem is that the size guard doesn't guard the window it was written for.

---

## Blocking

### B1. Argument evaluation order defeats the size guard

**Location**: `tools/claude-projects:1035-1036`

```python
if item.name in moved and superseded_remnant(item, work_dir / item.name):
    result = remove_if_unchanged(item, item.stat().st_size)
```

`remove_if_unchanged`'s docstring says it protects the window where "a session that resumed writing in between keeps its transcript." That window is between `superseded_remnant`'s read of the file and the `unlink`. But `item.stat().st_size` is evaluated **after** `superseded_remnant` returns — on the far side of the window. `remove_if_unchanged` then takes a second stat microseconds later and compares the two. It is comparing the post-race state to itself.

**Reproduced** (module imported, `state_only` wrapped to append after it returns, simulating a live session writing during the read):

- `old/sess.jsonl` = 26 B, `{"type":"mode","mode":"x"}` — classifies state-only
- `new/sess.jsonl` = much larger — strictly larger counterpart
- after `state_only()` returns True, `{"type":"user","message":"IRREPLACEABLE new prompt"}` is appended

```
  swept 1 superseded remnant(s) from old
  removed empty old
src still exists? False
failures: []
```

A file containing a `user` record was deleted, reported as a superseded remnant, and produced no failure. That is exactly the outcome the guard exists to prevent.

**Fix** — take one observation, before the content check, and carry it through. Cleanest is to have `superseded_remnant` return the size it validated:

```python
def superseded_remnant(src, dst):
    """Size of `src` if it is a leftover superseded by `dst`, else None."""
    try:
        st = src.stat()
        if not dst.is_file() or st.st_size >= dst.stat().st_size:
            return None
    except OSError:
        return None
    return st.st_size if state_only(src) else None
```

```python
size = superseded_remnant(item, work_dir / item.name) if item.name in moved else None
if size is not None:
    result = remove_if_unchanged(item, size)
```

The stat now precedes `state_only`, so an append *during* the read yields a stale size and trips the guard. Note both call sites must use `is not None` — a 0-byte remnant returns `0`, which is falsy, so `if not superseded_remnant(...)` at line 1229 would misread it.

Residual: a rewrite to the identical size is still invisible to a size check. These files are append-only in practice so size is adequate — but if you want the guard load-bearing rather than probabilistic, compare `(st_size, st_mtime_ns)`, which is the pattern `rewrite_claude_json` already uses at line 939-941. Consistency win too.

### B2. `summary` does not belong in `STATE_ONLY_TYPES`

**Location**: `tools/claude-projects:951`

The other six entries are regenerable session state: `mode`, `permission-mode`, `bridge-session`, `ai-title`, `agent-name`, `last-prompt`. `summary` is not — it carries compaction summary prose plus a `leafUuid`, and it is what `--resume` reads to reconstruct a compacted conversation. That's content, not state.

This matters because of the conceptual crux underneath the whole guard: **"strictly larger counterpart" is not evidence that `dst` contains `src`'s records.** By construction the remnant is a *divergent tail* — the live session wrote those records *after* the directory moved, so every record in the remnant is guaranteed **absent** from the counterpart. The sweep is never deleting a duplicate; it is always deleting a fork. That is defensible for regenerable state records and only for those. With `summary` in the list, the sweep can discard the only copy of a compaction summary while printing `swept 1 superseded remnant(s)`.

It's also unvalidated by observation, unlike the rest. Scanning 1143 transcripts in this install:

| type | occurrences |
|---|---|
| `last-prompt` | 8126 |
| `mode` | 7943 |
| `ai-title` | 7806 |
| `permission-mode` | 7623 |
| `bridge-session` | 7592 |
| `agent-name` | 74 |
| **`summary`** | **0** |

Six entries are earned by the corpus; the seventh is a guess, and it's the one whose payload is irreplaceable. Suggest dropping it. If it stays, the docstring should say why a summary counts as regenerable, because on its face it doesn't.

Same scan says the list is otherwise well-calibrated: exactly **1 of 1143** files classifies as state-only (`agent-name` + `ai-title`, 0 files over 10 KB). Types like `relocated` (100), `worktree-state` (100), `pr-link` (4849), `queue-operation` (4880), `file-history-snapshot` (2962) are correctly omitted and fail safe.

---

## Should fix before merge

### S1. A stranded transcript is reported but exits 0 with "Done."

**Location**: `tools/claude-projects:1027-1051` vs `1293-1302`

`kept` is printed in yellow; nothing is appended to `failures`. A relocation that left a real transcript behind still prints `Done. Resume with: cd NEW && claude --resume` and exits 0.

**Reproduced** end-to-end against a sandboxed `HOME`. Old project holds a 9200 B `sess.jsonl` of `user` records; target already holds a 27 B `{"type":"mode"}` stub from an earlier relocation:

```
  skip sess.jsonl (already in target)
  moved …-work-A → …-work-B
  left behind 1 item(s) in …-work-A:
    sess.jsonl
  Done. Resume with:  cd …/work/B && claude --resume
EXIT=0
```

The 9200 B conversation is still under the old slug. Resuming in B gets the 27 B stub. Nothing was destroyed — but the exit status, the `moved A → B` line, and the closing line all say the relocation succeeded.

**Fix** — split `kept` by ownership. Names in `moved` are transcripts this command was supposed to relocate and didn't; those are failures. Items we never owned stay informational:

```python
stranded = [k for k in kept_names if k in moved]
if stranded:
    failures.append(f"{len(stranded)} transcript(s) left in {old_dir.name}: "
                    + ", ".join(sorted(stranded)[:3]))
```

That flips the summary to `Relocation incomplete` and exit 1, which is what the situation is.

### S2. The arrival check tests for a name, not for the transcript

**Location**: `tools/claude-projects:1016`

```python
missing = [n for n in moved if not (work_dir / n).exists()]
```

In the merge path a colliding file is never moved, so the name at the target belongs to the *pre-existing* file. The check passes while the transcript this command was asked to relocate sits untouched in `old_dir` — that's the false negative that let S1's run report success. `missing` means "no file by this name at the target," not "the transcript arrived." S1's fix covers the reporting; the variable name should stop implying more than it checks.

Root cause of that scenario, one layer down: at line 1226-1231 the merge resolves collisions by "target always wins," even when the target is the stub and the source is the 12 MB original. `superseded_remnant` already knows how to tell those apart, so the natural symmetry is to also handle `superseded_remnant(target, item)` — target is the remnant — by replacing it. Reasonable to defer that to its own PR, but then S1's exit-code fix is required, not optional, so the run stops claiming success.

### S3. Silent non-skip loses the explanation when the handoff doesn't complete

**Location**: `tools/claude-projects:1226-1231`

When a collision *is* a remnant, nothing is printed and the file is handed to `verify_move`. Usually it gets swept — the handoff does work in the common case. But `verify_move` re-evaluates `superseded_remnant` against a `work_dir` file that the intervening cwd rewrite (line 1249-1255) may have **shrunk**: replacing a long old path with a shorter new one removes bytes from every rewritten record. If the counterpart drops below the remnant's size, the second evaluation returns False and the file lands in `kept` as a bare filename with no explanation — neither `skip (already in target)` nor `swept`, just an unexplained leftover. Reachable for a small pair: src 631 B, dst 700 B, rewrite trims 100 B.

So the answer to "can a file end up neither moved, nor swept, nor reported?" is: it is always reported, but it can be reported without any of the context that would let someone act on it. Either keep a line at the merge site so the two messages tell one story:

```python
print(f"  {dim('deferred')} {item.name} (remnant — sweeping after move)")
```

or defer the collision decision entirely to `verify_move` instead of evaluating it twice against different file states.

---

## Non-blocking

**N1 — symlinks are judged by target, acted on as link.** `superseded_remnant` uses `src.stat()` / `dst.is_file()` (both follow links); `remove_if_unchanged` uses `path.stat()`. A symlink in `old_dir` named like a moved transcript gets its delete decision made from a file that may live anywhere on disk. Verified the `unlink` only removes the link and leaves the target intact, so this is containment rather than data loss — but an `item.is_symlink()` check routing to `kept` would make the sweep honest about what it's evaluating. One line.

**N2 — `state_only` returns True for 0-byte and whitespace-only files.** Combined with `src.st_size < dst.st_size` an empty leftover is swept. That is the right outcome, but it's right by accident; a comment would make it deliberate.

**N3 — convention: `state_only` opens without `newline=""`.** `rewrite_jsonl` (line 768) uses `newline=""` reading the same format. Universal-newline translation means a bare `\r` splits a record here but not there. Fails safe (parse error → keep), so cosmetic — but matching the sibling reader is free.

**N4 — `moved` is "intended to move," not "moved."** In the merge path some of them demonstrably don't. The name pushes the reader toward the wrong reading of both the arrival check and the `item.name in moved` sweep condition.

---

## What's right

- **Directory handling.** `rmdir` only, never `rmtree`. Verified: empty `subagents/` is removed, populated `tool-results/` is kept and reported. No recursive delete anywhere in the sweep.
- **Allowlist over denylist**, and the docstring's reasoning for it is correct. Verified fail-safe on truncated JSON, bare `null`, and a dict with no `type` — all return False.
- **Scope discipline holds.** `verify_move` touches only `old_dir` and `work_dir / item.name`; `superseded_remnant` compares only against the named counterpart. Nothing iterates `PROJECTS_DIR`. The one global iteration in `cmd_relocate` is the pre-existing "similar directories" hint at line 1078, on the not-found error path.
- `old_dir == work_dir` correctly skips the sweep when both paths encode to the same slug, and `verify_move` runs only under `--execute` (dry run returns at 1185).

## Assessment

The design is sound and the failure modes were thought about. But the test plan's "size guard: a file that grew between scan and delete is not removed" passes for the wrong reason — the growth in that test happens outside the window the guard actually covers, and B1 shows the real window is unprotected. Worth a regression test that mutates the file *between* the content check and the unlink, since that's the case the guard is for.

B1 and B2 are the two that gate this: one lets a file with conversation records through, the other widens what counts as disposable to include the one record type that isn't. S1 is small and stops the command from reporting success it didn't achieve.

aaronsb:
Remediated in 4f95b14 — both blocking findings reproduced before fixing, thanks.

**B1 — the size guard was theater.** Confirmed exactly as described: `item.stat().st_size` evaluated *after* `superseded_remnant` returned, comparing the post-race state to itself. Reproduced by appending during the content read; the remnant gained a user record and was deleted anyway. `superseded_remnant` now returns a stamp taken before it reads, callers pass it through and test `is not None` (an empty remnant stamps as `0`).

Went one further: the stamp is `(size, mtime_ns)`, not size alone. Your "what if a session rewrites the file to the same size" case did slip through a pure size check — verified before and after.

**B2 — `summary` removed.** Your framing is the part that matters: the remnant is a *divergent tail*, so the sweep always discards a fork, never a duplicate. That only holds up for regenerable records. Corpus check reproduces your counts — `last-prompt` 8136, `mode` 7953, `ai-title` 7816, `permission-mode` 7633, `bridge-session` 7602, `agent-name` 74, **`summary` 0**. It was the one entry I guessed and the one with an irreplaceable payload.

**S1/S2** — stranded transcripts now land in `failures`; verified your scenario exits 1 (was 0 with `Done.`) and the 9 KB conversation survives. **S3** — the merge path now decides at the collision instead of deferring, so the shrinking-counterpart disagreement can't arise. Also added: `superseded_remnant` refuses symlinks.

On the test-plan note — you were right that the old "grew between scan and delete" test passed for the wrong reason. It now mutates inside the guarded window, in both the append and same-size-rewrite forms.

Full fail-safe matrix passes: real conversation, larger-than-counterpart, unparseable, unknown record type, and `summary` all kept; only the true remnant swept; populated `tool-results/` kept, empty `subagents/` rmdir'd.

Issue #415, opened 2026-07-30
Markdown line handling: core sentence + authoring way + postcheck nudge + shared detector (83% of the corpus is already flat; 23 files drift)


aaronsb:
## What's the problem?

Agents hard-wrap markdown prose at ~80 columns by default. For a system whose primary authoring surface *is* an agent editing markdown, that default is actively hostile:

1. **It breaks `Edit`.** Exact-string replacement means `old_string` has to reproduce arbitrary interior line breaks. A one-sentence change in a wrapped paragraph becomes a multi-line match the agent must transcribe perfectly. Unwrapped, the same edit is a single line. Worse than the token cost: this is where `Edit` calls *miss* and get retried, and a retry is a full round-trip plus another output attempt.
2. **It makes diffs non-semantic.** Changing three words reflows the rest of the paragraph, so `git diff` shows five changed lines for one changed thought. Review cost scales with wrap width, not with the size of the change.
3. **The edit tax is permanent, not one-time.** Re-authoring a wrapped file once is cheap. The wrapped file then imposes the `old_string` cost on every future edit, for the life of the file.

**There is no configuration for this.** Nothing in `settings.json` controls the line width of text an agent writes to a file — the settings surface covers permissions, hooks, env, model, statusline, output style. Terminal rendering wraps at terminal width, but that is display only and never reaches disk. So the available levers are all instruction-level, which is exactly why this is a way plus a check rather than a config change.

**And this repo currently teaches the habit.** `CLAUDE.md` is hard-wrapped at exactly 80 and is in context every session. Wrapped ways get injected verbatim when they fire, so an agent reading `meta/deployment` sees hard-wrapped prose modeled as house style, then writes the next file the same way. The corpus is the training surface for the agent that writes into it — that feedback loop is the reason the repair below is not cosmetic.

### Measured state of the corpus

Detector run over `hooks/ways/**/*.md` (138 files):

- **23 files fire** — contain at least one confirmed hard-wrap run
- **115 files are silent**

Worst by run count: `meta/skills` (5 runs / 16 lines), `meta/deployment` (5 / 15), `meta/start` (4 / 15), `meta/knowledge/optimization/tuning` (4 / 13), `data/migrations/checkpoint` (4 / 13), `meta/goals` (4 / 12), `softwaredev/tooling` (3 / 12), `meta/develop` (3 / 12), `ea/briefing` (3 / 10).

[...]

PR #416, opened 2026-07-30
Markdown line handling: convention, reactive nudge, and a shared detector

aaronsb:
PR1 of #415. Prevention and detection; the lint rule and the corpus sweep are deliberately held for PR2.

## Summary

- **`ways_core::reflow`** — the shared detector and repair transform, plus `ways reflow` as a thin caller. It lives in the library because three callers need the same fingerprint (this command, the way's `postcheck.sh`, and a coming `ways lint` rule); a second implementation would drift, and the failure mode is the way firing on a file the tool then calls clean.
- **`core.md` gains `## Line handling`** next to `## Language`, which is already a mechanical file-output rule of the same class. Core is the only surface present *before* the first write — a postcheck is reactive by construction.
- **`documentation/markdown`** carries the full convention; **`documentation/markdown/reflow`** is the trigger-inert reactive child whose postcheck fires on freshly written wrapped markdown and whose macro names the file.
- **ADR-123 amendment** permitting a postcheck to write session-scoped state for its own macro.

## The design was set by measurement, not reasoning

The precision claim needed evidence, so it was prototyped against the real 138-file corpus first. That changed the design twice:

| Repair scope | Result |
|---|---|
| Whole paragraph | **Modified 16 of the 114 files with no detected wrap** — welded deliberately parallel sentences, and **de-indented nested list items** |
| Detected window only | Left paragraphs half-flat — one joined line, rest still wrapped |
| **Enclosing paragraph of a window** | **24/24 detected files changed, 0/114 others touched** |

**The token-stream oracle is insufficient, demonstrated rather than theorized.** It passed the nested-list de-indent corruption, because leading whitespace is not a token. So indentation and line-initial structure are structural guarantees in the transform, not properties the check verifies — and both regressions are carried as fixtures.

Detection is `>=3` prose lines clustered in `[55,100]` within a 12-char band with `>=2` mid-phrase breaks. The mid-phrase test is what separates column wrapping from deliberate one-clause-per-line prose, which the convention explicitly permits.

## Review focus

- `reflow()`'s block-splitting — the guarantee that authored structure survives.
- `postcheck.sh` on the hot path: `check-post.sh` runs every postcheck on every `Edit`/`Write`/`Bash`/`Task`, so it gates on tool and extension before spawning, and checks for exit code *exactly 1* so a binary error reads as "no finding" rather than firing on every markdown write.
- Exit-code inversion: `ways reflow` uses lint convention (0 clean), `postcheck.sh` wants 0 = fire. One flag can't serve both, so the wrapper inverts.
- The ADR amendment's four constraints, particularly that state is never an input to the predicate — that is what preserves determinism.

## Test plan

- `cargo test` — full workspace green; 16 new fixtures cover tables, fences, blockquote status blocks, setext underlines, indented code, trailing-two-space hard breaks, nested list continuations, and parallel one-sentence-per-line prose (must *not* fire).
- Corpus sweep on a copy: 24 modified, 114 untouched, zero clean files changed, no token mismatches.
- Postcheck end-to-end against synthetic PostToolUse payloads: fires on wrapped markdown and names the file, then stays quiet on the same file again, on flat prose, on a `.rs` path, on a `Bash` tool call, and on an install whose `ways` predates the subcommand.
- `ways lint` 0 errors 0 warnings; `check-portability.sh` passes; `clippy` clean.
- The three new/edited markdown files pass `ways reflow` themselves.

## Not in this PR

The `ways lint` rule and the sweep of the 24 wrapped files land in PR2, so the rule's first observed state is green — shipping it here would print 24 files of warnings on every contributor and CI run, and a surface that nags trains its reader to ignore it.

Refs #415

### DIFF
--- a/docs/architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md
+++ b/docs/architecture/system/ADR-123-firing-dynamics-progression-axis-unification.md
@@ -203,6 +203,25 @@ A new `ways tune` subcommand mirrors `attend tune`. It surveys recent sessions v
 
 Tuning parameters come from real session distributions, not guesses — the same discipline ADR-119 used when it sized attend's parameters to "Claude's actual turn cadence, not biological neuron kinetics."
 
+## Amendment — 2026-07-29: postcheck state handoff to its own macro
+
+Decision 5 describes postcheck evaluation as "metric-predicate style — cheap, deterministic, side-effect-free." That clause bundles three properties, and one of them is narrowed here: a postcheck may write **session-scoped state for the sole purpose of handing a finding to its own way's `macro.sh`**. Cheap and deterministic are unchanged, and the reason they survive is spelled out in constraint 4 below.
+
+**Why the channel is needed at all.** `check-post.sh` invokes each postcheck with `>/dev/null 2>&1` and reads only the exit code — ADR-135's amendment already recorded the consequence: "a postcheck's stdout is discarded... the content injected on a fire is the way's own body." `macro.sh` cannot close the gap either: `show/helpers.rs` runs it with no stdin and no arguments, so it never sees the tool input. `show/mod.rs` applies no trigger condition to macro execution, so a macro *does* run on a postcheck-triggered fire — the ordering (postcheck, then gate, then macro) is guaranteed by `check-post.sh`. A session-scoped stash is therefore the only channel by which a reactive way can name the file it is reacting to. Without it, every reactive way is limited to a static body, which is adequate for a way whose finding *is* its body (over-build names a replacement) and inadequate for one whose finding is a location.
+
+**Constraints, so this does not become a general escape hatch:**
+
+1. **Session-scoped only**, under `sessions-root.sh`. Not project files, not global state, not anything a later session or another agent inherits.
+2. **Own-way only.** A postcheck writes state that its own way's macro consumes. Reading or writing another way's state is out of scope and would reintroduce the coupling Decision 5's separation was protecting.
+3. **Bounded.** State is a set with a cap and a TTL, not an append-only log — because the inward gate may deny the fire that would have consumed an entry, so unconsumed entries are the normal case rather than an error.
+4. **Never an input to the predicate.** The exit code stays a pure function of the observed post-state. State flows one way — postcheck writes, macro reads — so the *predicate* remains deterministic and re-runnable, which is the property the original clause was protecting. A postcheck that consulted its own prior state to decide whether to fire would be a genuine violation, not a narrowing.
+
+**A set, not a single slot.** `refire` keys on the way, not on the artifact the way is reacting to. With a single overwritten slot, a permissive cadence re-fires about a file the operator already declined to act on, and a gate denial leaves a stale path that a later, unrelated fire would name incorrectly. A seen-set addresses both: the macro names what has not yet been surfaced, and the way stays quiet about what has.
+
+Unchanged: reactive requests are not privileged over predictive matches, and both continue through the same inward gate.
+
+First application is the markdown line-handling way (#415), whose postcheck detects hard-wrapped prose in freshly written markdown and whose macro names the file and the remediation command.
+
 ## Consequences
 
 ### Positive
--- a/docs/reference/ways-cli.md
+++ b/docs/reference/ways-cli.md
@@ -202,15 +202,20 @@ ways template itops/alerts -d "alerting runbooks" --global
 
 **Run from:** Project directory to scan all project ways. Pass a specific file path to lint just one file. Add `--global` to lint global ways instead.
 
-**Tells you:** Validation errors and warnings per file against the frontmatter schema. `--fix` auto-corrects multi-line YAML and missing check sections. `--schema` prints the full frontmatter schema reference. `--check` exits non-zero for CI use.
+**Tells you:** Validation errors and warnings per file against the frontmatter schema. `--schema` prints the full frontmatter schema reference. `--check` exits non-zero for CI use.
+
+`--fix` auto-corrects what is fixable, and **takes its scope from `path`, not from the flag** — the same way `eslint --fix` does. Without a path it refuses rather than rewriting every way in the resolved corpus; pass `--all` when that is what you actually want. Each correction is disclosed on its own `FIXED:` line, and fixes are writes, so review them with `git diff`.
 
 ```
 ways lint                                          # scan project ways
 ways lint ~/.claude/hooks/ways/meta/knowledge/knowledge.md  # single file
-ways lint --fix                                    # auto-correct what's fixable
+ways lint <path> --fix                             # auto-correct within that path
+ways lint --fix --all                              # auto-correct the whole resolved corpus
 ways lint --check                                  # CI mode — non-zero exit on errors
 ways lint --schema                                 # show the frontmatter schema
 ```
+
+Exit codes: `0` clean, `1` errors found (with `--check`), `2` the invocation was wrong — kept distinct so a caller can tell "the corpus has problems" from "you used the flag wrong".
 
 ---
 
--- a/hooks/ways/core.md
+++ b/hooks/ways/core.md
@@ -54,6 +54,10 @@ The ledger is not the whole method, and recording a decision does not make it tr
 
 All file output (commit messages, comments, documentation, PR descriptions) must be in English regardless of interface language setting.
 
+## Line handling
+
+Write markdown prose one line per paragraph or list item. Don't hard-wrap to a column — a renderer reflows paragraphs, so wrapping buys nothing and costs edit-ability: an `Edit` then has to reproduce interior line breaks exactly, and a three-word change reflows the paragraph so the diff stops being semantic. Wide content (tables, long links, code) is allowed to be wide. This is about markdown; in plain text the wrap *is* the layout, and commit bodies keep the 72-column convention.
+
 ## Attribution
 
 Do NOT append the Claude Code attribution to commits.
--- a/hooks/ways/documentation/markdown/markdown.md
+++ b/hooks/ways/documentation/markdown/markdown.md
@@ -48,7 +48,9 @@ The reasoning above depends on a renderer that reflows paragraphs. Where there i
 
 `ways reflow <file>` reports hard-wrapped paragraphs and exits non-zero when it finds any; `--fix` flattens them, backs the original up first, and prints the backup path.
 
-It repairs the *enclosing paragraph* of each detection and leaves everything else byte-identical, so it won't flatten authored one-clause-per-line prose elsewhere in the file. Read the diff anyway — the tool's token-stream check catches dropped words but cannot see a join that shouldn't have happened.
+It repairs the *enclosing paragraph* of each detection and leaves everything else byte-identical, so it won't flatten authored one-clause-per-line prose elsewhere in the file.
+
+Before writing, it reparses the result and compares the document's structure against the original. If anything moved beyond the line breaks inside a paragraph, it refuses to write and says so — a construct it misread becomes a refusal rather than a corrupted file. Read the diff anyway: two things that check cannot see are a join that erased an authored line break, and non-CommonMark syntax like `:::` directives.
 
 ## See Also
 
--- a/hooks/ways/documentation/markdown/markdown.md
+++ b/hooks/ways/documentation/markdown/markdown.md
@@ -1,0 +1,58 @@
+---
+description: markdown authoring mechanics — line handling, hard wrapping versus flat prose, when a line break carries structure, tables and fences
+vocabulary: markdown wrap unwrap reflow flatten line length column paragraph prose hard wrap fill width flow text file authoring plaintext txt
+files: \.md$
+scope: agent, subagent
+refire: 0.15
+---
+<!-- epistemic: convention -->
+# Markdown Line Handling
+
+Write prose **one line per paragraph**. One line per list item, one line per table row, one line per blockquote line. Don't hard-wrap prose to a column.
+
+This is a machine-ergonomics convention, not a style preference. Two concrete costs:
+
+- **Hard wrapping breaks `Edit`.** Exact-string replacement means `old_string` has to reproduce every interior line break perfectly. A one-sentence change inside a wrapped paragraph becomes a multi-line match to transcribe; flat, it's a single line. This is where edits *miss* and get retried, not merely where they get long.
+- **Hard wrapping makes diffs non-semantic.** Change three words and the rest of the paragraph reflows, so `git diff` shows five changed lines for one changed thought. Review cost starts scaling with wrap width instead of with the size of the change.
+
+Neither cost is one-time. A flat file stays cheaper to edit for its whole life.
+
+## Line breaks that do carry structure
+
+A line break is fine — required, even — where the break *is* the structure. Keep these on their own lines:
+
+| Construct | Why it stays |
+|---|---|
+| List items | The marker is line-initial syntax |
+| Table rows | Row boundary is the line boundary |
+| Headings, thematic breaks | Block-level, line-initial |
+| Blockquote label lines (`**Status:**`, `**Next:**`) | Reads as a field list; joining makes one run-on line |
+| Deliberate hard breaks (two trailing spaces, or `\`) | The break renders as `<br>` — joining silently deletes it |
+| Parallel one-clause-per-line prose | Authored rhythm, not wrapping |
+
+That last one is the judgment call. The test that separates it from hard wrapping: **does the break land at a clause boundary, or mid-phrase?** Wrapping breaks wherever the column runs out — mid-phrase, with line lengths all clustered just under the fill width. Authored breaks land where the thought turns, and their lengths vary freely.
+
+## Wide content is allowed to be wide
+
+Tables, long links, deep code lines, and long fenced blocks may run past any comfortable reading width. Readers scroll; authors shouldn't reflow. Never "fix" a table by rewrapping its cells.
+
+## Other text formats are not markdown
+
+The reasoning above depends on a renderer that reflows paragraphs. Where there is none, wrapping is the layout and should stay:
+
+- **Plain `.txt`** — no renderer, so the wrap is the presentation. Leave it alone; 72 columns there is correct, not drift.
+- **Commit message bodies** — keep the conventional 72-column wrap. Git tooling assumes it.
+- **Code comments** — follow the language's own line-length convention.
+
+## Repairing a wrapped file
+
+`ways reflow <file>` reports hard-wrapped paragraphs and exits non-zero when it finds any; `--fix` flattens them, backs the original up first, and prints the backup path.
+
+It repairs the *enclosing paragraph* of each detection and leaves everything else byte-identical, so it won't flatten authored one-clause-per-line prose elsewhere in the file. Read the diff anyway — the tool's token-stream check catches dropped words but cannot see a join that shouldn't have happened.
+
+## See Also
+
+- mermaid(documentation) — the sibling convention-plus-tool pair, for diagrams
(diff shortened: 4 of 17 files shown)


aaronsb:
## Review — PR #416

Everything in the test plan reproduces. `cargo test -p ways -p ways-core` green (254 / 7 / 11 / 134), clippy clean, `check-portability.sh` passes, and the corpus sweep on a copy of `hooks/ways` gives exactly the claimed numbers: **24 detected, 24 changed, 0 clean files modified, 0 non-convergent, 140 files**. The exit-code contract holds end-to-end against the *currently installed* `ways 1.5.1`, which predates the subcommand — it exits 2, the postcheck exits 1, the way stays quiet.

So the measurement is real and the shape is right. What follows are cases the 138-file corpus does not contain.

## Blocking

### 1. A bare `>` is a paragraph boundary, and reflow welds across it — with the token oracle reporting clean

`classify` (`tools/ways-core/src/reflow.rs:243-256`) has no case for a blockquote line whose body is empty, so `>` becomes `Prose`. `starts_block` (`reflow.rs:425-429`) splits on quote/non-quote *transitions* but a bare `>` is still a blockquote, so it does not split there either.

```
> The first quoted paragraph runs long enough to sit inside the detect band
> and it continues here mid-phrase so the mid-break count reaches its floor
> and a third line keeps the window at the minimum run length it requires.
>
> A second quoted paragraph, which must remain a separate paragraph always.
```

`ways reflow --fix` collapses this to one line, and reports:

```
  tokens: identical (4 bare '>' marker(s) dropped)
```

Two quoted paragraphs are now one, and the oracle called it clean. This is the `QuoteMarkersDropped` masking case, answered concretely: because the arm accepts *any* difference consisting solely of `>` tokens, it cannot distinguish "dropped the continuation markers of a joined quote" from "dropped the marker that was a paragraph boundary." The `checked_sub` at `reflow.rs:478` is correct and its test is good, but it is guarding the wrong direction — the danger is an *unaccounted* drop, not an underflow.

Note what is currently preventing this from firing more often: a `>` line is one character, so it fails `min_len`, so most windows containing one never qualify. **`min_len: 55` is accidentally load-bearing for a safety property.** That is the argument for fixing it structurally rather than leaving it to a tuning constant.

Suggested fix, in order of value:

- In `classify`, emit `Blank` when `quote_body(raw).trim().is_empty() && is_blockquote(raw)`. It genuinely *is* a paragraph boundary, so this both stops the weld and lets the second quoted paragraph be detected on its own.
- Then narrow the oracle: require `dropped` to equal the number of joins the transform actually performed inside quoted blocks, and return `Mismatch` otherwise. That turns "any `>`-only difference is fine" into an accounted number, which is what makes it an invariant rather than an exemption.

### 2. `detect_with` is O(n³) in paragraph length, on the hot path

`reflow.rs:337-345`. Measured on a single-paragraph bullet list whose band never closes, so no `break 'windows`:

| paragraph lines | wall time |
|---|---|
| 200 | 14 ms |
| 400 | 101 ms |
| 800 | 774 ms |
| 1600 | 6132 ms |

That is ×7.9 per doubling. A "paragraph" here is a maximal run of `Prose` lines, so a long bullet list with no blank lines between items is *one* paragraph — the largest prose run anywhere in this repo is ~132 lines (a vendored llama.cpp README), which is why the corpus never showed it. But `postcheck.sh` runs on arbitrary markdown in any project on every `Edit`/`Write`, so the input is not this corpus.

The fix is one keyword. All three band predicates are monotone in window length — `lo` is non-increasing and `hi` is non-decreasing as `len` grows, so `lo < min_len`, `hi > max_len` and `hi - lo > band` are each monotone-true. A failing window at an offset means every longer window at that offset fails the same way, so the inner `continue` at `reflow.rs:344` can be `break`. I brute-forced 20,000 random length vectors against both variants: **zero divergence** in the set of windows passing the band filter. A paragraph-size cap on top of that would make the hot path unconditionally bounded.

### 3. Raw HTML blocks and tab-indented code are not opaque

`is_html` (`reflow.rs:161-163`) matches only the *opening* line; body lines classify as `Prose`. So a `<pre>` body is joined:

```
<pre>
  some preformatted text line one that is long enough to be within band ok
  more preformatted text line two continuing mid phrase to trigger detect
  third preformatted line so the run length reaches the configured minimum
</pre>
```

becomes one line inside the `<pre>`. Preformatted content is exactly the class where the line break *is* the rendering, which is the property `reflow`'s doc comment promises to protect.

Separately, `leading_spaces` (`reflow.rs:97-99`) counts only `' '`, so a tab-indented code block — valid CommonMark, a tab counts as four columns — never reaches `is_indented_block` and gets joined. Verified.

Both are reachable in this PR, not just PR2: the way body tells the operator to run `ways reflow --fix <file>`.

Fix: track a raw-text HTML block in `classify` for the elements CommonMark treats as literal (`pre`, `script`, `style`, `textarea`), closing on the matching end tag; and expand a leading tab to four columns in `leading_spaces`.

## Should fix

### 4. `is_label_line` swallows ordinary sentences, producing half-flat repairs

`reflow.rs:178-186` marks any trimmed line ending in `:` whose body is only alphanumerics, space, `-`, `_` or `/` as `Verbatim`. That catches `Status:` as intended, but it also catches every prose sentence that ends in a colon without interior punctuation — `so we ended up considering the following three alternatives in sequence:` qualifies.

This already happens in the shipped corpus. `hooks/ways/softwaredev/environment/attend/attend.md:16` (`other two so agents receive consistent guidance at every point:`) is classified `Verbatim`, splitting the paragraph, and the file repairs to one joined line followed by a still-wrapped remainder — the "half-flat" outcome the PR's table says the enclosing-paragraph scope was chosen to avoid. It is a quality-of-repair regression rather than corruption, but it is the one measured failure mode reappearing through a different door.

Bound the predicate: a word count (`head.split_whitespace().count() <= 5`) or a length cap keeps `Status:` and `Next steps:` while releasing sentences.

### 5. Link reference definitions are joined into prose

`is_footnote_def` (`reflow.rs:165-168`) handles `[^x]:` but not `[x]: url`. Three wrapped prose lines followed with no blank line by

```
[adr-142]: https://github.com/aaronsb/agent-ways/blob/main/docs/adr142.md
```

weld the definitions into the paragraph body as literal text, and `token_verdict` returns `Identical`. A *pure* block of definitions is safe — `[` is in neither the lowercase test nor the `“ " ( \`` resume set of `is_mid_phrase_break`, so `mids` stays 0 — so this needs adjacency to wrapped prose, which is uncommon but not rare. Widening the predicate to a leading `[...]:` closes it.

### 6. The seen/pending trim can zero the state file

`postcheck.sh:74-78` writes through a fixed temp name shared by every concurrent invocation in the session:

```bash
tail -n 100 "$f" >"${f}.trim" && mv -f "${f}.trim" "$f"
```

Two postchecks trimming at once both truncate `${f}.trim`; one `mv`s the partial result over `$f` while the other's `mv` fails on a file that is already gone. Reproduced with six concurrent workers over eight rounds: `seen` ended up **0 lines in six of the eight rounds**, not merely short. The consequence is that the seen-set is wiped and the way re-nags about every file the operator already declined — precisely the failure the seen-set exists to prevent, introduced by the constraint-3 bound that protects it.

Concurrency here is real: parallel `Edit` or `Task` calls mean parallel `PostToolUse` hooks, and subagents share the session directory. It also needs >200 accumulated entries, so likelihood is low — but the fix is `${f}.trim.$$`, which makes the rename atomic and costs nothing. The read-then-rename also loses appends landing between `tail` and `mv`; if that matters, take an `flock`, otherwise say so in the comment.

### 7. `macro.sh` consumes `pending` non-atomically

`macro.sh:31-37` reads the file and then `rm -f`s it. A postcheck appending between the loop and the `rm` is lost silently. `mv "$PENDING" "$PENDING.consume.$$"` first, then read the renamed file — rename is atomic, so nothing is lost and nothing is double-reported. Lower severity than #6, since the outcome is a fire that names no file rather than wiped state.

## Verified sound

These were the flagged risks, and they hold up:

- **`set -euo pipefail` interactions.** `[[ cond ]] && exit 1` correctly falls through on a false condition — `set -e` does not fire on a non-final element of an `&&` list. Worth stating explicitly because if it *did*, lines 36 and 48 would make the way never fire at all. The `set +e` / `verdict=$?` / `set -e` bracket reads the pipeline's last status; `pipefail` is inert there because `$?` is taken directly. And `{ read; read; read; } < <(jq ...)` aborts with status 1 on malformed JSON, which is the correct "no fire" direction — confirmed with a garbage payload.
- **Exit-code contract.** Full matrix, all correct: wrapped `.md` → 0 (fire); same file again → 1 (seen); flat prose → 1; `.rs` path → 1; `Bash` tool → 1; own-way path → 1; malformed JSON → 1; binary predating the subcommand (exit 2) → 1. Gating on tool and extension before spawning is right, and checking for *exactly* 1 is what makes an error read as "no finding."
- **ADR-123's four constraints are honored, including the load-bearing one.** State is session-scoped under `sessions-root.sh` (1), own-way only (2), capped and TTL'd (3, modulo #6), and — constraint 4 — never an input to the predicate: `postcheck.sh:48` consults `.seen` only to suppress *re-disclosure*, and does so before the detector runs, so the verdict never depends on stored state. Deleting the state entirely would change how often the way fires, not what `ways reflow` reports about a file. That is the right shape, and the amendment's claim about determinism survives.
- **TTL and staleness.** A 20-minute-old entry is dropped, a non-numeric stamp is skipped, fresh entries are named, `pending` is consumed, the second run is silent, and a missing `CLAUDE_SESSION_ID` exits 0 without guessing. The refusal to infer the session from mtimes is the right call and the `run_macro` change that makes it unnecessary is the right fix.
- **The nested-list de-indent guarantee.** Holds structurally, for the reason claimed: `starts_block` fires on `is_list_item` and `flush` takes the indent from `block[0]`, so an item can only ever be rejoined at its own indent.
- **`is_rule_or_setext` precedence.** Correct as written. `&&` binds tighter, so `reflow.rs:154` reads `(only && count>=3) || (only && (first is '-'|'=') && count>=1)`; the second disjunct subsumes the first for `-`/`=`, leaving the first clause to admit `***` and `___`. Behavior matches intent. Explicit parens would let the next reader confirm that without re-deriving it. Side effect: a bare `--` is treated as a rule though CommonMark makes it a paragraph — the safe direction.
- **`unreachable!` at `cmd/reflow.rs:125`** is genuinely unreachable; the `Mismatch` arm bails at line 100.

## On the constants

Sound independent of the numbers, in the way that matters. The fingerprint — consecutive lines whose lengths cluster tightly, broken mid-phrase — is the real signature of column wrapping, and `min_mid_breaks` carries nearly all the precision. The constants set recall, not safety. Two observations:

- `max_len: 100` with `band: 12` makes a document wrapped at 120 columns invisible, and 100/120 are common editor defaults. A recall gap, which the module explicitly accepts.
- `min_len: 55` is doing unadvertised safety work, as noted in #1. Worth not depending on.

## Smaller notes

- `classify` (`reflow.rs:205-216`): a file whose first line is `---` as a *thematic break* with no second `---` turns the entire file opaque and detection silently returns clean. Verified. Safe direction, and nearly unreachable in a repo where a leading `---` is always frontmatter — but it fails silently rather than visibly.
- `write_backup` (`cmd/reflow.rs:152`) is the only `mktemp` shell-out in the Rust tree; every other temp-dir need uses `std::env::temp_dir()` (a dozen call sites across `attend*`). The stated rationale — no predictable path, no symlink surface — is fully satisfied by `temp_dir().join(unique)` plus `create_dir`, which fails if the directory already exists. That also drops a spawn and a PATH dependency a native `ways.exe` would not have; the portability workflow runs Git Bash on the Windows leg, so it would not catch it.
- `--quiet` still emits JSON when `--json` is passed (`cmd/reflow.rs:57-73`), while the help text reads "exit code only." Defensible, since JSON is not human-readable output, but worth pinning the intent since `postcheck.sh` depends on `--quiet` alone.

## What is good here

The heuristic/guarantee split is the right architecture and the doc comments earn it: detection is allowed to be a heuristic because the *transform* carries the structural guarantees, and the module says plainly that the token oracle is insufficient and why. Carrying both discovered regressions as fixtures (`never_deindents_a_nested_list`, `leaves_bold_definition_lines_alone`) is how a measurement should be encoded — the nested-list one in particular documents a corruption the oracle provably cannot see.

`min_mid_breaks` is the load-bearing idea and it is the correct one: it is what separates column wrapping from the one-clause-per-line style the convention deliberately permits, and it does not depend on the fitted constants. Holding the lint rule for PR2 so its first observed state is green is also right — a surface that nags on arrival trains its reader to ignore it.

Findings 1, 2 and 3 are all cases where the corpus was silent rather than where the design was wrong. 1 and 3 corrupt documents through the `--fix` path this PR ships; 2 puts a multi-second stall on a hook that runs on every write. I would fix those three before merge, and 4-7 before the PR2 sweep turns `--fix` loose on 24 files.

---

aaronsb:
## Remediation — all 7 findings addressed

Every finding is fixed and carried as a fixture (23 reflow tests now, up from 16). Pushed in 854be47.

| # | Finding | Disposition |
|---|---|---|
| 1 | Bare `>` welds quoted paragraphs; oracle reports clean | **Fixed** — `classify` treats an empty blockquote line as `Blank`; oracle tightened (below) |
| 2 | `detect_with` O(n³) on the hot path | **Fixed** — monotone `break`, lengths hoisted per paragraph |
| 3 | Raw HTML bodies and tab-indented code joined | **Fixed** — raw-text block tracking; `leading_columns` counts a tab as four |
| 4 | `is_label_line` swallows prose sentences | **Fixed** — bounded to five words |
| 5 | Link reference definitions joined | **Fixed** — `is_reference_def` widened to `[label]:` |
| 6 | Shared trim temp name empties state | **Fixed** — per-PID temp, atomic rename |
| 7 | `macro.sh` consumes `pending` non-atomically | **Fixed** — claims by rename before reading |

### On #1 — the oracle became an invariant

Your framing was the useful part: the `checked_sub` was guarding the wrong direction, and the real hazard is an *unaccounted* drop. Both halves are now in:

- `classify` emits `Blank` for an empty blockquote line — it genuinely is a paragraph boundary, so the weld stops and the following quoted paragraph becomes detectable on its own.
- `reflow_with_stats()` reports how many blockquote continuations it actually joined, and `token_verdict_accounted()` requires the dropped `>` count to equal it. Anything else is `Mismatch`. `--fix` uses the accounted form.

The note that **`min_len: 55` was accidentally load-bearing for a safety property** is why this is structural rather than a constant nudge. That was the sharpest observation in the review.

### On #2 — the monotonicity argument

Verified rather than accepted: `lo` is non-increasing and `hi` non-decreasing as the window grows, so `lo < min_len`, `hi > max_len` and `hi - lo > band` each fail monotonically. A fixture now asserts 1600 lines detect in under 500 ms.

### On #4 — this was live, not hypothetical

Confirmed at `attend.md:16` exactly as you described: the paragraph split and repaired half-flat. That is the measured failure mode returning through a different door, which makes it the most valuable of the non-blocking findings.

### On #6 — honest note about verification

The fix is in and verified working: 6 concurrent workers × 8 rounds, `seen` trims to 100 every time, never empty. **I could not reproduce the pre-fix failure in my own harness** — both my reconstruction and the true pre-fix line showed no trim at all, which points at my harness rather than at your measurement. I took the fix on your reproduction plus the mechanics (shared temp name, concurrent truncate, racing renames), since it is a one-line change that costs nothing. Flagging it so the record is accurate rather than implying I confirmed the bug independently. The read-then-rename window you noted remains, and the consequence is bounded — one entry that may re-fire once, not wiped state — so no `flock`.

### Also taken

`write_backup` now uses `std::env::temp_dir()` with `create_dir` instead of shelling out to `mktemp`: matches the rest of the tree, drops a spawn from the repair path, and needs no PATH — which a native Windows build wouldn't have. `create_dir` rather than `create_dir_all` is what refuses a planted symlink, so the original rationale is preserved. Explicit parens added to `is_rule_or_setext`.

### Not changed

- **`max_len: 100` missing 120-column wrapping** — a recall gap the module already documents. Widening it without re-measuring precision is the trade the whole design is trying not to make blind.
- **Leading `---` as a thematic break making a file opaque** — safe direction, and unreachable in a corpus where a leading `---` is always frontmatter. Recorded rather than fixed.
- **`--quiet` with `--json`** — JSON is machine output, not human output, so they compose. `postcheck.sh` depends on `--quiet` alone.

### Re-verified after the fixes

Corpus sweep unchanged: 24 detected, 24 modified, **0 clean files touched**, and a second `--fix` pass changes nothing. 38 suites green, clippy clean, `ways lint` 0/0, portability passes.

Findings 1–3 were real corruption paths through the `--fix` this PR ships, and #2 a multi-second stall on a per-write hook. Good review.

aaronsb:
## Reworked onto a real parser

Following the prior-art question: we never searched for one. Doing that turned up [`flowmark`](https://github.com/jlevy/flowmark) — which on inspection is a thin formatter over **comrak**, with 386 total downloads against comrak's 6.4M and pulldown-cmark's 126M. Its own manifest pointed at the right answer: take the parser, skip the wrapper. (That lesson is now its own way, #417.)

### What changed

`pulldown-cmark` (`default-features = false`) replaces the hand-rolled classifier, and adds a structural post-condition.

Three layers, and only the middle one is ours:

| layer | source | job |
|---|---|---|
| Eligibility | **parser** | fences, frontmatter, headings, rules, tables, HTML blocks *and comments*, indented code, tabs, reference definitions |
| Whether to touch | **ours** | the band + mid-phrase heuristic — no parser can answer this |
| Whether the result is sound | **parser** | `structure_preserved()` reparses and compares; `--fix` refuses to write on mismatch |

The third layer is the important one: it does not depend on having enumerated markdown correctly. An enumeration gap becomes a *refusal* instead of a corrupted file.

### It found a live bug in this PR

Run across the corpus, the post-condition immediately flagged `softwaredev/environment/attend/attend.md`: a **multi-line HTML comment** whose body the transform was joining. The token check passed it. PR2's sweep would have shipped it.

That file now classifies as opaque, so detection goes **24 → 23**. The drop is the fix, not a regression.

Measured against deliberate corruptions:

| corruption | token check | post-condition |
|---|---|---|
| nested-list de-indent | missed | **caught** |
| bare `>` weld | missed | **caught** |
| code fence body joined | missed | **caught** |
| table row welded | missed | **caught** |
| setext underline consumed | caught | caught |
| legitimate paragraph join | pass | **pass** — no false positive |

### Two review findings the parser corrected rather than patched

- **`is_label_line` is gone.** It existed to protect `Status:` and was swallowing ordinary sentences (finding 4). The parser has no such notion — it is all one paragraph — so the fix was deletion, not a word-count bound.
- **Reference definitions were guarded on the wrong case** (finding 5). CommonMark says a link reference definition *cannot interrupt a paragraph*, so `[x]: url` directly under prose is already paragraph text and joining it changes nothing. A real definition after a blank line is its own block and was never at risk.

### Known blind spots, both principled

- **Welding authored line breaks** is invisible to any structural check — two sentences joined are the *same* CommonMark document. That is precisely what the detection heuristic is for. Prevention, not verification.
- **Non-CommonMark syntax** (`:::` directives, MDX, Liquid) is paragraph text to any CommonMark parser. Covered by an explicit `EXTENSION_DELIMITERS` list — short and enumerable, which is the point of adopting a parser: the unknown surface shrinks to a nameable set.

### Verification

- 29 reflow fixtures, 38 suites green, clippy clean, `ways lint` 0/0, portability passes
- Corpus sweep: 23 detected, 23 modified, **0 clean files touched, 0 post-condition refusals**
- Postcheck end-to-end unchanged: fires on wrapped markdown, names the file, stays quiet on repeat / flat prose / `.rs` / `Bash` / old binary
- Way body updated to state the refusal guarantee and both blind spots

This also reverses my earlier recommendation to drop `--fix`. With a post-condition that refuses to write, the failure mode inverts from "I hope the enumeration was complete" to "the write only lands if the parser agrees nothing structural moved."

PR #418, opened 2026-07-30
Sweep the 23 hard-wrapped ways, and fix what the post-merge review found

aaronsb:
## Summary

Runs `ways reflow --fix` over the 23 hard-wrapped files in `hooks/ways`, so the shipped corpus stops contradicting the convention `core.md` introduced in #416.

This matters more than tidiness: `hooks/ways` is the projected artifact, and a wrapped way gets **injected into user sessions as guidance**. An agent reads it, then writes the next file the same way. That feedback loop is what produced the drift, and it is what this closes.

This is also the tool's first real use — the dogfooding pass that #415 planned as the integration test.

## Result

| | |
|---|---|
| detected | 23 |
| fixed | 23 |
| declined by the post-condition | 0 |
| still detected afterward | 0 |
| structurally preserved | **23 / 23** |

80 insertions, 314 deletions. The deletions are the wrap.

## Verification

Not taken on the tool's own say-so. Each swept file was reparsed with a **separate probe** and compared against its pre-sweep structure — same blocks, same nesting, same code-block content, differing only in line breaks inside paragraphs.

Diffs were also read by eye, because the structural check has one blind spot it cannot cover: a join that erases an authored line break produces an identical CommonMark document. That is the detector's job to avoid, not the verifier's to catch.

`ways lint --check` 0/0, portability passes.

## Not in this PR

The `ways lint` reflow rule. Detection and flagging already ship via the postcheck; the lint surface is the third and least load-bearing of the three, and it is worth landing on its own rather than alongside a 23-file prose diff.

### DIFF
--- a/hooks/ways/data/data.md
+++ b/hooks/ways/data/data.md
@@ -8,11 +8,7 @@ refire: 0.15
 <!-- epistemic: premise -->
 # Data
 
-**Scope the data layer to the actual need before reaching for machinery.** The
-cost of a database is not the store — it's the discipline the store pulls in:
-migrations, schema docs, backups, review. That discipline is worth it when the
-data is shared and long-lived, and pure overhead when it isn't. Match the tool
-to the pressure, not to habit.
+**Scope the data layer to the actual need before reaching for machinery.** The cost of a database is not the store — it's the discipline the store pulls in: migrations, schema docs, backups, review. That discipline is worth it when the data is shared and long-lived, and pure overhead when it isn't. Match the tool to the pressure, not to habit.
 
 | The data is… | Reach for | Skip |
 |---|---|---|
@@ -21,10 +17,7 @@ to the pressure, not to habit.
 | Shared, evolving, multi-writer | a server RDBMS (Postgres/MySQL) with **versioned migrations** | — |
 | Densely interconnected, traversal-first | a graph store (or a graph extension) | forcing it into flat tables |
 
-The deeper children below are for the bottom two rows. If someone says "let's
-just use SQLite for this," they need *this* orientation — not a lecture on
-migration consolidation or ER-diagram generation. Don't pull the heavy guidance
-until the need is real.
+The deeper children below are for the bottom two rows. If someone says "let's just use SQLite for this," they need *this* orientation — not a lecture on migration consolidation or ER-diagram generation. Don't pull the heavy guidance until the need is real.
 
 ## Going deeper
 
--- a/hooks/ways/data/documentation/documentation.md
+++ b/hooks/ways/data/documentation/documentation.md
@@ -27,11 +27,7 @@ from the DDL — the source of truth — so they can't lie.**
 
 ## Diagrams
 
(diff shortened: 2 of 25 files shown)


aaronsb:
## Scope grew: this PR now also carries the post-merge review remediation

The review of `854be47..main` landed while this was open, and two of its findings are **silent corruption on the `--fix` path** — the same `--fix` this PR uses to sweep 23 files. Fixing them here rather than merging a sweep performed by a tool with known corruption paths.

The reviewer built an independent oracle (render before/after with `pulldown_cmark::html`, diff) and ran ~20k generated cases plus every `.md` in the tree through it. **The post-condition held — zero false negatives, zero false positives.** All four defects were elsewhere.

### Fixed

| # | Defect | Class |
|---|---|---|
| 1 | Extension delimiters not seen through `>` — `> :::note` welded | **corruption** |
| 2 | CRLF `\r` stripped on repaired paragraphs only | **corruption** |
| 3 | Write-path IO errors exited 1, colliding with "declined to repair" | contract |
| 4 | `--fix --json` emitted two documents, or JSON then raw markdown | contract |

Both corruption reproductions re-run clean. A fixture per defect; 153 tests green.

Advisories taken, each closing an *enumeration gap* rather than adding a rule: `Event::Code` gets its own fingerprint arm; `ENABLE_MATH` / `ENABLE_PLUSES_DELIMITED_METADATA_BLOCKS`; the duplicated parser-options block collapsed to one (classifier and post-condition disagreeing about what the document *is* is the one coupling here that must not break); zero-length events skipped explicitly rather than relying on `RangeInclusive` degenerating `n..=n-1`.

### The sweep was re-verified against the fixed tool

Since it had been performed with the buggy binary, I re-ran it from the pre-sweep state and diffed. **One file differed** — `data/migrations/idempotent/idempotent.md`, which contains `DO $$ BEGIN IF NOT EXISTS`. With `ENABLE_MATH` on, `$$` parses as display math and the following prose goes opaque, so two paragraphs are no longer repaired.

So `ENABLE_MATH` is not free, contrary to the advisory: `$$` is also PostgreSQL dollar-quoting and bash's PID. Two of 141 corpus files hit it. Keeping the option — a missed repair is a wrapped paragraph, a bad one is a broken document — but the comment now states the cost rather than repeating that there isn't one.

The committed sweep output stands: that file detects clean under the fixed tool and verifies structurally identical to its pre-sweep state.

### Verification

- 153 tests, clippy clean, `ways lint --check` 0/0, portability passes
- Corpus: **0 hard-wrapped files remain**
- Content integrity across the sweep: 12,325 → 12,320 words, delta = exactly 5 blockquote continuation markers, **prose loss 0**
- Structural check on all 23 swept files with an independent probe: **23/23 preserved**

Deferred to a follow-up, per the reviewer: footnote bodies are over-conservatively `Verbatim`; `repair` re-parses ~6× per `--fix` (irrelevant on the read-only hook path); `ways lint --global --fix` still asks for `--all` without the message saying so.

PR #435, opened 2026-08-02
fix(reflow): lower the detection floor to the wrap shapes agents actually write

aaronsb:
## Problem

The wrap detector's floor sat exactly above the shapes the model most often produces, so the reactive `documentation/markdown/reflow` way never fired: zero postcheck state across every session since boot, while wrapped markdown kept shipping. Upstream context: anthropics/claude-code#33666 confirms the model hard-wraps at ~75–80 columns regardless of instructions, so the reactive layer is the load-bearing one — and it was blind to:

- **Two-line wraps** — at most one mid-phrase break; the floor demanded two.
- **Three-line wraps with a short tail** — the tail sat in every candidate window (`min_run: 3`) and failed `min_len` for all of them.
- **Wrapped list items** — a list was one "paragraph" whose item-boundary breaks never count as mid-phrase.

## Fix

- `min_mid_breaks` becomes a cap: a window is asked for as many mid-phrase breaks as it has line breaks; `min_run` drops to 2.
- A unit's final line is exempt from the length band — a wrap tail is short by construction.
- Detection now works on the same *units* the repair joins (paragraphs split at list items, labels, quote transitions, hard breaks), via a `starts_join_block` predicate shared with the flush loop so the scopes cannot drift.

## Validation

- 158 ways-core tests pass, including five new regression tests for the blind spots and two authored-prose guards.
- Full corpus diff old-vs-new: 25 newly detected files, **every finding a genuine mid-phrase wrap, zero false positives** on authored one-clause-per-line prose.
- All repairs pass the existing token and structure gates; sweep commit flattens the 25 plus two post-#418 drifts, leaving hooks/ways fully clean.

## Out of scope, noted for follow-up

- docs/ carries ~90 files the old detector already flagged (never swept) — separate chore.
- `postcheck.sh` appends to the seen-set at detection time, but ADR-123's amendment says the way should stay quiet only about what has been *surfaced* — a gate-denied fire currently suppresses the file for the whole session unseen. Fixing that touches the amendment's one-way state-flow constraint, so it wants its own ADR discussion.

### DIFF
--- a/docs/install-guide.md
+++ b/docs/install-guide.md
@@ -60,9 +60,7 @@ git remote add upstream https://github.com/aaronsb/agent-ways
 ./scripts/install.sh
 ```
 
-Running the installer from inside the app dir is what links the suite binaries
-(`ways`, `ways-audit`, `attend`, `attend-chat`) onto your `PATH`; `make setup`
-alone builds them but does not. Pull upstream improvements later:
+Running the installer from inside the app dir is what links the suite binaries (`ways`, `ways-audit`, `attend`, `attend-chat`) onto your `PATH`; `make setup` alone builds them but does not. Pull upstream improvements later:
 
 ```bash
 cd "$XDG_DATA_HOME/agent-ways"
--- a/docs/migration-1.0.md
+++ b/docs/migration-1.0.md
@@ -33,22 +33,15 @@ Background: [ADR-142](architecture/system/ADR-142-agent-ways-1-0-xdg-application
 
 ## Do I need to migrate?
 
-You're a **legacy in-place** install — the one population this targets — if `~/.claude/.git`
-exists and its `origin` points at `aaronsb/agent-ways` (or your fork). That's the
-clone-in-place model 1.0 supersedes.
+You're a **legacy in-place** install — the one population this targets — if `~/.claude/.git` exists and its `origin` points at `aaronsb/agent-ways` (or your fork). That's the clone-in-place model 1.0 supersedes.
 
 If you started fresh on 1.0, or never made `~/.claude` a git repo, there's nothing to migrate.
 
 ## Before you start
 
-The migrator is built to be safe — it **backs up `~/.claude` first** (to
-`~/.claude.backup-<epoch>`), reconciles the new shape **before** removing the old one, and
-is **resumable**: if a phase aborts it stamps a marker, and re-running `--execute` continues
-where it stopped rather than starting over.
+The migrator is built to be safe — it **backs up `~/.claude` first** (to `~/.claude.backup-<epoch>`), reconciles the new shape **before** removing the old one, and is **resumable**: if a phase aborts it stamps a marker, and re-running `--execute` continues where it stopped rather than starting over.
 
-That said, the one genuinely irreplaceable thing is **`~/.claude/projects/`** — your session
-transcripts. It's preserved in place, but snapshotting it yourself costs nothing and is
-cheap insurance for any migration:
+That said, the one genuinely irreplaceable thing is **`~/.claude/projects/`** — your session transcripts. It's preserved in place, but snapshotting it yourself costs nothing and is cheap insurance for any migration:
 
 ```bash
 cp -a ~/.claude/projects ~/claude-projects-keepsake
@@ -73,9 +66,7 @@ ways migrate            # the plan (default): what would change, read-only
 ways migrate --what-if  # dry-run the executor: assert every phase's contract without writing
 ```
 
-`--what-if` is the stronger check — it runs the executor's logic against your real install
-and verifies each phase *could* succeed, without touching anything. Read its output before
-proceeding.
+`--what-if` is the stronger check — it runs the executor's logic against your real install and verifies each phase *could* succeed, without touching anything. Read its output before proceeding.
 
 **3. Execute.** Gated, backs up first, resumable:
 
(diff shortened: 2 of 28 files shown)


aaronsb:
## Review: detection-floor fix (b6f2a94) + corpus sweep (e9a6c7b)

Scope reviewed: `git diff main...HEAD`, full read of `tools/ways-core/src/reflow.rs` and `tools/ways-cli/src/cmd/reflow.rs`, all 27 sweep hunks, 158-test run (pass), plus differential probes against a main-branch build of the detector.

### Finding 1 — BOUNDARY_END gaps are now single-handedly sufficient to fire (High)

**Location**: `tools/ways-core/src/reflow.rs:118` (`BOUNDARY_END`), interacting with the mids cap at `:484` and the tail exemption at `:450`.

**Problem**: Under the old floor, a break at a character missing from `BOUNDARY_END` needed a second corroborating mid-phrase break inside a fully-banded 3-line window. Now one such break in a 2-line window — with only *one* length-checked line when the tail is short — fires on its own. `BOUNDARY_END` protects `*`, `)`, `"` but not their siblings `_`, `]`, `”`, `’`, `'`, `—`, `–`. Verified differentially (old binary = clean, new = fires and joins):

- em-dash line end: `…the make targets —` / `it was the discipline…`
- curly close quote: `…on fire at the moment now”` / `and then the pager…`
- reference bracket: `…is recorded in [ADR-142]` / `and the migration mechanics…`

**Escalation**: because detection is per-window but repair flattens the whole unit, one such break inside an otherwise authored unit welds *every* authored break in it. Probe (old = clean, new = flattens all four lines to one, erasing three breaks that end in `.`):

```
If the check passes: merge it and move on to the next branch in queue.
If it fails: read the log before touching anything else in the tree —
the failure names the construct, and guessing wastes the whole afternoon.
If the log is empty: rerun with tracing enabled and capture the output.
```

This is exactly the shape the module's own bias forbids firing on ("firing on deliberate structure is not" acceptable), and the two new guard tests (comma end, capitalized start) don't cover it.

**Live corpus evidence both ways**: `hooks/ways/softwaredev/code/quality/quality.md:44` is an authored em-dash line end that survives only because the next line starts with `**`; `docs/finish-install.md:74` is the same character shape as a *genuine* wrap. So extending `BOUNDARY_END` costs only two-line wraps that break exactly at such a character — longer wraps still fire on their other windows — which is the trade the module says to make (miss a wrap, never weld structure).

**Suggestion**: add `—`, `–`, `”`, `’`, `'`, `_`, `]`, `}` (and arguably `…`) to `BOUNDARY_END`, plus guard tests for the em-dash and whole-unit-weld shapes above. Fine as an immediate follow-up if you'd rather not re-sweep in this PR, but it should land before much new prose is authored against the loosened floor — the sweep validated the corpus as it exists, not the prose to come.

### Finding 2 — dead clause in the flush condition (Low)

**Location**: `tools/ways-core/src/reflow.rs:585`.

`(block.is_empty() && starts_join_block(line, line))` can only ever invoke `flush` on an empty block, and `flush` early-returns on empty (`:545`). The clause is a no-op in every reachable case and can be deleted with zero behavior change; keeping it implies it does something. While verifying this I also confirmed the `has_hard_break(prev)` addition to `starts_join_block` causes no drift in `reflow`'s own loop: a hard-break line is flushed immediately after push (`:590-592`), so `block.last()` can never carry one when the next line is tested — no double-flush, byte-identical behavior on old inputs.

### Finding 3 — doc drift on WrappedParagraph (Info)

`WrappedParagraph::start/end` (`:74-77`) still say "enclosing paragraph," and the module heading at `:41` still says "scoped to the enclosing paragraph"; both are unit bounds now. `ways reflow --json` output semantics shift with them (harmless — the postcheck consumes exit codes only).

### What I verified and found sound

- **Window scan** (`:452-495`): `enumerate().take(offset+len).skip(offset+upto)` yields exactly `[offset+upto, offset+len)` — full window on the first iteration, the single new line after. Indices are unit-relative, matching `exempt`'s `n-1`. The monotone early `break` survives the exemption (skipping the tail never removes a line from `lo`/`hi`, so the band predicate stays monotone per offset). The `lo > hi` guard is only reachable at `min_run <= 1` and `continue` is the right move there; `min_run: 0` configs don't panic.
- **`units()`** (`:358-386`): `close`'s `e > s` matches the old single-line-paragraph drop exactly; in the `(Prose, Some(s))` arm `lines[i-1]` is guaranteed prose, so the `prev` reference is sound; the quote-transition inequality is equivalent to the old two clauses.
- **Convergence** (item 4): a detected unit always joins to exactly one line (no mid-unit boundary can exist by construction), the joined line becomes a single-line unit `units()` drops, every unit boundary persists post-join (list/`**`/`>` prefixes preserved by `flush`; trailing two-space and backslash hard breaks retained at `:565-573`), and all four boundary causes are precisely `is_mid_phrase_break`'s exclusions — so a joined line can never form a new mid-phrase pair with a neighbor. No re-trigger loop; the CLI's refuse-to-write backstop (`cmd/reflow.rs:113-134`) stays theoretical. One deliberate narrowing worth knowing: wrapped-but-undetected sibling units in the same paragraph no longer get joined "for free" as they did under paragraph-scoped repair.
- **Sweep** (e9a6c7b): read all 27 files' hunks; every join is a genuine mid-phrase wrap. The develop.md `uncertainty-location` hyphen hand-fix is correct in the committed result. Re-ran the new detector over `hooks/ways`: 0 files flag. Two expected half-flat residues from the `**` rule, not sweep errors, both stable and un-nagging: `hooks/ways/meta/deployment/deployment.md` ("…is now / `**legacy**`: …") and `hooks/ways/softwaredev/delivery/merge/merge.md` ("…— and by / `**whether the operator…`"). They keep a mid-sentence source break the tool can never fix; hand-edit if it grates.

### Assessment

The detector rework is careful and the shared-predicate design (detection scope == join scope) is the right fix for the drift class. Findings 2 and 3 are cosmetic. Finding 1 is real: the loosened floor transfers the entire false-positive burden onto `BOUNDARY_END`, and that list has inconsistent coverage today, with a demonstrated whole-unit weld on authored parallel prose. Merge is defensible given the corpus-level validation, but the `BOUNDARY_END` extension and its guard tests should be treated as part of this change, not backlog.

---

aaronsb:
Review findings closed in e2b84c3: BOUNDARY_END extended with the seven missing pause characters (`—`, `–`, `”`, `’`, `'`, `_`, `]`) with a guard test per probed shape; the dead empty-block clause in the flush condition removed; paragraph→unit wording finished in the WrappedParagraph docs and module heading. Re-verified: all four reviewer probes clean, 159 tests pass, clippy clean, hooks/ways still 0 flags, and docs/finish-install.md (the genuine wrap of the em-dash shape) stays detected through its other breaks.

PR #436, opened 2026-08-03
fix(scan): emit the canonical hookSpecificOutput envelope on SessionStart

aaronsb:
## Problem

The base posture (`core.md`) and ways catalog have never reached the model. Transcript evidence across every session in `~/.claude/projects/-home-aaron-Projects-ai-agent-ways/`:

- The SessionStart hook chain runs and `hook_success` records its stdout (the full ~19KB payload) — on every session.
- Zero `hook_additional_context` attachments for SessionStart, and the payload appears in zero model-visible messages (`grep` across all user/system message content: no hits).
- Canonical-envelope paths (UserPromptSubmit, PostToolUse, SubagentStart) all deliver, same sessions.

The felt symptom — "core way doesn't carry the same weight as CLAUDE.md" — was total non-delivery, not attention decay. CLAUDE.md is re-supplied by the harness every context window with authority framing; the ways posture was arriving via zero paths.

## Cause

`emit_hook_context` emitted a bare top-level `{"additionalContext": ...}` on SessionStart, believed to be a legacy tolerance. The current hooks reference (code.claude.com/docs/en/hooks.md) documents exactly two SessionStart output shapes: plain text, or `{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": ...}}`. Undocumented JSON stdout is dropped silently.

## Fix

- SessionStart joins the canonical branch; scenario 10's envelope guard flipped to lock canonical in.
- `--hook-event` doc comment updated.
- PreToolUse deliberately untouched: its legacy shape carries `decision: approve` (permission semantics), and transcripts suggest its `additionalContext` may also be dropped — follow-up investigation, not a blind flip.

## Verification

`ways scan state --hook-event=SessionStart` emits valid canonical JSON with the full catalog + posture (18.6KB); all session_sim scenarios pass; clippy clean. Real-world confirmation lands on the next session start after the installed binary rebuilds: the transcript should show a SessionStart `hook_additional_context` attachment for the first time.

### DIFF
--- a/tools/ways-cli/src/cmd/scan/mod.rs
+++ b/tools/ways-cli/src/cmd/scan/mod.rs
@@ -75,11 +75,10 @@ impl WayCandidate {
 /// Match user prompt against ways and emit matched bodies for the agent.
 ///
 /// Wired only from the `UserPromptSubmit` hook (`check-prompt.sh`), so the
-/// envelope event name is hardcoded. The call routes through the canonical
-/// `hookSpecificOutput` default branch of `emit_hook_context`. If this is
-/// ever reused from another hook event, just pass that event's name —
-/// `SessionStart` and `PreToolUse` are the only events that take the
-/// legacy top-level `additionalContext` envelope.
+/// envelope event name is hardcoded. The call routes through
+/// `emit_hook_context`, which emits the canonical `hookSpecificOutput`
+/// envelope for every event. If this is ever reused from another hook
+/// event, just pass that event's name.
 ///
 /// `response_context` (ADR-155 §3) is Claude's last response, stored raw by
 /// the Stop hook. It feeds ONLY the embed input — never the regex lane — so
@@ -1036,34 +1035,25 @@ fn regex_span(pattern: &str, text: &str) -> Option<String> {
     })
 }
 
-/// Emit accumulated context using the envelope shape required by the
-/// invoking hook event. The Claude Code hook contract treats
-/// `hookSpecificOutput` as canonical for all events, and the current hooks
-/// reference documents no other JSON shape. The bare top-level
-/// `additionalContext` this branch once emitted for `SessionStart` was
-/// believed to be a legacy tolerance — session transcripts proved otherwise:
-/// the harness records the stdout in its `hook_success` bookkeeping and
-/// creates no context attachment, so the payload (the ways catalog and the
-/// core posture) never reached the model in any session on record. Canonical
-/// everywhere is what the delivered paths (UserPromptSubmit, PostToolUse,
-/// SubagentStart in the shell emitters) already use.
+/// Emit accumulated context in the canonical `hookSpecificOutput` envelope —
+/// the only JSON shape the current hooks reference documents, for any event.
 ///
-/// `PreToolUse` keeps its `decision`-bearing shape for now: its envelope is
-/// entangled with permission semantics, and transcripts suggest its
-/// `additionalContext` may be dropped too — tracked separately rather than
-/// changed blind here.
+/// A bare top-level `additionalContext` was emitted for `SessionStart` here
+/// until session transcripts proved the harness never delivered it: the
+/// stdout landed in `hook_success` bookkeeping, no context attachment was
+/// created, and the payload (the ways catalog and the core posture) reached
+/// zero sessions on record. Undocumented JSON stdout is dropped silently, so
+/// nothing but the canonical envelope belongs here. The PreToolUse emitters
+/// in [`command`] and [`file`] build their `decision`-bearing shape inline
+/// and never route through this function; whether their `additionalContext`
+/// is delivered is tracked separately.
 pub(super) fn emit_hook_context(hook_event: &str, context: &str) {
-    let payload = match hook_event {
-        "PreToolUse" => {
-            serde_json::json!({ "additionalContext": context })
+    let payload = serde_json::json!({
+        "hookSpecificOutput": {
+            "hookEventName": hook_event,
+            "additionalContext": context,
         }
-        _ => serde_json::json!({
-            "hookSpecificOutput": {
-                "hookEventName": hook_event,
-                "additionalContext": context,
-            }
-        }),
-    };
+    });
     println!("{payload}");
 }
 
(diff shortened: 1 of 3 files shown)


aaronsb:
## Review Summary

**What this changes**: Moves `ways scan state` SessionStart output from the bare top-level `{"additionalContext": ...}` envelope (which transcript evidence shows the harness silently drops) to the canonical `hookSpecificOutput` branch of `emit_hook_context`, updates the `--hook-event` doc comment, and flips the scenario 10 envelope guard. The core fix is correct and the blast radius is small. Ran scenarios 1, 10, 11 locally: all pass.

### Check 1 — Other consumers of `scan state` stdout: clean

The only call site is `hooks/ways/check-state.sh:23`, which is a pure passthrough — stdout goes straight to the harness, no jq on the output. No script, test, or tool anywhere in `hooks/`, `tools/`, or `tests/` parses `.additionalContext` at top level. `settings.json` also wires `check-state.sh` to **UserPromptSubmit** (settings.json:253) in addition to the four SessionStart sources, but that path already took the canonical branch pre-change (`hook_event_name` from the hook input), so it is unaffected.

### Check 2 — Other assertions of the old shape: two stale comments

- **Minor** — `tools/ways-cli/src/cmd/scan/mod.rs:79-82`: the `prompt()` doc comment still says "`SessionStart` and `PreToolUse` are the only events that take the legacy top-level `additionalContext` envelope." Now contradicts the change this PR makes 950 lines below it.
- **Minor** — `tools/ways-cli/src/main.rs:816-818`: the misroute stderr trace still says "If the invoking hook is not SessionStart/PreToolUse, the output envelope shape will be wrong and the agent will silently receive nothing." Post-change the fallback emits canonical, which is right for every event *except* PreToolUse — the message's exception list is now inverted.
- Docs/ADRs are clean: nothing documents the old bare SessionStart shape. `hooks-and-ways.md:46` documents canonical for SubagentStart; ADR-153 / the introspection design note describe harness-*recorded* transcript entries, not hook stdout, and no introspection code parses hook envelopes yet.

### Check 3 — Scenario 10 vs 11 consistency: consistent, one coverage nit

Scenario 11 asserts only the stderr trace plus a content substring; the substring survives inside the canonical JSON, so it passes unchanged. A misrouted event now emits canonical-with-`hookEventName: "SessionStart"` instead of the bare shape — since the bare shape was proven dropped, this is no worse, and the stderr trace still surfaces the misroute. **Nit**: `scan_state()` calls `scan_state_full(None)` (session_sim.rs:192-193), so scenario 10's envelope guard actually exercises the *fallback* path — identical match arm, but no test invokes `--hook-event=SessionStart` explicitly, which is what `check-state.sh` actually sends. Scenarios 10 and 11 therefore test the same invocation shape. Worth an explicit-flag variant eventually.

### Check 4 — Compact/resume/clear double-injection: one Moderate risk to verify

Per-source re-fire is bounded: `clear-markers.sh` wipes the session state dir on startup/compact/clear, so core + session-start ways re-inject exactly once per source event; resume skips clearing (marker persists, or a fresh session id gets a fresh single injection).

**Moderate** — the stale-core safety net (`tools/ways-cli/src/cmd/scan/state.rs:38-56`) re-injects core when transcript-size-since-last-summary < 5000 bytes AND marker age > 30s. Those thresholds were tuned in a world where the SessionStart delivery silently failed — the net was, in effect, the only working delivery path. Now SessionStart actually delivers ~19KB. Failure scenario: first `UserPromptSubmit` more than 30s after startup or compact, before the transcript grows past 5000 bytes post-summary — the net clears the marker and injects core a second time, ~38KB duplicated in the window per startup/compact. Whether this fires depends on whether the harness records the SessionStart `hookSpecificOutput` content into the transcript *before* the first prompt's hook runs (which would push size past 5KB and suppress it). That is empirically checkable in one live transcript. If the attachment is not recorded early, gate the net on evidence of delivery (e.g., a core sentinel string in the transcript) rather than a byte heuristic. Not blocking — the failure mode is duplication, not loss — but verify before the next release.

### Check 5 — emit_hook_context comment accuracy: one inaccuracy

**Minor/Moderate** — mod.rs:1051-1053 says "`PreToolUse` keeps its `decision`-bearing shape for now," but the retained arm (mod.rs:1057-1059) emits `{"additionalContext": ...}` with **no** `decision` field. The actual decision-bearing emissions are the inline `println!`s in `scan::command` (mod.rs:591-600) and `scan::file` (mod.rs:670-678), which never route through `emit_hook_context`. The retained arm is unreachable from production wiring (callers: `scan_prompt_surface` with hardcoded UserPromptSubmit/PostToolUse, and `scan::state` whose event comes from `check-state.sh`, wired only to SessionStart + UserPromptSubmit) — and it preserves exactly the bare shape this PR just established the harness drops. Suggest either fixing the comment to say the real PreToolUse envelopes are emitted inline in `command`/`file`, or removing the dead arm so a future PreToolUse wiring can't silently pick up a known-dropped shape. The rest of the comment's claims check out: check-post.sh:85-90, inject-subagent.sh:161-167, check-queued.sh, and `scan_prompt_surface` all use the canonical wrapper.

### Assessment

Core change is sound and the harness-drop evidence justifies the flip. Recommend fixing the three stale-comment items (mod.rs:79-82, main.rs:816-818, the `decision`-bearing claim at mod.rs:1051) in this PR since doc coherence is its stated goal, and verifying the check-4 double-injection timing before or shortly after merge.

aaronsb:
Review findings closed in dac41fe: the unreachable PreToolUse arm is deleted (emit_hook_context is now unconditionally canonical — the real decision-bearing PreToolUse envelopes live inline in scan::command/scan::file and never routed through it), the three stale comments are corrected, and scenario 10 now exercises the explicit `--hook-event=SessionStart` path that check-state.sh actually sends.

The moderate finding (stale-core safety net at state.rs:38-56 double-injecting ~19KB now that delivery works) is deliberately left untuned: whether it fires depends on whether the harness records the SessionStart attachment into the transcript before the first prompt's hook runs. One live transcript after this merges answers that empirically; tuning the thresholds before knowing would be guessing. Follow-up filed against the first post-merge session.

Issue #438, opened 2026-08-04
adr-tool: bound the active ADR set — `adr archive`, active-set semantics, and read the supersession frontmatter that already exists


aaronsb:
## Problem

`adr-tool`'s number-range domains organise an ADR set. Nothing bounds one.

This repo's own numbers, today:

| | |
|---|---|
| ADRs on disk | **85** |
| ADRs in the generated `INDEX.md` | **85** |
| already `Superseded` / `Deprecated` / `Rejected` | **10** |
| mention supersession in prose | **27** |
| declare `superseded_by:` in frontmatter | 6 |
| declare a non-empty `supersedes:` | 3 |
| …of those 9 declarations, read by any code | **0** |

Three separate things fall out of that.

**1. There is no active set.** Ten ADRs are formally not in force and still occupy the main tree, the listing, and every line of the generated index. A reader building a mental model has no way to ask for "the decisions that are in force" — the tool's answer to `list` and to `index` is always "all of them." That number only goes up.

**2. The supersession frontmatter is a producer with no consumer.** `superseded_by: ADR-167` is already being written. `ADRInfo` doesn't parse it, `lint` doesn't validate that the target resolves, `list` and `index` don't surface it. So the convention exists, decays silently, and a typo'd or dangling reference is invisible.

**3. Supersession is almost always partial, and `status` is a whole-document binary.** Of the 27 ADRs whose prose discusses supersession, most are saying some version of *"§4 of this was replaced; the rest stands."* There is no honest status for that. Marking the document `Superseded` discards 90% that is still live; leaving it `Accepted` hides that a section is dead. Today the prose carries it and nothing indexes it.

## Proposal

Three parts. **A** is the core and stands alone. **B** is small and makes **A**'s decisions defensible. **C** is a convention question worth deciding before the set grows further.

---

### A. `adr archive` and active-set semantics

```
adr archive <number> --reason "..." [--superseded-by ADR-N[,ADR-M]] [--status S] [--dry-run]
```

Moves `docs/architecture/<domain>/ADR-NNN-*.md` → `docs/architecture/archive/<domain>/ADR-NNN-*.md`.

**Archiving is not deletion, and that is the whole point.** In a set where 27 documents reference each other's supersession, deleting breaks the web the archive exists to untangle. The document stays tracked, linted and linkable; every existing cross-reference still resolves. It stops being part of the set a reader has to hold in their head.

Three behaviours, each load-bearing:

[...]

PR #441, opened 2026-08-06
feat(tools): version-stamped vendored tools with direction-aware drift detection (ADR-177)

aaronsb:
## What

ADR-177 (Draft, accept-on-merge) plus its implementation in one change:

- **`TOOL_VERSION` stamp + `--version`** in all three vendorable tools: `adr-tool`, `doc-tool`, `chart-tool` (each at 1.0.0).
- **Direction-aware macro** — the adr way's macro previously ran a bare `diff -q` and answered every difference with *"expected for customized setups"*, which reassures exactly when the copy is stale. It now greps `TOOL_VERSION` from both sides and compares with `sort -V`:

| Local copy | Disclosure |
|---|---|
| no marker | predates versioning — re-vendor |
| older | stale — re-vendor |
| equal, bytes differ (version line excluded) | customized — expected |
| newer / install unversioned | agent-ways install is stale — `ways-update` |

- **adr skill**: new *Updating a vendored copy* section (plain re-copy when unmodified; diff-and-carry-forward when customized).

## Why now

Issue #438 (`adr archive` + supersession parsing) is about to grow the tool. Landing the stamp first means post-#438 vendored copies are born versioned, and "no marker = predates versioning" cleanly bounds the unversioned era.

## Validation

- All six macro states exercised in a sandbox (fake `$HOME` + project dir): unversioned-local, older, newer, equal+customized, identical (silent), install-unversioned — each discloses the right message.
- `py_compile` clean on all three tools; `--version` verified on each.
- `adr lint` clean on ADR-177; index regenerated.

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -97,6 +97,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-174](./system/ADR-174-progressive-core-decoration-guidance-and-the-core-re-disclosure-gap.md) | Progressive core — decoration guidance and the core re-disclosure gap | Draft |
 | [ADR-175](./system/ADR-175-standing-delegation-authorization-satisfies-the-harness-permission-gate-rather-than-overriding-it.md) | Standing delegation authorization satisfies the harness permission gate rather than overriding it | Draft |
 | [ADR-176](./system/ADR-176-contract-identification-as-the-develop-loop-front-gate.md) | Contract identification as the develop-loop front gate | Accepted |
+| [ADR-177](./system/ADR-177-version-stamped-vendored-tools-with-direction-aware-drift-detection.md) | Version-stamped vendored tools with direction-aware drift detection | Draft |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-177-version-stamped-vendored-tools-with-direction-aware-drift-detection.md
+++ b/docs/architecture/system/ADR-177-version-stamped-vendored-tools-with-direction-aware-drift-detection.md
@@ -1,0 +1,104 @@
+---
+status: Draft
+date: 2026-08-06
+deciders:
+  - aaronsb
+  - claude
+related:
+  - 138
+---
+
+# ADR-177: Version-stamped vendored tools with direction-aware drift detection
+
+## Context
+
+Three tools ship inside ways and get **vendored** — copied, never symlinked —
+into consuming repositories: `adr-tool` (documentation/adr), `doc-tool`
+(documentation/linting), and `chart-tool` (softwaredev/visualization/charts).
+ADR-138 put the vendoring procedure in skills; the copies it produces are
+snapshots that never hear about upstream changes.
+
+None of the three carries a version marker. The one drift check that exists —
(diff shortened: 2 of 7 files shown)


aaronsb:
## Review Summary

**What this changes**: Stamps the three vendored tools with `TOOL_VERSION`/`--version`, rewrites the adr macro's drift check to be direction-aware (grep + `sort -V`), documents re-vendoring in the adr skill, adds ADR-177.

**Assessment**: Nothing blocking. The six macro states are logically sound and the fallthrough ordering is correct (both-unversioned lands in the old byte-diff note, equal+identical stays silent). Findings, roughly by weight:

1. **Version string is echoed unvalidated into hook-injected context** — `macro.sh:46-47` captures `[^"]+`, i.e. arbitrary text up to a quote from an untrusted project file, then interpolates it verbatim at lines 50/53/58/61. A project copy stamped `TOOL_VERSION = "0.0.0 — <instructions>"` rides straight into the disclosure. Cheap hardening: tighten the capture to a version shape (`\K[0-9]+(\.[0-9]+)*[0-9A-Za-z.+-]*`) or post-validate (`[[ "$local_ver" =~ ^[0-9A-Za-z.+-]+$ ]] || local_ver=""`), treating non-matches as unversioned. This also keeps `sort -V` semantics sane for junk input.

2. **GNU grep dependency is consistent but degrades silently** — `grep -oP` at `macro.sh:46-47` is PCRE/GNU-only. Precedent exists (`hooks/pre-commit:79`), so this is an established repo assumption, not a new one. On BSD grep both extractions fail, `2>/dev/null || true` swallows it, both vars come back empty, and the logic falls through to the old byte-diff branch — a benign, pre-ADR-177 fallback. `sort -V` (line 55, also GNU-leaning) is unreachable in that case, so its portability is moot. Fine as-is; just be aware the degradation is invisible.

3. **E402 / placement inconsistency in adr-tool** — the stamp is inserted mid-import-block (`adr-tool:24-28`, between `import sys` and `from datetime import date`), so the subsequent imports now trip pycodestyle E402 if linted — and `doc-tool` carries a `# noqa: E402`, so that convention is live in this codebase. doc-tool and chart-tool both place the stamp after imports. Suggest moving it below the import block for consistency.

4. **argparse interactions: clean** — neither adr-tool nor doc-tool subparsers define `--version`/`-v`, so no conflict; the top-level flag only intercepts before a subcommand token (`adr-tool list --version` errors as unrecognized, which is correct). chart-tool's manual `"--version" in sys.argv[1:]` matches anywhere in argv, but the tool takes no other CLI args (stdin JSON only), so nothing collides; placing it before the `isatty` gate is right.

5. **Filename/quoting injection: none** — `ADR_SCRIPT` comes from a fixed candidate list, every path expansion is quoted, and process substitution is safe under the existing bash shebang.

6. **Doc nits** — ADR-177 says "Four states" but the macro discloses five (it adds "installed template unversioned", `macro.sh:52-53`); worth reconciling before the ADR leaves Draft. And `SKILL.md:115` writes `/ways-update` while `macro.sh:53,61` and `hooks/ways/meta/skills/skills.md` use `ways-update` — pick one form.

**Considerations**: The sandbox exercise covered state selection; the residual risks above are exactly the untested surface (hostile version strings, BSD userlands, lint). Item 1 is the only one I'd fix before merge — it's a one-line pattern change.

---

PR #442, opened 2026-08-06
feat(adr-tool): adr archive, active-set semantics, and supersession reading (ADR-303)

### LINKED ISSUE #438: adr-tool: bound the active ADR set — `adr archive`, active-set semantics, and read the supersession frontmatter that already exists

aaronsb:
Closes #438.

## What

ADR-303 (Draft, accept-on-merge) adopts issue #438's parts A and B as specified and resolves part C as the issue's preferred **option 2** — section references in `superseded_by`, no new status.

**A — `adr archive` + active set.** `git mv` to `docs/architecture/archive/<domain>/` (plain rename outside a work tree), status rewrite (default `Superseded`, validated against `adr.yaml`), mandatory `--reason` banner after the H1. `list`/`index` narrow to the active set (`--archived`/`--all`; index gets a collapsed archived section); `lint` deliberately still covers the archive.

**B — the frontmatter gets a consumer.** `supersedes`/`superseded_by` parsed into `ADRInfo`; dangling references are lint errors, non-reciprocal pairs warn; `list`/`index` render the relationship ("partially superseded by ADR-102 §4").

**C — partial supersession.** `superseded_by: ["ADR-167#4"]` + in-force status = partially superseded; `archive` refuses those — the archive is for documents a reader no longer opens.

## What the new lint immediately caught (fixed here)

1 error + 7 warnings on our own corpus, exactly the decay #438 predicted: a dangling `supersedes: ADR-107 Draft (2026-03-20)` pseudo-reference, six one-directional links, and two unread field variants in the wild (`superseded_in_part_by`, `supersedes_in_part`) now folded into the read fields. `Rejected` added to statuses (config + template) — ADR-127 already used it.

## Tests

`tests/adr-archive-test.sh` — 34 checks in a throwaway git repo (never the host), wired into `run-all.sh`: round trip, staged-R + post-commit `--follow` history (checked separately per the issue's gotcha), banner honesty + byte-identical body, all five refusals + the partial-supersession refusal, index exclusion at both old and new paths, archived-still-linted, dry-run inertness. Existing lint suite: 12/12. `adr lint --check` clean on the full corpus. `TOOL_VERSION` 1.0.0 → 1.1.0 per ADR-177 — vendored copies will now correctly read as stale.

## Notes for the issue author

The spec transferred cleanly; both trip-wires you called out (staged-rename `--follow`, defensive read of the moved file) are asserted in the suite.


aaronsb:
## Review — feat(adr-tool): archive, active-set semantics, supersession reading

Reviewed against issue #438. Tests pass, but the focus here is what the suite doesn't exercise. All findings below were **reproduced empirically** in a throwaway repo, not just read from the diff. Three are blocking.

### Blocking

**1. `adr new` reissues archived numbers — duplicate ADRs**
`hooks/ways/documentation/adr/adr-tool:481` — `cmd_new` calls `get_all_adrs()`, which now defaults to the active set. Archiving an ADR frees its number, and the first-free-in-range allocation loop (lines 490–494) hands it out again. Reproduced: archived ADR-101, ran `adr new system ...` → created a second active `ADR-101` alongside `archive/system/ADR-101-*.md`. Duplicates then make `find_by_ref` ambiguous, break supersession resolution, and make lint's `by_number` order-dependent. Fix: `get_all_adrs(include_archived=True)` in `cmd_new` — number allocation must see the full corpus. This is the one place the active-set default is actively wrong.

**2. Banner injected inside frontmatter when frontmatter contains a YAML comment**
`adr-tool:1019` — the H1 scan `next(i for i, line in enumerate(lines) if line.startswith('# '))` starts at line 0 and matches YAML comments (`# reviewed by ops`) in the frontmatter. Reproduced: archive exits 0 and writes the ARCHIVED banner between frontmatter keys, corrupting the YAML — the file then fails to parse (status Unknown) and the tool reports success. Fix: start the scan at `fm_end + 1` when frontmatter was found.

**3. `is_archived` matches any ancestor directory named `archive`**
`adr-tool:302` — `'archive' in path.parts` runs against absolute paths (`get_project_root()` returns the git toplevel). A repo checked out under any path segment named `archive` (e.g. `~/archive/myrepo/`) reports **0 active ADRs**: `list` is empty, `index` collapses everything, and via finding 1, `cmd_new` restarts numbering from the bottom of the range. Reproduced. For a vendored tool (ADR-177) shipped into arbitrary repos this is a realistic environment, not a contrived one. Same check also collides with a legitimate domain `folder: archive` in `adr.yaml`. Fix: test relative to the architecture root — `path.relative_to(arch_root).parts[0] == 'archive'` (pass or derive `arch_root`).

### Should fix

**4. Missing `status:` in frontmatter → archived file has no status at all**
`adr-tool:991–1013` — if no `status:` line exists, `status_idx` stays `None` and no status is ever written; the tool exits 0. Reproduced. The docstring's own rationale ("a document in an archive still claiming Accepted is a contradiction the tool would be creating itself") applies equally to *no* status. Fix: append `status: <archive_status>` to `new_fm` when `status_idx is None` after the loop.

### Minor

**5. CRLF files are wholly rewritten to LF** — `dest.read_text()` (adr-tool:975) applies universal-newline translation, so archiving a CRLF file converts every line ending. Reproduced (`file` confirmed CRLF in, pure LF out). Cosmetically this makes the banner's "Nothing below this line has been edited" false at byte level and produces a whole-file diff on top of the rename. `read_text(newline='')`-style handling or a doc note would do.

**6. Lint `by_number` keyed on title number only** — `adr-tool:685`. `setdefault` silently keeps the first (path-sorted) ADR on collision — nondeterministic reciprocity results once duplicates exist (see finding 1). And because the map ignores filename numbers, a valid ref to an ADR whose H1 is malformed reports the misleading `resolves to no known ADR` instead of pointing at the target's own problem.

**7. `norm_ref` uppercases section refs** — `adr-tool:326`. `.upper()` runs before the `#` partition, so `ADR-167#decision-anchor` becomes `#DECISION-ANCHOR` in both the banner and the frontmatter it writes. Fine for the numeric `#4` convention; silently mangles anchor-style refs. Uppercase only the number half.

**8. Rename fallback can silently overwrite** — `adr-tool:969`. If `git mv` fails and `dest` already exists, `Path.rename` clobbers it on POSIX. A `dest.exists()` guard is cheap. Also, the "(plain rename — not a git work tree)" message (line 1041) is wrong when `git mv` failed for another reason, e.g. an untracked file inside a work tree.

**9. `list --all --archived`** — `--all` silently wins (adr-tool:362–364). Either error on the combination or document precedence.

**10. `cmd_rename` can no longer see archived ADRs** (adr-tool:588) and says "ADR not found" — arguably correct behavior, misleading message. Consider "ADR-N is archived" via an `include_archived=True` lookup for the error path.

### What's solid

- Inline `superseded_by: ["ADR-102#4"]` frontmatter form is handled correctly through the rewrite (verified — rebuilt as a quoted block list).
- Archived index links (`relative_to(arch_root)` → `./archive/<domain>/...`) resolve correctly from INDEX.md; `status_cell` annotations read well.
- Lint deliberately covering the archive is honored, including the explicit-paths branch resolving refs against the full corpus.
- The test suite's throwaway-repo isolation and the `git log --follow` assertion are exactly the right tests to have.

**Assessment**: request changes for findings 1–3 (1 is a guaranteed corpus-corruption path, 2 and 3 are silent-success corruption/misbehavior). 4 belongs in the same pass. The rest can trail.

---

aaronsb:
Review findings resolved in a133bf7: blockers 1–3 fixed (archived numbers reserved in `adr new`; banner scan starts below the frontmatter; `is_archived` reads only segments under `docs/architecture`, with `archive/` documented as reserved), 4 fixed (missing status appended), minors 6–10 fixed. Finding 5 (CRLF normalization on archive) accepted as a known minor — universal-newline reads make byte-exact CRLF preservation cost more than the edge is worth. Seven regression checks added; suite is 41/41, corpus lint clean.

Issue #433, opened 2026-07-31
meta/subagents ranks fifth on its own subject — vocabulary pass needed


aaronsb:
## Problem

`meta/subagents` surfaces weakly on its own subject. Measured during the ADR-175 work:
on *"delegate the code review to a subagent"* — a query naming delegation outright — it
ranked **fifth at 0.3**, below a project-local way.

That is the moment the way is supposed to own. ADR-175 worked around it by putting the
operative authorization inline at each delegation site (Tier 2), which does not depend on
matching. But Tier 1 still holds the doctrine and the payoff guidance, and it only earns
its place if it discloses when delegation is actually named.

## Why vocabulary rather than pattern

ADR-155 §5 routes suggestive common words to `vocabulary:`, which shapes the embedding
coordinate, rather than to `pattern:`. `review` is in the linter's `COMMON_WORDS` and is
flagged regardless of anchoring, so pattern-widening is not the lane. The current
vocabulary is:

```
subagent delegate spawn background task parallel worker teammate planner
```

Candidates worth testing: `review`, `reviewer`, `fan out`, `explore`, `dispatch`,
`in parallel`, `independent`, `code review`.

## Verification this needs

Not a blind edit — it changes matching:

- re-embed the corpus
- score before/after with `ways match` on a query set that includes the moment-of-use
  phrasings and the control queries the way should *not* win
- confirm no regression in fire breadth (ADR-158 ship gate)

## Caveat on the existing measurement

The scores above came from single-vector matching — the queries were too short to chunk,
and the tool notes that view no longer reflects the fire path. Directional, not ground
truth. Re-measure properly as part of this work.

Follow-up to #432 / ADR-175.

PR #456, opened 2026-09-08
docs(adr): ADR-180 GitHub issues as shared truth for the session task list

aaronsb:
## Summary
- Draft ADR-180: GitHub issues carrying a `tasklist` label are the shared truth for Claude Code's session task list; each session store is a cache.
- Bridge is hooks, one `gh-tasks` executable, an `issues` skill, and a way. No attend, no daemon. Field ownership replaces merge.
- Records three in-session findings: the task store accepts external writes, subagents inherit the parent session id, the id allocator reconciles from disk.

## Test plan
- [x] `docs/scripts/adr lint --check` passes
- [x] `docs/scripts/adr index -y` regenerated
- [ ] Reviewer reads the ownership split and the hook wiring for holes

### DIFF
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -96,6 +96,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-177](./system/ADR-177-version-stamped-vendored-tools-with-direction-aware-drift-detection.md) | Version-stamped vendored tools with direction-aware drift detection | Accepted |
 | [ADR-178](./system/ADR-178-register-transfers-by-demonstration-core-md-carries-policy.md) | Register transfers by demonstration - core.md carries policy | Accepted |
 | [ADR-179](./system/ADR-179-remove-the-pre-1-0-in-place-migrator-keep-the-guards-and-the-transition-fallbacks.md) | Remove the pre-1.0 in-place migrator; keep the guards and the transition fallbacks | Accepted |
+| [ADR-180](./system/ADR-180-github-issues-as-the-shared-truth-for-the-session-task-list.md) | GitHub issues as the shared truth for the session task list | Draft |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/INDEX.md
+++ b/docs/architecture/INDEX.md
@@ -96,7 +96,7 @@ _Ways architecture, matching, macros, hooks, session lifecycle_
 | [ADR-177](./system/ADR-177-version-stamped-vendored-tools-with-direction-aware-drift-detection.md) | Version-stamped vendored tools with direction-aware drift detection | Accepted |
 | [ADR-178](./system/ADR-178-register-transfers-by-demonstration-core-md-carries-policy.md) | Register transfers by demonstration - core.md carries policy | Accepted |
 | [ADR-179](./system/ADR-179-remove-the-pre-1-0-in-place-migrator-keep-the-guards-and-the-transition-fallbacks.md) | Remove the pre-1.0 in-place migrator; keep the guards and the transition fallbacks | Accepted |
-| [ADR-180](./system/ADR-180-github-issues-as-the-shared-truth-for-the-session-task-list.md) | GitHub issues as the shared truth for the session task list | Draft |
+| [ADR-180](./system/ADR-180-github-issues-as-the-shared-truth-for-the-session-task-list.md) | GitHub issues as the shared truth for the session task list | Accepted |
 
 ## Governance
 _Provenance, traceability, controls, compliance mapping_
--- a/docs/architecture/system/ADR-180-github-issues-as-the-shared-truth-for-the-session-task-list.md
+++ b/docs/architecture/system/ADR-180-github-issues-as-the-shared-truth-for-the-session-task-list.md
@@ -7,6 +7,7 @@ deciders:
 related:
   - ADR-113
   - ADR-114
+  - ADR-152
   - ADR-172
 ---
 
@@ -17,15 +18,14 @@ related:
 Claude Code ships a task list: `TaskCreate`, `TaskList`, `TaskGet`,
 `TaskUpdate`, with `TaskCreated` and `TaskCompleted` hook events. The list is
 persisted to `~/.claude/tasks/session-<first 8 chars of session id>/`, one JSON
-file per task, with a `.lock` and a `.highwatermark` beside them. Resumed
-sessions keep their tasks. Each task carries `id`, `subject`, `description`,
-`activeForm`, `status`, `owner`, `blocks`, `blockedBy`, and a free-form
-`metadata` object.
+file per task, with a `.lock` and a `.highwatermark` beside them. Each task
+carries `id`, `subject`, `description`, `activeForm`, `status`, `owner`,
+`blocks`, `blockedBy`, and a free-form `metadata` object.
 
 This project tracks work in GitHub issues. Every session that picks up an issue
 re-derives its task list from the issue body by hand, and nothing carries
-completion back. Feature request anthropics/claude-code#79096 asks for exactly
-this bridge and is open with no response. No first-party integration exists:
+completion back. Feature request anthropics/claude-code#79096 asks for a
+bridge and is open with no response. No first-party integration exists:
 `claude-code-action` reads issues as prose context for a workflow run, the
 GitHub MCP server has no notion of a session task, and the Agent SDK exposes
 the task tools as observable blocks without wiring them anywhere.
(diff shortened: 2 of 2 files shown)


aaronsb:
## Review: ADR-180

Reviewed against ADR-113, ADR-114, ADR-172, the Claude Code hooks reference, and the task store on this machine. Verdict at the bottom.

The core bet is sound. Field ownership instead of three-way merge is the right call for v1, the store is a real injection point, and confining the undocumented layout to one module on the pull half is the correct risk split. Five findings below need answers before this moves off Draft. Most are specification gaps in the decision, not flaws in the bet.

---

### 1. `status` has two owners, and a reopened issue ping-pongs

**Where**: field ownership table (L70-74), plus the `pull` bullet (L91-92).

The table gives `status` to the session. The paragraph under it maps `open`/`closed` to `pending`/`completed`, which gives GitHub the same field. The `pull` rule "a task whose issue closed elsewhere is marked `completed`" patches the close direction and leaves reopen unhandled.

**Failure sequence:**

1. Session marks the task for #12 `completed`. `TaskCompleted` → `push` closes #12.
2. A maintainer reopens #12: the work was wrong, or was closed as duplicate by mistake.
3. Next `pull` runs. `status` is session-owned, so pull must not touch it. The task stays `completed`.
4. `Stop` → `push` reconciles status outward. Status is `completed`, so push closes #12 again.

The bridge re-closes an issue a human deliberately reopened, every turn, and the local list never shows the reopen. The inverse also loses data: an issue closed as `not planned` marks the task `completed`, so the model reports finished work it never did.

**Suggestion**: make the closed/open bit GitHub-owned in both directions and let `completed` be a *request* to close rather than an assertion. Concretely, add a `metadata.pushed_state` recording the last state this session pushed. Push only when local status disagrees with both the observed remote state and the last pushed state. That is a two-field snapshot, far short of the three-way merge the alternatives section rejects, and it kills the ping-pong.

Multi-session hits the same edge: session A closes #12, session B is mid-work with it `in_progress`. B's pull marks it `completed` and B stops. Or B's push labels a closed issue `in_progress`. No rule resolves that today.

---

### 2. `PostToolUse on Bash(gh issue *)` is not a matcher

**Where**: hook wiring table (L112).

Hook matchers match the tool *name* only. `Bash(gh issue *)` is permission-rule syntax. The docs put argument filtering in a separate `if` field on the individual hook entry:

```json
{
  "matcher": "Bash",
  "hooks": [
    { "type": "command", "if": "Bash(gh issue *)", "command": "gh-tasks pull" }
  ]
}
```

Every hook installed in this repo confirms the shape: `~/.claude/settings.json` uses `"matcher": "Edit|Write|Bash|Task"` and filters inside the script.

Two coverage gaps follow even after the syntax is fixed. A close performed through `gh api`, through the GitHub web UI, or by a merge commit closing keyword never produces a `Bash(gh issue *)` call, so the immediate-reflect promise only covers one of several paths. And parallel `gh issue` calls in one assistant turn fire the hook concurrently, which feeds finding 3.

---

### 3. Two id allocators, one lock whose protocol is unknowable from outside

**Where**: `pull` bullet (L91), store I/O paragraph (L100-104).

The ADR says `pull` writes "under `.lock`". The lock file on this machine is zero bytes. Its discipline is undocumented: `flock(2)`, `O_EXCL` create-and-unlink, and a pid-file convention are three mutually non-excluding protocols. `gh-tasks` picking one gives no mutual exclusion if Claude Code picked another. "Writes under `.lock`" asserts a guarantee the external writer cannot verify.

**Collision window**, which the ADR's own evidence bounds: the highwatermark advanced from 1 to 2 *after* the external `2.json` was read. Reconciliation happens on read. So:

1. `gh-tasks pull` allocates id 2, writes `2.json`, sets `.highwatermark` to 2.
2. The model calls `TaskCreate` before any `TaskList` in this session.
3. If the in-process counter is memoized rather than re-read, `TaskCreate` writes `2.json` and the pulled issue task is gone.

The hook wiring widens this rather than narrowing it, because `PostToolUse` pull can fire mid-turn while the model is creating tasks.

**Suggestion**: write `.highwatermark` before the task file, never after, so a racing allocator sees the reservation. Better, allocate externally-created tasks from a disjoint high block and never touch the counter. State explicitly that ordering is best-effort against an allocator whose caching behavior is unobserved, and record which lock protocol was probed.

---

### 4. `pull` is an untrusted-input ingestion path and the ADR does not treat it as one

**Where**: opt-in section (L130-134), Consequences.

Issue bodies become task `description`, which reaches the model through `TaskList` with no provenance marker. On a public repository an outside contributor files an issue, a maintainer labels it `tasklist`, and the author then edits their own body to carry instructions addressed to the agent. The label gate is an authorization boundary held by whoever can apply labels; the ADR treats it purely as an opt-in convenience.

`SessionStart` whispering the full open list widens the surface, since titles reach context with no label at all.

**Suggestion**: name the label as a trust boundary in the Decision, and have the store module fence pulled body text as untrusted content when it lands in `description`. Given ADR-152's deny baseline, silence on this reads as an oversight rather than a judgment.

---

### 5. `CLAUDE_CODE_SESSION_ID` is not a documented surface

**Where**: L49-51 ("documented or stable surfaces"), L100-101.

The hooks reference does not list it among the variables Claude Code sets. It lists `CLAUDE_PROJECT_DIR`, `CLAUDE_PLUGIN_ROOT`, `CLAUDE_PLUGIN_DATA`, `CLAUDE_EFFORT`, and states that hooks otherwise inherit the parent environment. The documented path for a hook is `session_id` in the stdin JSON, and all eleven hooks installed in this repo use exactly that.

The empirical half holds: a subagent Bash shell in this session has `CLAUDE_CODE_SESSION_ID` set to the parent id with `CLAUDE_CODE_CHILD_SESSION=1`, and `~/.claude/tasks/session-<id8>/` matches. So the variable works for skills and subagent shells, where no stdin JSON exists.

Three gaps remain:

- Hooks should read stdin `session_id` and treat the env var as fallback, reversing the ADR's stated order.
- Resume was not among the three verified facts. If a `--resume` session gets a fresh id, the store rebinds and the ADR's "resumed sessions keep their tasks" premise breaks.
- Context point 1 says agent teams share a directory "under a team name"; the Decision resolves the path from the session id alone. Those cannot both be the path rule. This machine shows only `session-<id8>` directories, including for a team session, so the team-name form needs re-verification or removal.

ADR-172 already solved the shape of this. Decision 4 refuses to write consumption records under an unresolved identity and degrades to the other conduit. `gh-tasks` should refuse to write when it cannot resolve a store path with confidence, rather than guessing.

---

### Smaller findings

**`blockedBy` loses local edges.** GitHub owns `blockedBy`, so pull overwrites it. A task blocked by a *local* task with no issue has that edge silently erased on the next pull, and the model starts blocked work. The id spaces also differ: GitHub gives issue numbers, tasks carry task ids, and a blocking issue that lacks the label is never pulled, so the translation has no target. Neither the translation nor the local-edge case is specified.

**Pull silently reverts description edits.** A session-side description edit posts a comment and stays in the local file until the next pull overwrites it from the body. The refinement disappears from the task list and lives only as a comment nothing reads back. Worth stating as a consequence, since the current text reads as though the comment preserves the edit.

**Whole-file writes clobber `owner` and `activeForm`.** Real task files on this machine carry `owner` (teammate name under agent teams) and `activeForm`. Neither appears in the ownership table. If pull writes whole files, a teammate assignment made in-session is erased. Specify read-modify-write over the three GitHub-owned fields.

**Fail-closed can still leave a half-written store.** Shape validation on read does not catch a *write*-side change: a new required field passes a check over the fields the reader knows, and the files it writes are then wrong. And if the check fails partway through a multi-task pull, tasks 1-5 land while 6 aborts. Validate once before writing anything, stage into a temp directory, then rename. Silence also has no observer: the ADR says it "has to be noticed" without saying by what. One `whisper` line on `SessionStart` saying the layout is unrecognized costs almost nothing and turns silence into signal.

**`TaskCompleted` fires before the completion commits.** The docs describe it as blocking, where exit 2 prevents the task from being marked completed. Pushing a close from that hook closes the issue before the local write is confirmed, and a later hook exiting 2 leaves a closed issue with an uncompleted task. `Stop` is the safer place for the close; keep `TaskCompleted` for validation only.

**The push half needs write access.** Closing, labeling, and commenting all require it, and `gh issue edit --add-label` fails outright if the label does not exist. Read-only contributors and fork workflows get a pull that works and a push that errors every turn. The opt-in section covers repositories without the label but not users without the permission.

---

### Citation check

**ADR-172 is cited for the drain pattern without its constraints.** ADR-172 is more than "run it on `Stop`". It carries three properties this design needs and does not inherit: merge/append semantics so two writers cannot clobber each other's marks (Decision 3), refusal to write under an unresolved identity (Decision 4), and a re-entry guard with a ceiling because inject-and-continue means the continued turn ends in another `Stop` (Decision 6). If the `Stop` push injects whisper output, the re-entry risk applies unchanged. Cite the constraints or say which ones do not carry over.

**"Out of reach for subagents" overstates the ADR-113 limitation.** ADR-113 gives `attend` a normal CLI, and a subagent with Bash can run `attend inbox`. What a subagent cannot receive is Monitor delivery, which is the parent's channel. That is the same distinction ADR-172 draws between the two conduits, and it is the accurate version of the rejection.

**ADR-114 as "the insistent trigger path" is loose.** ADR-114's mechanism is on-demand: `attend` emits an affordance and Claude chooses to run `ways show attend/<signal>`. Its own text says no new hook event and no automatic firing. Describing it as a path that surfaces deltas reads more like push than ADR-114 delivers.

**anthropics/claude-code#79096 checks out**: open, zero comments, titled "Resume a self-filed GitHub issue's checklist as the session's task list."

---

### Alternatives the section should have weighed

**One issue holding a checklist, rather than one issue per task.** This is what the upstream feature request asks for, per its title. It removes id allocation, removes the number-to-id translation, and makes the whole task list one API object. The per-issue model buys real things, chiefly independent close and label state, but the cheaper design deserves an explicit rejection.

**Read-only pull as v1.** Push is the half that mutates other people's issues and carries findings 1 and 4. Pull alone delivers seeding on session start, which is most of the stated value, and defers every write-side failure mode. That the ADR ships both halves at once should be a decision, not a default.

**A git-tracked intermediate file.** A `.tasks.yaml` in the repo, synced to GitHub separately, decouples the undocumented-store risk from the GitHub risk and gets history and review for free. Slower to converge; worth a line.

**Adopting the label and prefix convention now, deferring the machinery.** The Neutral section says supersession by #79096 carries the convention over. That implies a cheaper option: take the convention, wait on the executable. Not weighed.

**Sub-issues and native dependencies.** `gh` 2.100 exposes `blockedBy`, `blocking`, `parent`, and `issueType` on `gh issue view --json`, verified against issue #433 in this repo. The hedge "where GitHub exposes them" is resolved, and the fenced-yaml fallback for `blockedBy` is likely dead weight. Good that it was already deferred.

---

### Verdict

**Needs revision before merge.** The approach is sound and I would build it. Findings 1, 2, 3, and 5 are specification gaps that produce wrong state or silently fail as written, and finding 4 is a missing security consideration. All five are fixable inside the current design without abandoning field ownership.

The framing worth keeping intact: ownership over merge, one module for the undocumented layout, no daemon, and the opt-in label gate. Those are the load-bearing calls and they hold up.

---

aaronsb:
Revised per review. Every gated finding has a resolution in the ADR text:

1. **Status ping-pong**: open/closed is GitHub-owned in both directions; `completed` is a close request gated on `metadata.pushed_state`. Reopen returns the task to `pending`. Not-planned lands with `state_reason`.
2. **Matcher syntax**: `matcher: Bash` with `if: Bash(gh issue *)`. Web, API, and merge-keyword closes reach the store on the next throttled pull, stated as such.
3. **Id collision**: pulled tasks take `10000 + issue number`, never touch `.highwatermark`. Idempotent pull, no allocator race. Lock protocol recorded as unverified until implementation probes it.
4. **Untrusted input**: the label is named as a trust boundary, descriptions carry a provenance fence, whisper covers labeled issues only, ADR-152 cited.
5. **Session id**: hooks read stdin `session_id`, env var is the fallback for skills and subagent shells, refuse to write when unresolved (ADR-172 D4). Team directories out of scope for increment one. Resume flagged unverified.

Smaller: `body_sha` keeps local description edits until the body changes; read-modify-write over GitHub-owned fields only; local `blockedBy` edges preserved; staged-and-renamed writes with a whisper line on unrecognized layout; `TaskCompleted` validation-only with close on `Stop`; push degrades to pull-only without write permission. Citations corrected for ADR-113, ADR-114, ADR-172. Alternatives section gains checklist, read-only pull (adopted as increment one), git file, convention-only. Native `blockedBy` adopted, yaml fallback dropped.

aaronsb:
## Re-review of f2fc2eb: four of five resolved, one new collision

All five gated findings are answered. `pushed_state` plus `body_sha` closes the reopen ping-pong and names the two-session case explicitly. The matcher moves to `Bash` with `if: Bash(gh issue *)`, and dropping `TaskCompleted` for `Stop` on veto grounds is the right read. The trust-boundary section, the provenance fence, and the ADR-152 relation close the ingestion gap. Identity follows ADR-172 Decision 4 and scopes team directories out. The two unverified facts are now labeled as such, which is the honest form.

One new hole, introduced by the fix to finding 3.

### Deterministic ids collide with locally allocated ones

**Where**: "Identity and ids" (`10000` plus the issue number), read against "the allocator reconciles from disk on read and continues above the block."

Continuing above the block is exactly what reintroduces the collision, in the opposite direction:

1. Pull issue #100. Task id `10100`. The next `TaskList` reconciles `.highwatermark` to `10100`.
2. The session creates five local tasks. They take `10101` through `10105`.
3. A maintainer labels issue #105. The next pull wants id `10105`.
4. That file exists and belongs to a local task.

Issue numbers here are dense and past #450, and sessions routinely create tasks, so the two id spaces overlap rather than merely abut. The old text keyed writes on `metadata.github_issue`; the revision dropped that phrase in favor of read-modify-write over GitHub-owned fields, which does not say what happens when the file at the target id is not that issue's task.

**Suggestion**: one sentence, either shape. Have pull refuse to write an id whose existing file carries no matching `metadata.github_issue`, and whisper the refusal. Or reserve the block above anything the allocator will reach, since the offset is arbitrary and a per-repo base far above the local counter costs nothing. Restoring the `metadata.github_issue` key phrasing alongside would make the invariant explicit.

Nothing else in the revision needs another pass.

PR #501, opened 2026-09-16
fix(reconcile): refuse to replace a real directory or file at a projection root

forayconsulting:
## Summary

- `ways reconcile` now stops before touching anything when a projection root (`~/.claude/skills`, `agents`, `commands`, `hooks/ways`, `hooks/check-config-updates.sh`, `bin/*`) is a real directory or file rather than a symlink. It lists the paths and exits non-zero. Nothing is linked, `settings.json` is not merged, nothing is deleted.
- New `--force` renames each such path to `<name>.ways-backup-<unix-seconds>` in the same parent, then links. Still no delete.
- `--dry-run` reports `refused <root> (...)` per real path and exits 0, so the preview shows the refusal.
- Wrong or dangling symlinks are still replaced without `--force` (unchanged).
- `ways update` (`update.rs:322`, `:443`) and the installer pass no `--force`; a refusal fails them loudly with the same message. The installer's fresh path now prints a recovery card instead of `Done.` over a failed step.
- Docs and installer strings that promised "your files are untouched" now say what happens at a projected root. `docs/reference/ways-cli.md` gains a `ways reconcile` entry; there was none.

## Why

`reconcile_symlink` treated anything at a root as replaceable and called `std::fs::remove_dir_all` (`reconcile.rs:159-163`, `186-194`). The comment justified it as "a real dir/file from a prior copy-mode projection", but a user's own `~/.claude/skills` (Claude Code's personal skills location, which README.md:222 points people at) looks the same. A first install, or any `ways update`, deleted it silently. The three existing reconcile tests never put a real directory at a root.

install-guide.md:18 and :24, README.md:101, install.sh:63 and :381, and deployment.md:11 and :18 all promise preservation, and ADR-142 section 7 records that the only mutation to a user-owned file is the `settings.json` merge. This restores that premise, so no new ADR.

How it was found: installing 1.12.0 on a machine with about twenty hand-built skills in `~/.claude/skills`, one agent, and two commands. Reading the reconciler first was the only reason they survived.

## Trade-off (called out for review)

[...]

PR #502, opened 2026-09-16
fix(settings-merge): seed the first-apply base from our hooks only

forayconsulting:
## Summary

- On a first apply (no last-applied base), `apply_to_files` now seeds the base from the live hooks that are structurally ours (`entry_is_ours`), instead of from every live hook.
- Result: a fresh 1.0 install into an existing `~/.claude` keeps the user's own hooks. A stale agent-ways hook from an older layout is still claimed and replaced, not duplicated (the case the seed was added for).
- New `seed_ours_from_live` helper; two tests. No behavior change once a base exists.

## Why

Root cause: `settings_merge.rs:451-468` seeds `base.hooks` from all of `live.hooks` when `base_path` is missing. The merge then filters `theirs` by `!base_entries.contains(e)`, so every pre-existing user hook is treated as something we wrote last time and dropped. The self-audit passes because the base agrees. The seed predates the structural ownership predicate from #247, which is exactly the tool that separates the two cases.

Found in the sandbox run for #501: a `settings.json` with `{"hooks":{"UserPromptSubmit":[...echo user-hook...]}}` came out of a first reconcile without it. The real-world shape that prompted the check: an operator with three hand-written hooks (UserPromptSubmit, SessionStart, Stop) in `~/.claude/settings.json` installing agent-ways for the first time.

## Trade-off (called out for review)

A user hook that lives under `.claude/hooks/` or `.claude/bin/` is still classified as ours. That is the documented trade-off at `settings_merge.rs:416-418` and is unchanged here.

## Test plan

```
cargo test --manifest-path tools/Cargo.toml -p ways -- --test-threads=1
  main:      260 passed, 1 ignored (+7 +11 integration)
  this PR:   262 passed, 1 ignored (+7 +11 integration)
```
- `first_apply_seed_claims_only_our_hooks_and_keeps_the_users` (file-level, `ways-setmerge-*` harness): three user hooks across three events plus one stale ours; asserts each user entry survives by identity in its event, the stale ours is replaced exactly once, and the persisted base holds only `SessionStart` with only our entry. Seen red with the seed line reverted, green with it.
- `seed_from_live_takes_only_structurally_ours`: user-only events are not seeded; a user hook that merely passes a projection path as an argument is not claimed.
- `migrating_existing_install_seeds_base_and_passes_self_audit` (existing) still passes: the stale-ours replacement path is intact.

```
cargo clippy ... -D warnings     only the pre-existing memory_seed.rs:272 error on main; nothing new
bash scripts/check-portability.sh   passed
```

Sandbox: fresh reconcile into a `~/.claude` whose `settings.json` had three user hooks (UserPromptSubmit, SessionStart with `timeout`, Stop). After the merge: all three present unchanged, four ways `SessionStart` entries added, `model` kept, second apply reports `settings.json already up to date`.

## Not in scope

- The reconciler's handling of real directories at projection roots (#501).
- Ownership classification of user hooks placed under `.claude/hooks/`.


aaronsb:
Reviewed and merging. The root cause reads as you describe: with no base, every live hook was seeded as ours and the merge filter dropped the user's entries while the self-audit agreed with itself.

Two observations for a follow-up, neither blocks:

- The seed is now redundant with the structural backstop: the merge filter and `stripped_user_view` both already exclude everything `entry_is_ours` claims. Seeding an empty map passes every test, including the new file-level one. Keep it as defense in depth and say so in the comment, or seed empty with a pointer at `entry_is_ours`. Either way the comment should credit the backstop for the separation.
- One historical gap: the SessionStart hooks shipped between 2025-12-12 and 2026-01-03 used `cat` as the executable and are not claimed by the predicate, so a pre-1.0 install last synced in that window keeps the stale entry beside ours. The cost is core.md injected twice, not data loss. A sentence in the comment is enough.

Verified locally: 262 tests, portability and register checks pass on the branch.

PR #505, opened 2026-09-17
feat(attend): keepwarm sensor keeps the prompt cache warm (ADR-182)

aaronsb:
## Summary

- A `keepwarm` sensor emits one Monitor wake at 50 idle minutes while an armed window is open and the context is 50k tokens or more. The wake is a turn over the full context, so it reads the cached prefix and resets the one-hour timer. ADR-182 records the decision and the fork-versus-wake substitution from the cache-tax mod.
- The transcript is the clock and the scorecard. `ways context --json` gains `--session` and a `usage_tail`; the sensor takes its idle clock from the newest entry, scores paid cold writes (which arm three hours), and reads a wake's verdict from the first entry after it. A wake that rewrote the prefix, or produced no turn inside the margin, stops the window with the reason on record.
- `attend keepwarm on [WINDOW] | off | status` owns the per-session arm file and renders the card. `attend status` shows a keepwarm row. The agent contract, one word and no tools, is in the three synchronized attend guidance files.
- Side commit: the sub-agents way now names `/start`, `/develop`, and `/merge` as procedures carrying the delegation grant, with a section on picking the Agent model by the job.

## Test plan

- `cargo test --workspace` passes; 15 new tests in `sensor-keepwarm` cover the wake, both verdicts, cold-write scoring and auto-arm, the arm file ruling the deadline, the no-turn stop, the lapse-while-down stop, the card, and file round trips. Two new tests in `ways-cli` cover the ISO parser and the usage tail.
- Verified on this session's transcript: `ways context --json --session <id>` returns the tail; `attend sensors` lists keepwarm active; `attend keepwarm on 90m`, `status`, `off`, and a bad window all behave; `attend status` shows the row.
- Not verified live: an actual wake needs 50 idle minutes under the new binary. The first armed idle stretch after install is the proof; `attend keepwarm status` will show `last wake read Nk` on success or a stopped reason on failure.
- `scripts/check-register.sh` and `ways lint` pass on every edited way and skill; `docs/cli/attend.md` regenerated; ADR index regenerated, `adr lint` clean.


aaronsb:
## Review: keepwarm sensor (ADR-182)

**What this changes**: a new `sensor-keepwarm` crate whose pure `State::step` decides one wake per idle stretch from the transcript's usage tail, a message-lane predicate in the tick loop, per-session arm file and ledger under attend's cache dir, `attend keepwarm on|off|status`, and `ways context --json --session <id>` with a `usage_tail`.

**Verified**: `cargo test -p sensor-keepwarm -p ways` passes (15 and 262). Traced `step` against the scheduler: `min_interval == base_interval == 60s`, so `AdaptiveInterval` cannot back the poll off past a minute, and the wake lands between 50m00s and 51m00s of idle. The message lane never records engagement, so the slot's absolute refractory cannot drop the observation, and the permissive governor's 3s cooldown and 30-per-minute cap cannot hold it past the ten-minute margin. A magnitude of 3.0 maps to `medium` in `try_disclose`, which `emit_batch` writes to stdout. The cold-verdict path scores the miss first and then disarms, so the disarm wins (the test proves it). `iso_to_epoch` matches `date -u` on the test vectors and the era arithmetic is right across the 2000-03-01 boundary. The three guidance files carry the `keepwarm:` paragraph.

I also ran the scoring rule over this session's own transcript. That is where the first finding comes from.

### 1. Cold-write scoring counts every line of a multi-line response (high)

**Location**: `tools/ways-cli/src/cmd/context.rs` `read_usage_tail`; `tools/sensor-keepwarm/src/lib.rs:319-340`

**Problem**: Claude Code writes one transcript line per content block, and every line of the same API response carries the same `message.id` and the same `usage`. In this session's transcript, 41 of 54 message ids have more than one line. `step` treats each line as a request: after the first line sets `ctx = input + read + creation`, the second line's identical `cache_creation` is compared against that inflated prior and scores again.

Simulating the rule as written over this transcript scores two paid cold writes on a session that was never cold. Both come from one message (id `msg_011Cf9aq…`, `cache_read` 23805, `cache_creation` 24576, a large tool result appended to a warm 24k prefix). Deduplicated by message id, the same transcript scores zero. In production that message would have auto-armed a three-hour window and put "2 cold writes paid, $0.98" on the card.

**Suggestion**: deduplicate in `read_usage_tail` by `message.id`, keeping the last line per id. `USAGE_TAIL_LEN` then counts requests rather than lines, which is what the doc comment already claims. Add a second guard to the miss rule so a write that accompanies a read of the prefix is growth, not a miss:

```rust
// Current
if prior > MISS_FLOOR_TOKENS && entry.cache_creation >= prior / 2 {

// Suggested
if prior > MISS_FLOOR_TOKENS
    && entry.cache_creation >= prior / 2
    && entry.cache_read < prior / 2
{
```

A cold rewrite reads nothing; a big tool result reads the whole prefix. The second clause is what tells them apart.

### 2. A second wake in one idle stretch after a self-reload (medium)

**Location**: `tools/sensor-keepwarm/src/lib.rs:509-511`, `305-317`

**Problem**: `KeepwarmSensor::new` starts from `State::default()` and never reads the ledger. Attend self-reloads on a binary change, which is exactly what the install step in this PR does. Sequence: wake fires at 50m, attend reloads at 50m10s, the baseline poll takes `last_seen_at` from the transcript (still 50m old), the next poll sees `idle >= 50m` and `pinged_at == None` and wakes again while the first wake's reply is landing. The same reset drops `misses`, `last_ping`, and `stopped` from the ledger the first time `changed` is true, so `attend keepwarm status` forgets the session's cold writes after any reload. ADR item 5 promises those on the card.

**Suggestion**: seed `State` from `read_ledger` in `KeepwarmSensor::new` (`pinged_at`, `misses`, `last_ping`, `stopped`). A stale `pinged_at` is already bounded by the ten-minute no-turn stop.

### 3. Arm file races between the CLI and the sensor (medium)

**Location**: `tools/sensor-keepwarm/src/lib.rs:171-175`, `526-537`; `tools/attend/src/cmd/keepwarm.rs:34-49`

**Problem**: both writers use `std::fs::write`, which truncates and then writes. The sensor reads the arm file at the top of `poll`, shells to `ways context`, and writes at the bottom. An `attend keepwarm on` that lands in that gap is overwritten by a disarm (window ended, no-turn stop, or cold verdict); an `off` in the gap is overwritten by an auto-arm. A torn read returns `None`, which `step` treats as deadline 0 for one poll and heals on the next. None of these lose money, but the operator's verb silently loses.

**Suggestion**: write to a sibling temp file and `rename` in `write_arm`. In `poll`, re-read the arm file before writing and only write when its deadline still equals the `file_deadline` the step started from. When it differs, the operator moved it and the next poll picks that up.

### 4. A coincident real turn can read as a cold verdict (medium-low)

**Location**: `tools/sensor-keepwarm/src/lib.rs:343-374`

**Problem**: the verdict comes from the first usage line at or after `pinged_at`. If a user or peer turn starts in the same window and its first request appends a tool result of a tenth or more of the context, `cache_creation < cache_read / 10` is false and the window stops with "the cache was already gone" while the prefix was in fact read. The ADR's item 3 states the rule this way, so the text moves with the code.

**Suggestion**: judge cold by the read, not the write ratio: `cache_read < prior / 2` is cold, anything else is warm. The write ratio can stay in the log line.

### 5. `disarm` records a stop reason on an already-off session (low)

**Location**: `tools/sensor-keepwarm/src/lib.rs:282-288`, `392-398`

**Problem**: `attend keepwarm off` while a wake is outstanding sets `deadline = 0` but leaves `pinged_at`. Ten minutes later the no-turn check fires and `status` reads "stopped, the wake produced no turn within ten minutes". Guard the reason on `self.deadline > 0` at the time of the disarm.

### 6. Auto-arm below the wake floor (low)

**Location**: `tools/sensor-keepwarm/src/lib.rs:328-338`

**Problem**: a cold write on a context between 20k and 50k arms three hours and logs "keeping the cache warm for 3h00m", but the wake needs `ctx >= BIG_TOKENS` so nothing fires. Either arm only when `entry.context() >= BIG_TOKENS` or say in the log that the context is under the floor.

### 7. `window_secs` on the sensor's write (low)

**Location**: `tools/sensor-keepwarm/src/lib.rs:532-533`

`deadline.saturating_sub(now).max(window)` records the old file's 6h against a 3h auto-arm deadline. Nothing reads the field today. Drop it or write `AUTO_WARM_SECS`.

### 8. The idle clock is the end of streaming, not the request (note)

The transcript timestamp is when each content block finished streaming, so idle is under-measured by the last response's duration. The ten-minute margin absorbs a normal turn. Worth one sentence in decision 1 of the ADR.

### ADR delivery

Items 1 through 8 are delivered. Item 4 is delivered with the defect in finding 1, and item 5's "the session's cold writes" survives only until the next reload (finding 2). The lapse-while-down stop turns an operator's 6h window into a 3h auto-arm after the next paid turn; that follows from the design and the card names the reason.

### Assessment

The cut is right: a pure `step` with the file and process reads outside it, a one-line lane predicate, and tests over the decision table. Finding 1 fires on real data from this session, so I'd land it before merge; the rest can follow in the same PR or the next.

**Not checked**: a live wake (needs 50 idle minutes under the new binary), `cargo test -p attend`, the regenerated `docs/cli/attend.md`, and the sub-agents way side commit.

aaronsb:
Remediated in the follow-up commit, all seven findings:

1. `read_usage_tail` collapses lines by `message.id`, keeping the last line's timestamp. The miss rule adds `cache_read < prior / 2`. Replayed over this session's live transcript through the new binary: 32 entries, 32 distinct timestamps, 0 misses scored.
2. `KeepwarmSensor::new` seeds `State` from the ledger. At baseline, a pending wake whose turn already landed is cleared; one that has not is kept, so no second wake in the stretch.
3. Arm and ledger writes go through temp file plus rename. The sensor re-reads the arm file before writing and leaves it alone if the deadline moved during the poll.
4. Warm verdict is `cache_read >= prior / 2`; the write is not consulted. ADR-182 item 3 reworded.
5. Operator `off` clears `pinged_at` and records no stop reason.
6. Auto-arm requires the new context to be at or above 50k. The miss is still scored.
7. The sensor's arm write records `window_secs` as deadline minus now.

Five new tests cover 1, 4, 5, 6, and the restart case in 2. `cargo test -p attend` passed in the workspace run before the PR; the whole workspace is green again after these changes.

PR #504, opened 2026-09-16
feat(reconcile): project scope, wire ways into one repository (ADR-182)

forayconsulting:
Stacked on #501 and #502; the first four commits are theirs. Rebases onto `main` once those land, then marked ready.

## Summary

- **ADR-182 (Draft):** install scope as a flag on the one reconciler.
- **`ways reconcile --scope project [--project DIR]`:** links only `hooks/ways`, `hooks/check-config-updates.sh`, and `bin/*` into `~/.claude`; leaves `skills/`, `agents/`, `commands/` alone; merges the hooks block (hooks only, no `permissions`) into `<project>/.claude/settings.local.json`; keeps a per-project base at `$XDG_STATE/agent-ways/projects/<key>/settings-applied.json` with the project path recorded beside it. The user `settings.json` is never written.
- **Scope is remembered:** a bare `ways reconcile` (what `ways update` runs) infers project scope when no user-scope base exists and project bases do, and re-merges every recorded project. `--scope user` / `--scope project` override.
- `manifest::hook_roots`; `settings_merge::{Slices, merge_with, apply_to_files_with}` with the existing `merge`/`apply_to_files` as the user-scope form; `reconcile::run_with(Options)`.
- Installer: `--scope=project [--project=DIR]` passthrough on the fresh and update paths, project-scope banner.
- Docs: install-guide scenario, README install-scope paragraph, extending.md section, `ways reconcile` reference flags, development.md sandbox example. `ways init` gitignores `settings.local.json.bak`. `adr index -y` also adds the missing ADR-181 row.

## Why

An install is active in every session in every directory, because hooks are merged only into `~/.claude/settings.json`. Claude Code concatenates hooks across the user file and the project's `.claude/settings.json` / `.claude/settings.local.json` (the `.local.json` layer is project-only; ADR-169:72-79), so a hooks block in one repo's local settings fires only there. Nothing in the binary or docs tooled that.

The case that surfaced it: an operator with two Claude Code profiles switched per folder (`CLAUDE_CONFIG_DIR` via direnv), the second profile symlinking `skills/`, `agents/`, `commands/`, `settings.json` back to `~/.claude`, about twenty hand-built skills, three custom hooks, who wanted ways in one repository. What they installed by hand and what works is exactly this shape; the PR makes it a supported mode.

[...]

Issue #503, opened 2026-09-16
Honor CLAUDE_CONFIG_DIR for the projection root and hook paths (multi-profile setups)


forayconsulting:
## Setup

Two Claude Code profiles selected per folder with `CLAUDE_CONFIG_DIR` (switched by direnv). The second profile symlinks `skills/`, `agents/`, `commands/`, and `settings.json` back to `~/.claude`, so skills and hooks are shared while sessions, transcripts, credentials, and plans are kept apart. Sessions started under the second profile write transcripts under `$CLAUDE_CONFIG_DIR/projects/<slug>`, not `~/.claude/projects/<slug>`.

## What agent-ways assumes

- `tools/ways-core/src/paths.rs:129-131`: `projection_root()` is `home/.claude` with no env override. The four XDG roots next to it (`:49-73`) do honor `$XDG_*`.
- `tools/ways-cli/src/session.rs:259` and `:312-313` build `.claude/projects/<slug>` from `home_dir()`. Under a relocated profile, token position and context-window detection read the wrong transcript store (a stale personal transcript, or nothing), so re-disclosure cadence and `ways list` distances are wrong or inert. `corpus.rs:165` enumerates `.claude/projects` the same way.
- `settings.json` ships all 32 hook commands as literal `${HOME}/.claude/...`. `settings_merge.rs:383-386` `command_is_ours` matches the `.claude/hooks/` and `.claude/bin/` substrings; the doc comment at `:378-382` records `CLAUDE_CONFIG_DIR` as out of scope, guarded by `shipped_hooks_are_all_recognized_as_ours`.
- 20+ hook scripts hardcode `${HOME}/.claude` (`check-setup.sh:8,21`, `require-ways.sh:10`, `check-prompt.sh:31,50,55`, `check-file-pre.sh:21`, `check-bash-pre.sh:28`, `inject-subagent.sh:88,118,122`, ...). Only `hooks/ways/softwaredev/delivery/issues/gh-tasks:50` reads `CLAUDE_CONFIG_DIR`.
- `scripts/install.sh:20`: `DEST="${HOME}/.claude"`, no override.
- Doc drift: `hooks/ways/meta/skills/skills.md:34,46` tell skill authors to resolve `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` and cite `skills/ways-update` as the reference, but that skill hardcodes `$HOME/.claude` (`SKILL.md:118`) since fb1c4b0.

## Effect

[...]

PR #508, opened 2026-09-17
feat(ways): targets as the unit of activation, preflight plan, withdrawal (ADR-184/185)

aaronsb:
## Summary

- **ADR-184**: installation and activation are separate states. A target is a Claude Code config directory recorded under `targets:` in the user config. `ways reconcile` converges every enabled target and withdraws from every disabled one. The installer's handoff to a bootstrap is the next increment; in this one the default target is implicit and every existing install behaves as before.
- **ADR-185**: CLI output contract. `ways config show` renders a table; `--json` is the stored file; `--json --effective` the resolved state.
- **Preflight.** `ways config target plan <dir>` classifies every projection root and dry-runs the settings merge in memory: kept, added, replaced, removed. `target add` prints the plan and exits 3 when a real path sits at a root or an entry of the operator's would be removed, unless `--force`. This is the structural answer to the clobber Clayton hit in #501 and #502: nothing is written until the plan is shown.
- **Withdrawal.** `target disable` and `remove` unlink our symlinks and remove our hooks block and permissions through the same merge base that wrote them. User hooks, user permissions, and real directories are untouched. Re-enabling starts from the empty base.
- **Project switch.** `enabled: false` in a project's `.claude/ways.yaml` makes every scan lane inject nothing there.
- `ways status` opens with the install state and the targets. Docs: CLI reference, install guide, deployment way.

## Test plan

- `cargo test -p ways -p ways-core`: 278 + 162 pass. New: 4 config tests (implicit default, explicit empty, writer round-trip preserving other keys, enabled switch), 6 reconcile tests (converge then withdraw with user hooks and permissions preserved and re-enable from the empty base, withdraw leaves real paths and foreign symlinks, empty list touches nothing, plan names refused roots and kept and added entries, plan flags a user hook the predicate would claim, default root keeps the legacy base path), 1 scan test for the project switch.
- Clippy clean on the changed crates.
- Live on this machine: `ways config show`, `targets`, `target plan ~/.claude` (all linked, settings unchanged), `ways status` first line. Sandbox config dir with a real `skills/` and a hook under `.claude/hooks/`: plan refuses the directory, flags the hook as one the merge would claim, `target add` exits 3 and writes nothing.
- Not exercised live: `target add` on a second directory, since a second target is not honest until #503 relocates the hook path anchors. The reconcile tests cover the path.

## Follow-ups

Increment 2 (installer hands off to the targets bootstrap, discovery, attended flow) and increment 3 (#503 path anchors, plural transcript resolver, observe flag, config-dir stamp on events) are filed as issues. #504 reworks onto this model as a hooks-only target kind.


aaronsb:
## Review: targets model (ADR-184/185), at d5113be8

**Verdict: revise.** The model is right and the seams are in the right places: `plan_for` runs the merge in memory, `run_targets` finishes every target before reporting the first error, `withdraw_one` removes only a symlink that resolves into the source, `base_for` is shared by plan and apply, the targets key is read from the user layer alone with a test pinning that. Three findings block: the suite fails under a default `cargo test` and writes into the real state root, withdrawal can delete a user hook with no warning and no revert, and `target add` fails on a config file that `ways config init` wrote.

### High

**1. Test suite fails in parallel and leaks into the real state root.**
`cmd::reconcile::tests::enabled_target_converges_and_disabled_target_withdraws` failed at the "our deny baseline gone" assertion in 4 of 4 default runs and passed in 3 of 3 with `--test-threads=1`. Four tests set and remove `XDG_STATE_HOME` process-wide (`reconcile.rs` tests at the `set_var`/`remove_var` pairs), so `base_path_for` resolves to another test's sandbox, or after a sibling's `remove_var` to `~/.local/state/agent-ways/targets/`. My runs left five `-tmp-ways-reconcile-*-proj` directories there. Thread the base path explicitly: give `run_targets` a base resolver parameter, or have the tests call `converge_one` and `withdraw_one` with a sandbox base. A mutex would stop the failures and still write to real state whenever a test panics between set and remove.

**2. Withdrawal drops a user hook that `entry_is_ours` claims.**
`merge_impl` filters `entry_is_ours(e)` regardless of `withdrawing` (`settings_merge.rs:169-177`), and `stripped_user_view` strips the same predicate on both sides (`:291`), so the self-audit in `withdraw_from_files` is symmetric by construction and cannot see it. The forward path surfaces this as `hooks_replaced` in the plan. `target disable`, `target remove`, and a bare `ways reconcile` over a disabled target reach `run_for_targets` with no plan. A hook at `${HOME}/.claude/hooks/my-own.sh` under `Stop` is deleted on `target disable`. A pre-#502 base that recorded user hooks is a second path, through `base_entries.contains(e)` at `:172`; that one is latent on main for the forward merge as well. Options, strongest first: withdrawal drops only what the base records plus the hooks the source `settings.json` ships; or withdrawal runs the plan first and exits 3 when a non-base entry would go; or at minimum the withdraw summary lists every entry it removed.

**3. `target add` fails on a config written by `ways config init`.**
Reproduced in a sandbox: `ways config init` then `ways config target add <dir>` prints `Error: user config is not a mapping`. The template is comments only, `serde_yaml::from_str` returns `Value::Null`, and the guard in `write_targets_to` (`config.rs:241`) covers the empty file alone. Since d5113be8 every implicit-install `ways reconcile` attempts this write, including the one `check-setup.sh` runs at session start on version drift, so anyone who ran `config init` sees "could not record the target" on every run with no way to clear it. Treat `Value::Null` as an empty mapping.

### Medium

**4. A relative directory is stored as typed.** `resolve_dir` returns the typed string and `target_add` stores it (`config_cmd.rs:199-210`, `:288`). Reproduced: `ways config target add dest2` writes `path: dest2`, and a hook-driven reconcile later resolves it against its own cwd. Store the canonical path unless the operator typed `~`.

**5. `target add --json` leaks plain lines to stdout.** After the JSON plan, `report` prints `linked skills (linked)` and friends through `println!`. Reproduced. ADR-185 item 2 wants one document on stdout. Route the per-root lines to stderr, or pass a sink.

**6. `blocked` disagrees with four other places about a claimed user hook.** `plan_for` sets `blocked` from refused roots and `hooks_removed` (`reconcile.rs:267-268`). `hooks_replaced` does not block. `print_plan`'s blocked text says "listed under REMOVE or replace", the deployment way and the PR body say an entry of the user's stops `add`, and the test `plan_flags_a_user_hook_the_predicate_would_claim` says "must say so, and block" while asserting only `hooks_replaced`. The live check in the PR body exited 3 because of the real `skills/` directory. Recommend blocking on `hooks_replaced` too: on a converged install our own hooks land in `merged` and never show as replaced, so the signal is clean.

**7. The migration write in d5113be8 rewrites config.yaml from a hook.** `write_targets_to` round-trips through `serde_yaml::Value`, so every comment is lost; the doc comment acknowledges this for an operator-typed verb. The trigger is now a bare `ways reconcile`, which `check-setup.sh` runs at session start. An annotated config gets flattened without a verb the operator typed. It also opens a read-modify-write race with `target add`: both load the implicit list, the last writer wins, and `target add`'s new entry can vanish from the file while its directory stays reconciled. Keep the migration behind an operator verb, or make the writer append a `targets:` block when the key is absent.

**8. `disable` and `remove` need the directory to exist.** `resolve_dir` bails on a missing dir, so a deleted profile cannot leave the list. `withdraw_one` already handles an absent dir cleanly. Relax the existence check for those two verbs and let `find_target` fall back to lexical comparison of expanded paths.

### Low

**9. Withdrawal with no base keeps the deny baseline.** `base_for` seeds deny empty by design, so `withdraw` on a target whose base file is gone removes hooks and allow entries and leaves our nine deny entries, while the summary says it withdrew "hooks and permissions". Say what remained. `withdraw_from_files` also returns before writing the empty base when nothing changed, so its doc line "leaves an empty base behind" holds only on a changed file.

**10. Write-then-reconcile ordering.** `target_add` and `set_enabled` write config before reconciling; `target_remove` withdraws before writing. A reconcile failure after the write leaves the record standing and the next `ways reconcile` retries, which fits the idempotent posture. The error path never says the record was written. One stderr line would close that.

**11. `--dest <non-default>` base path moved.** Before this PR every `--dest` used `state/settings-applied.json`; a non-default dest now gets `state/targets/<key>/`. A relocated `CLAUDE_CONFIG_DIR` install that ran the installer with `--dest` seeds from live on its first reconcile after upgrade, which works, and its old base file stays stale. Worth a line in the install guide.

**12. Install guide** says "the next `ways reconcile` or `ways update` records it". True only after d5113be8 and only when the write succeeds (findings 3 and 7).

### The specific questions

- **Can withdrawal drop a user hook?** Yes, finding 2. An under-recording base loses nothing extra; `entry_is_ours` backstops our own entries. An over-recording base from the pre-#502 seed drops what it recorded, and the self-audit is symmetric there too.
- **`target add` on an existing enabled target?** Safe and idempotent: plan, then converge. It makes the implicit entry explicit with the absolute path. Finding 3 applies when the config is the init template.
- **Implicit default identical to today?** Yes for destination and base path. `base_path_for` canonicalizes both sides and falls back to lexical equality when `~/.claude` is absent. The copy-mode and empty-roots guards now precede the in-place guard, which changes only which error you see first. Since d5113be8 the run also writes config.yaml.
- **Self-audit views in `withdraw_from_files`?** Correct. Before strips base hooks, `entry_is_ours` hooks, base perms and base deny from live; withdraw removes exactly that set; after strips `entry_is_ours` from the result with an empty base. The union `apply_to_files` uses collapses to the prior base because `merged.base` is empty. That symmetry is also why finding 2 is invisible to it.
- **Races with a hook-driven reconcile?** `write_targets_to` is temp plus rename, so a reader never sees a torn file. Two converges on one dest can hit EEXIST on the symlink create; that predates this PR. New since d5113be8: the read-modify-write race in finding 7.

### Verified

- Read every file in the focus list, both ADRs, the doc diffs, and the d5113be8 diff.
- `cargo test -p ways -p ways-core`: default run fails 1 of 278 in `ways` (4 of 4 attempts); `--test-threads=1` passes 278, 7, 11, 163.
- `cargo clippy`: 8 warnings, all in files this PR does not touch.
- Built the debug binary and reproduced findings 3, 4, and 5 in a sandbox with `XDG_CONFIG_HOME` and `XDG_STATE_HOME` redirected.

### Not checked

The Windows `remove_symlink` path, `agent-fmt` table rendering, `status.rs` beyond reading, `ways update` end to end live, and docs prose beyond the claims cited above.

---

aaronsb:
Remediated in the follow-up commits:

1. **Test race.** `run_targets_in` and `plan_for` take the state root as a parameter; no test touches the environment. Passes under default parallel `cargo test`, twice locally and on CI. The stray sandbox directories under the real state root are deleted.
2. **Withdrawal dropping a user hook.** Withdrawal now removes exactly the hook entries this version ships, raw or quoted. Neither the ownership predicate nor the base decides removal, so a hook added under `.claude/hooks/` after activation and a hook an old base over-recorded both survive. The self-audit strips the shipped set on both sides, with the base's permissions on the before side only. The summary line counts recorded entries left in place. New test `withdrawal_keeps_a_user_hook_the_predicate_claims_and_an_over_recorded_base_entry` covers both cases.
3. **Comments-only config.** The writer edits the file textually: it replaces the existing `targets:` block or appends one, and accepts a null document. Comments survive both the first write and a rewrite; test `write_targets_keeps_comments_and_accepts_the_init_template`.

Medium: `blocked` now includes `hooks_replaced`; the stored path is canonical unless typed with `~`; `target add --json` applies quietly so stdout is the plan document, with changed-root lines on stderr under `--quiet`; `disable` and `remove` accept a directory that is gone, `enable` refuses one; a failed reconcile after the record is written says so and names the two ways out. The migration write versus `target add` race remains a last-writer-wins on the same key with identical or superset content; noted, not closed here.

Low: `--dest <non-default>` now keys its base under `state/targets/<key>/`; the ways-update skill and installer use the default root and keep the legacy base. Install-guide wording updated to say the record happens on the next reconcile.

Issue #525, opened 2026-09-18
Bash lane: derive write targets so file-triggered ways fire under auto mode


aaronsb:
## Gap

Auto mode routes edits through Bash (`sed -i`, heredoc redirects, `tee`, `cat >`), so the PreToolUse `Edit|Write` matcher never runs. Ways keyed on `files:` (markdown, adr, validate, and every project-local file way) never fire in an auto-mode session, however much the session edits those files. Observed by a four-agent team on 2026-09-18: heavy ADR citing and README/CLAUDE.md edits, zero file-triggered disclosures.

## Design

`check-bash-pre.sh` already hands the command to `ways scan command`. Extend that lane to derive write targets from the command text and run the file matcher on each:

- `sed -i ... <path>`, `perl -pi`, `tee <path>`, `> <path>`, `>> <path>`, `cp/mv ... <path>`, `python - <<EOF` scripts that open a path for writing where the path is a literal
- Resolve relative to `cwd`; skip targets under the scratchpad or `/tmp`
- Feed each target through the same `files:` regex the Edit lane uses, deduped per command, and disclose with trigger `bash-write`

Out of scope: inferring targets from variables or subshells. A literal path is the contract; anything else stays unmatched.

## Acceptance

- A Bash command that writes a literal `README.md` path fires the same ways an Edit of `README.md` fires.
- A read-only command (grep, cat, git status) derives no targets.
- Unit tests in `scan/mod.rs` cover the target extraction table above.

PR #526, opened 2026-09-18
fix(ways): stop harness envelopes and the comms way from firing on attend traffic

aaronsb:
## Summary

A four-agent team reported ways guidance as roughly 40/60 signal to noise, with one contradiction: the comms way told an agent that sending requires approval while the attend way told it to reply without asking. Root causes and fixes:

- **Prompt lane matched harness envelopes.** A Monitor notification that wakes an idle session arrives as a `UserPromptSubmit` prompt inside a `<task-notification>` envelope, and the prompt lane matched its body as operator intent (48 of 199 notification turns in local transcripts carry a prompt-hook attachment; one Monitor-expiry notice fired five calendar ways). The envelope filter from ADR-161 now guards the prompt entry, and the attend drain header (ADR-172) counts as an envelope.
- **Session-start ways refired per prompt.** They rode the refire curve on every state scan. Now gated on the once-per-session marker, which startup, compact, and clear reset.
- **ADR macro nudged outside repos.** It exits silently when the project is not a git work tree.
- **Comms way retargeted.** Description and vocabulary name Slack, Microsoft Teams, and workspace chat; the body opens with a scope line deferring attend peer messaging to the attend way.
- **Tracking way retired.** It prescribed `.claude/todo-*.md` files; ADR-180 made GitHub issues the task list. References in the start way, start skill, compaction checkpoint, ways table, and docs updated.

The remaining gap, file-triggered ways never firing under auto mode because edits go through Bash, is filed as #525.

## Test plan

- `cargo test -p ways`: 284 + 7 + 11 passed; new unit tests cover the envelope predicate.
- `ways lint --check hooks/ways`: 0 errors, 0 warnings across 155 way files.
- `tests/test-routing-golden.sh`: 37/37 top-1.
- `scripts/check-register.sh`: clean.
- Isolated corpus built from this tree: comms scores below the fire floor on two attend-reply queries and above it on Slack and Teams queries; attend ranks first on the attend queries.
- ADR macro: empty output, exit 0 from `/tmp`; full table inside this repo.
- Local `cargo clippy` fails on eight sites in files this branch does not touch (`context.rs`, `list.rs`, `render.rs`, `rethink/`, `suggest.rs`) under the local toolchain. CI's pinned clippy is the gate.


aaronsb:
## Review Summary

**What this changes**: guards the prompt lane against harness envelopes, gates `session-start` ways on the per-session marker, silences the ADR macro outside a work tree, retargets the comms way to Slack and Teams, and retires `meta/tracking`.

**Verdict**: warn. The Rust changes are correct and traced to the description. Two leftovers from the tracking retirement should land before merge; the rest are low.

### Findings

**1. Medium. `hooks/ways/core.md:32` still teaches the retired discipline.** The line reads "After compaction, check `.claude/` for tracking files." Core is prepended to every session, so every session is still told to look for the files ADR-180 replaced with issues. The compaction checkpoint way on this branch now points at a `tasklist` issue instead. Suggest rewording to the issue list or dropping the clause.

**2. Medium. `hooks/ways/meta/compaction-checkpoint/compaction-checkpoint.md:69` links `tracking(meta)`.** The PR updated line 42 of this file but the See Also entry still names the deleted way. `ways lint` does not resolve See Also links (0 warnings on this tree), so nothing else will catch it.

**3. Low. `is_system_envelope` is case-sensitive across two lanes with different casing** (`tools/ways-cli/src/cmd/scan/mod.rs:218-226`). The prompt lane receives text `check-prompt.sh` has lowercased; the queued lane at `mod.rs:206` filters raw transcript content and lowercases only afterwards at `mod.rs:157`. Every current prefix is already lowercase, so this works today. A mixed-case envelope would need two spellings. One candidate exists: user turns beginning `[SYSTEM NOTIFICATION - NOT USER INPUT]` appear in local subagent transcripts (sidechain only; I found none in main-session transcripts, so it may never reach the prompt lane). Lowercasing inside the predicate removes the dependency on the caller.

**4. Low. The `[attend` prefix has no exercised path in the prompt lane.** In local main-session transcripts, every `[attend] ...` string is a tool result, and the drain from `attend-drain-stop.sh` is Stop-hook output. Monitor wakes arrive as `<task-notification>` and are already covered. The prefix is defensive; the only prompt it would swallow is an operator message that literally begins with "[attend". The unit test covers the predicate, not a delivery path.

**5. Low. Startup ordering.** `settings.json:133-151` registers `clear-markers.sh` and `check-state.sh` in the same `startup` matcher. Claude Code runs the hooks of one matcher concurrently, so `check-state.sh` can stamp a session-start marker before `clear-markers.sh` removes the session directory. Outcome: the way shows at startup and once more on the first prompt. This already applied to `core_is_shown` (`scan/state.rs:31-37`); the new gate inherits it. Worst case is one extra show, still far below the per-prompt refire this PR removes.

**6. Low. Stale references left behind by the retirement.**
- `hooks/ways/softwaredev/delivery/implement/macro.sh:21-23` counts `.claude/todo-*.md` as a complexity signal. Harmless (the find returns 0) but dead.
- `docs/hooks-and-ways/matching.md:261` gives "whether tracking files exist" as the file-exists example.
- `tools/ways-cli/src/cmd/scan/state.rs:129` comment example `.claude/todo-*.md`.
- `tools/ways-cli/src/cmd/tune_precision.rs:19,26,445,507,510` use `meta/tracking` as the doc example and test fixture id while the docs on this branch now say `meta/todos`. Tests do not resolve way files, so this is cosmetic.

**7. Nit. Test placement.** The two new tests in `mod queued_tests` (`mod.rs:1406-1418`) sit above `use super::*;`. They compile because they call `super::` explicitly. Move them below the `use` for consistency with the rest of the module.

### Verified

- **Q1, envelope guard.** All six prefixes are lowercase-stable. The early return still bumps the epoch (`mod.rs:114-117`), matching what `scan_prompt_surface` does at `mod.rs:240-242`. Transcript evidence supports the premise: 543 main-session user turns begin with `<task-notification`, 153 of them carry a prompt-hook attachment.
- **Q2, session-start gating.** No double gate. `state.rs:74` checks the marker at `session_dir/ways/<id>/.marker.<agent>` (`session.rs:202-208`). First pass: marker absent, `capture_show_way` reaches `way_scored`, `way_fire_outcome` returns `FirstFire` (`engagement.rs:126-127`), body emitted, marker stamped (`show/mod.rs:186`). Every later prompt: marker present, skipped before the refire curve is consulted. `clear-markers.sh` deletes the session directory on startup, compact, and clear; `resume` runs `check-state.sh` alone, so no reshow on resume, consistent with the new doc text. `refire: 0.15` on `teams.md` and `freshness.md` is now dead for the state lane; freshness still refires through the prompt lane and shares the marker, so no double show. Paths where `way_scored` returns empty without stamping (disabled way, missing `refire:`) retry silently each prompt at no cost. Teammate case holds: `inject-subagent.sh:56` writes the teammate marker after the teammate's startup, so scope resolves to teammate on its first prompt and the teams way fires.
- **Q3, tracking references.** `ways.json`, locale stubs, tests, skills, commands, and agents are clean. The way directory on main held only `tracking.md` (no provenance sidecar, no locales). `docs/design-notes/ways-functional-audit.md:107` and `docs/signal-analysis/scores.csv` are historical records and can stay.
- **Q4, comms way.** `ways lint --check` on the comms directory and the full tree: 0 errors, 0 warnings across 155 files. `scripts/check-register.sh` clean. Every rule from the old body survives: the always-safe list, the approval rule (target moved from "chat context" to "workspace and channel"), the four triage steps, the three context-layer bullets, and the auth note. The new scope line's claim that attend replies need no approval matches `attend.md:50`. All three See Also targets exist on disk.
- `cargo test -p ways queued_tests`: 8 passed, including the two new tests.
- ADR macro guard at `hooks/ways/documentation/adr/macro.sh:11` uses `&>` under a bash shebang; fine.
- The `tasklist` label named in `compaction-checkpoint.md:42` matches `issues-pull.sh:3` and ADR-180.

### Not verified

- The isolated-corpus scores for comms on attend and Slack queries, and the routing golden run.
- Whether subagent `[SYSTEM NOTIFICATION]` turns ever pass through `UserPromptSubmit`.

---

Issue #538, opened 2026-09-19
attend: MCP server mode for outbound and queries; inbound stays on Monitor and the Stop hook


aaronsb:
## Proposal

Run attend as two pieces:

1. **Inbound**: the Monitor-hosted `attend run` sensor loop and the Stop-hook drain, unchanged. These are the only paths that deliver into the model's turn.
2. **Outbound and queries**: an MCP server exposing `send`, `reply`, `peers`, `inbox`, `status`, `join`, `leave` as typed tools.

## Why

Three of the misuse classes the messaging reheat exists to prevent disappear structurally:

- No Bash surface for `attend run` or for reaching into attend-owned directories.
- Typed `message: string` ends the "always double-quote" rule; there are no shell metacharacters.
- Envelope fields (`to`, `on_behalf_of`, `channel`) become schema parameters instead of flags to remember.

The send-vs-reply guidance moves into the tool descriptions, which is tool-triggered disclosure: guidance present exactly when the tool is in hand. In the transcript audit that was the highest-adherence pattern.

## Constraint: MCP cannot push into the conversation

Checked against the 2026-07-28 protocol revision: the initialize handshake and sessions are removed, sampling/roots/logging are deprecated, and server-initiated requests are replaced by multi-round-trip responses to client requests. The only unsolicited server-to-client channel is `subscriptions/listen`, which carries list- and resource-change notifications to the host, not content into the model's turn. Inbound therefore stays on Monitor and the Stop hook. The two-piece split is forced, not a preference.

## Design notes

- `reply` auto-threads to "the most recent peer message in the inbox." The MCP server must share the seen-set and inbox cursor with the drain, not keep its own.
- CLI and MCP should be two frontends over one library, or the three-source documentation lockstep returns as a two-implementation one.
- Identity: the server runs per session; `whoami` resolution has to be per-connection, not per-process.
- Permissions: MCP tools get their own permission entries (`mcp__attend__send`), which is where an operator can pre-authorize outbound sends for unattended runs.

Needs an ADR: this changes the CLI-is-the-contract statement in the skill and the way.

aaronsb:
[...]

Issue #536, opened 2026-09-19
attend: PreToolUse deny for the two mechanically checkable misuses


aaronsb:
## Finding

The attend messaging guidance is stated in three synchronized sources (SKILL.md, the sensor-disclosure reheat, the attend way) because agents misuse the tool without the repetition. Two of the misuses have mechanical signatures:

- running `attend run` from Bash instead of Monitor
- reaching into `~/.cache/attend/` or `~/.config/attend/` instead of using the CLI

Both are constraints, and both are currently enforced by prose.

## Proposal

PreToolUse deny rules, shipped with the attend hooks:

- `Bash(attend run*)` → deny with a one-line reason pointing at `/attend`
- `Read`/`Edit`/`Write`/`Bash` touching paths under the attend cache or config directories → deny with a one-line reason pointing at the CLI

The reheat then carries only the judgment sentences (send vs reply, silence is valid), and the three-source lockstep shrinks to those. Anthropic's Opus 5 prompting guide and OpenAI's harness-engineering post both land on the same split: mechanical rules in the harness, judgment in text.

If the MCP-server-mode issue lands, the Bash surface for these misuses disappears and this becomes moot for MCP-using sessions; it remains useful for CLI-only sessions in the meantime.

Issue #532, opened 2026-09-19
attend: carry sender kind and on-behalf-of as structured fields on every signal


aaronsb:
## Finding

A signal names its sender only in display text. A receiving session learns "human or agent" by reading the name. On 2026-09-19 a coordination agent (Jovan) that runs outbound Slack and email under a send-requires-approval rule said it could identify the operator's request only because the sender string was `bockeliea@ssh`. A peer agent relaying "send this message" is indistinguishable in structure from operator approval.

The same gap misled `/insights` on 2026-09-19: eight sessions coordinating over the bus were scored as one operator directing one agent each, because the delivered drain rows carry the sender in prose and no `origin` field. Verified in transcripts: 19 Stop-hook drain rows and 84 teammate-message rows across six sessions, all with no `origin`; rows the harness stamps carry `origin.kind` of `human`, `task-notification`, or `peer`.

## Proposal

Structured fields on every signal, written by the sending `attend` process, not by the model:

- `from_kind`: `human` | `agent`
- `from_id`: canonical sender id (session id today; key fingerprint if signing lands)
- `on_behalf_of` (optional): the id whose request this carries, when the sender is relaying

The bus stays flat. These are evidence for the receiver to weigh, not a rank. A receiver applying an approval rule can check `from_kind` mechanically. Prior art for the shape: RFC 8693 `sub`/`act`, RFC 5322 `From`/`Sender`.

Both conduits (Monitor line, Stop-hook drain) should render the fields, and the drain should emit them in a form the transcript's `origin` object can carry.

## Evidence

- Bus history 2026-09-14 to 09-18: 128 messages, 24 human, 104 from eight agent sessions.
- Report and dataset: `~/Downloads/ways-measured-evidence/` on the measured workstation.
- Related: cross-machine relay (future ADR) needs these fields to have anything to sign.

aaronsb:
Status 2026-09-19: design note merged (PR #541, docs/design-notes/attend-envelope-fields.md). #532 and #533 build as one change following its six-step checklist. #534 landed first (PR #542) and gives the drain the canonical sender id these fields extend.

PR #540, opened 2026-09-20
ADR-187: Attend MCP server mode: outbound and queries as typed tools, inbound stays on Monitor and the Stop hook

aaronsb:
Proposed ADR for #538. Also settles the interim for #536.

**Decision summary**

- Two pieces, forced by the 2026-07-28 MCP revision: no server path reaches the model's turn, so inbound stays on the Monitor-hosted sensor loop and the Stop-hook drain (ADR-172). Outbound and queries (`send`, `reply`, `peers`, `inbox`, `status`, `join`, `leave`) become typed MCP tools.
- One library, two frontends. Verb bodies move out of `cmd/` into a library target; clap help, `docs/cli/attend.md`, and the MCP tool schema derive from one description string per verb. MCP `reply` reads the same `last_inbound` record and seen-set the drain writes; MCP `inbox` is read-only.
- Per-connection identity via the ADR-171 tuple; `send` and `reply` refuse under `resolved: false`.
- Per-tool permission names (`mcp__attend__send` and friends). Shipped settings allow all seven to preserve today's autonomy; removing one is the operator's gate, and unattended runs confirm the entries are present.
- Guidance split: mechanical layer (send vs reply, scope, length) into tool descriptions; judgment layer (autonomy, silence, two-conduit contract) stays in the way and the skill; the reheat shrinks to the judgment layer.
- Interim: the #536 denies ship as attend-owned guard hooks under ADR-181's contract, deactivated by default, with no separate activation switch. Marked interim, removed when MCP is the default frontend.
- Envelope fields from #532/#533 cross the tool boundary as typed parameters; `from_kind` and `from_id` are server-set. Semantics stay with the parallel design note.
- Changes the "CLI is the contract" statement in the skill and the way to "the tool surface is the contract".

**Numbering note**: the adr tool offered ADR-183 (a gap), but PR #504 already claims ADR-183 on its branch (commit 94069fed renumbers the install-scope ADR to 183). This PR uses 187, the first free number.

**Gates**: `docs/scripts/adr lint` clean (0 errors, 0 warnings); `INDEX.md` regenerated; no em dashes or antithesis shapes in the new file.

Refs #538, #536. Related: ADR-136, ADR-171, ADR-172, ADR-173, ADR-181, ADR-184, ADR-185.


aaronsb:
Operator decision (2026-09-19): item 6 stands as written. All seven MCP tools are allowed by default in the shipped settings, and the CLI frontend remains alongside the MCP server per items 3, 6, and 10.

